PR #5, opened 2018-09-27
Remove empty package CommonSocPkg

gdong1:
There is nothing in CommonSocPkg, so remove it.

TEST=Build QEMU and APL success


gdong1:
It is ok for me to keep it as a place holder.

PR #11, opened 2018-10-02
Resolve the problem that VS toolchain is often set to older version

auguchen:
When there are multi versions of VS installed on the Windows host
system, BuildLoader.py often set up old VS version for the toolchain
variable. Resolve it by changing the structure type of vs_ver_list.

Signed-off-by: Augustine Chen <augustine.chen@intel.com>


huang-jin-intel:
Fixed issue #12

PR #13, opened 2018-10-05
Add ConsoleInLib to abstract console input

mauricema:
Current implementation assumes serial port is the only input console
device supported.  But other input console devices can be added later
on. This patch added a ConsoleInLib to abstract the input console
interfaces.  It also added PCDs to control enabled input console
devices.

### DIFF
--- a/BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.c
+++ b/BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.c
@@ -1,0 +1,74 @@
+/** @file
+Implementation for console input library
+
+Copyright (c) 2018, Intel Corporation. All rights reserved.<BR>
+This program and the accompanying materials
+are licensed and made available under the terms and conditions of the BSD License
+which accompanies this distribution.  The full text of the license may be found at
+http://opensource.org/licenses/bsd-license.php
+
+THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+
+**/
+
+#include <Library/SerialPortLib.h>
+#include <Library/ConsoleInLib.h>
+
+/**
+  Poll a console to see if there is any data waiting to be read.
+
+  @retval FALSE                There is no data waiting to be read from the console.
+  @retval TRUE                 Data is waiting to be read from the console.
+
+**/
+BOOLEAN
+EFIAPI
+ConsolePoll (
+  VOID
+  )
+{
+
+  if ((PcdGet32 (PcdConsoleInDeviceMask) & ConsoleInSerialPort) != 0) {
+    if (SerialPortPoll ()) {
+      return TRUE;
+    }
+  }
+
+  return FALSE;
+}
+
+/**
+  Read data from a console device into a buffer.
+
+  @param  Buffer           Pointer to the data buffer to store the data read from the console device.
+  @param  NumberOfBytes    Number of bytes to read from the console device.
+
+  @retval 0                NumberOfBytes is 0.
+  @retval >0               The number of bytes read from the console device.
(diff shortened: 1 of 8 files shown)


BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.inf
```diff
@@ -0,0 +1,37 @@
+## @file
+#  INF for console input library
+#
+#  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
```
mborgerson: Year

BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.inf
```diff
@@ -0,0 +1,37 @@
...
+  MdePkg/MdePkg.dec
+  BootloaderCommonPkg/BootloaderCommonPkg.dec
+
+[LibraryClasses]
+  BaseLib
+
+[Guids]
+
+[Pcd]
+  gPlatformCommonLibTokenSpaceGuid.PcdConsoleInDeviceMask
```
mborgerson: Newline at EOF

BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.c
```diff
@@ -0,0 +1,71 @@
...
+  IN UINTN  NumberOfBytes
+  )
+{
+  UINTN    Count;
+
+  Count = 0;
+  while (Count < NumberOfBytes) {
+    if ((PcdGet32 (PcdConsoleInDeviceMask) & ConsoleInSerialPort) != 0) {
+      if (SerialPortPoll ()) {
+        Count += SerialPortRead (Buffer++, 1);
```
mborgerson: I think it would be safer to advance the buffer like `Buffer += Count` on the following line, in case SerialPortRead returns 0 for some reason.

BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -210,6 +211,9 @@
   gPlatformModuleTokenSpaceGuid.PcdCfgDataLoadSource      | $(CFGDATA_REGION_TYPE)
   gPlatformModuleTokenSpaceGuid.PcdCfgDatabaseSize        | $(CFG_DATABASE_SIZE)
 
+  gPlatformCommonLibTokenSpaceGuid.PcdConsoleInDeviceMask  | $(CONSOLE_IN_DEVICE_MASK)
+  gPlatformCommonLibTokenSpaceGuid.PcdConsoleOutDeviceMask | $(CONSOLE_OUT_DEVICE_MASK)
```
myguitar: Reserved for future ConsoleOutLib?

mborgerson:
Should we call this simply "ConsoleLib" instead to extend it later with write support?

BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.inf
mauricema: Good catch,  Will change.

BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.inf
mauricema: Sure, will add new line.

BootloaderCommonPkg/Library/ConsoleInLib/ConsoleInLib.c
mauricema: Good point.   Agreed, will change the code.

BootloaderCorePkg/BootloaderCorePkg.dsc
mauricema: Yes, plan to add both ConsoleInLib and ConsoleOutLib.

mauricema:
>           Should we call this simply "ConsoleLib" instead to extend it later with write support?

I think it is better to have ConsoleInLib and ConsoleOutLib separated since they can be controlled separated.

PR #18, opened 2018-10-12
Enable DebugAgentLib to support source level debug over serial

myguitar:
By default, source level debug is NOT enabled. ENABLE_SOURCE_DEBUG
needs to be set to 1 in BoardConfig.py of each Platforms.
- self.ENABLE_SOURCE_DEBUG  = 1

As an initial drop, there are some limitations on APL platform.
- Stage1A does not include DebugAgentLib due to Stage1A size
  limitation(32KB). Further optimization is required.
- DebugAgentLib supports ONLY POSTMEM debugging on APL currently.

Refer to EDKII Debugging:
- https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Debugging
- https://github.com/tianocore/tianocore.github.io/wiki/SourceLevelDebugPkg

Change-Id: Ia28c5470bc5755768f2b380cc1dabbcb8ee60f0d


Platform/QemuBoardPkg/BoardConfig.py
```diff
@@ -51,14 +51,20 @@ def __init__(self, *args, **kwargs):
...
+
 		# OS_PK | FWU_PK | CFG_PK | FWU_PLD | PLD | Stage2 | Stage1B
 		self.VERIFIED_BOOT_HASH_MASK  = 0x000000D7
 		if self.ENABLE_FWU:
 			self.VERIFIED_BOOT_HASH_MASK  |= 0x00000028
 
 		self.STAGE1B_XIP          = 0
 
-		self.STAGE1A_SIZE         = 0x00008000
+		self.STAGE1A_SIZE         = 0x0000A000
```
mauricema: Why it is 0xa000 now?   APL SoC can only support up to 32KB IBBL (Stage1A).

Platform/QemuBoardPkg/BoardConfig.py
```diff
@@ -96,7 +102,7 @@ def __init__(self, *args, **kwargs):
 		if not self.STAGE1B_XIP:
 			# For Stage1B, it can be compressed if STAGE1B_XIP is 0
 			# If so, 	STAGE1B_FD_BASE/STAGE1B_FD_SIZE need to be defined
-			self.STAGE1B_FD_SIZE    = 0x2C000
+			self.STAGE1B_FD_SIZE    = 0x30000
```
mauricema: If the size change is due to source level debug, then we should use the ENABLE_SOURCE_DEBUG to control it.  Otherwise, it will increase the size for all the build regardless of source level debug enable or not.

Platform/QemuBoardPkg/BoardConfig.py
myguitar: This is not for APL, but for QEMU to support Stage1A debugging. Let me increase the size depending on ENABLE_SOURCE_DEBUG.

Platform/QemuBoardPkg/BoardConfig.py
myguitar: Okay. I will do that.

Platform/QemuBoardPkg/BoardConfig.py
huang-jin-intel: I'd prefer enabling debugging by default in QEMU so we cover some testing automatically.

mauricema: Oops, my mistake! 
But let us use ENABLE_SORUCE_DEBUG to control the size.  :)

BootloaderCorePkg/Stage1A/Stage1A.c
```diff
@@ -288,6 +267,11 @@ SecStartup (
     Status = PeCoffRelocateImage (StageHdr->Base);
     if (!EFI_ERROR (Status)) {
       EnableCodeExecution ();
+
+      ImageContext.ImageAddress = (PHYSICAL_ADDRESS)(UINTN) StageHdr->Base;
```
mauricema: Can we move PeCoffLoaderGetPdbPointer and PeCoffLoaderRelocationImageExtraAction into  PeCoffRelocationImage() ?

BootloaderCorePkg/Stage1A/Stage1A.c
```diff
@@ -369,6 +411,9 @@ ContinueFunc (
 
   // Jump into Stage 1B entry
   if (StageBase != 0) {
+    ImageContext.ImageAddress = (PHYSICAL_ADDRESS)(UINTN) GET_STAGE_MODULE_BASE (StageBase);
```
mauricema: Maybe a good idea to add a StageEntry hook point to call these functions.   Then the code will be simpler.

BootloaderCorePkg/Stage1A/Stage1A.c
myguitar: Okay, I will do that.

BootloaderCorePkg/Stage1A/Stage1A.c
myguitar: Okay, let me make a common interface.

Platform/QemuBoardPkg/BoardConfig.py
myguitar: Aaron - In fact, QEMU can be debugged by gdb w/o DebugAgentLib. This is for testing purpose.

mauricema:
Looks much better now, thx!  :)

PR #20, opened 2018-10-16
Start DebugAgent from the beginning of Stage1B

myguitar:
Hardware Breakpoints must be used during CAR execution until
memory has been initiailzed for APL platform.

In order to use Hareware Breakpoints in UDK debugger, add
FlashRange in UDK config file and force to cover CAR area.
In SoftDebugger.ini of Windows UDK Debugger
Or In /etc/udkdebugger.conf of Linux UDK Debugger
  [Target System]
  FlashRange        = 0xFEF00000:0x1100000

* Includes correct Stage1A module base report

Change-Id: I2d4401bb82ec16531dc4b9e89a744d255ee36487


BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -278,6 +279,12 @@ SecStartup2 (
 
   DEBUG ((DEBUG_INFO, "\n============= Intel Slim Bootloader STAGE1B =============\n"));
 
+  StageBase = PCD_GET32_WITH_ADJUST (PcdStage1BLoadBase);
```
mauricema: Can we pass this info directly from Stage1A using STAGE1A_HOB ?  I think it might be more generic.

myguitar: Okay, let me add Stage1BBase in STAGE1A_HOB.

PR #22, opened 2018-10-20
Enable support for basic framebuffer text console

mborgerson:
This PR adds support for displaying `DEBUG` messages and shell command output to the frame buffer through a new **ConsoleOutLib**, with an interface similar in design to the newly added ConsoleInLib. 

Debug messages stored in the log buffer can be played back once the frame buffer is available. Messages over the serial port can still be routed directly through the SerialPortLib handlers, or through the new ConsoleOut library. Additionally, ConsoleOut supports routing messages to the serial port, exclusively or along with the frame buffer.

The default behavior still remains the same, only the splash screen is displayed on the frame buffer and messages are routed to the serial port (through this new ConsoleOutLib).


BootloaderCommonPkg/Library/BootloaderDebugLib/DebugLib.c
```diff
@@ -73,14 +74,17 @@ DebugPrint (
...
+  // Send the print string to debug output handler
   //
   if (PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_LOG_BUFFER) {
     DebugLogBufferWrite  ((UINT8 *)Buffer, AsciiStrLen (Buffer));
   }
   if (PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_SERIAL_PORT) {
     SerialPortWrite ((UINT8 *)Buffer, AsciiStrLen (Buffer));
   }
+  if (PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_CONSOLE) {
+    ConsoleWrite ((UINT8 *)Buffer, AsciiStrLen (Buffer));
```
mauricema: if serial is also in console out device,  will it cause issue in output ?

BootloaderCommonPkg/Library/ConsoleOutLib/ConsoleOutLib.c
```diff
@@ -0,0 +1,174 @@
...
+/**
+  Perform a blanking backspace on the virtual console.
+
+  @param  Count     Number of backspaces.
+
+  @retval >=0       Number of backspaces performed.
+**/
+UINTN
+EFIAPI
+ConsoleBackspace (
```
mauricema: do we need this ConsoleBackSpace () API ?   Maybe better to keep it as internal function.

BootloaderCommonPkg/Library/ConsoleOutLib/ConsoleOutLibNull.c
```diff
@@ -0,0 +1,60 @@
+/** @file
```
mauricema: I think fixed ConsoleOuptutDeviceMask PCD will remove all the code from the ConsoleOutLib if we set it to 0.  Maybe we won't have to have NULL instance lib.

BootloaderCorePkg/Stage2/Stage2Support.c
```diff
@@ -122,6 +122,79 @@ DisplaySplash (
...
+  return EFI_SUCCESS;
+}
+
+/**
+  Update console framebuffer pointer based on gLoaderFrameBufferInfoGuid HOB.
+
+**/
+VOID
+EFIAPI
+ConsoleUpdateFrameBuffer (
```
mauricema: BuildExtraInfoHob()  in State2Support.c already did this.   I think we can break out one function called  UpdateFrameBuffer().   Then everybody call the same function to update framebuffer info.

BootloaderCorePkg/Stage2/Stage2Support.c
```diff
@@ -122,6 +122,79 @@ DisplaySplash (
...
 
+/**
+  Initialize the virtual console framebuffer back-end.
+
+  @retval EFI_UNSUPPORTED Could not initilaize the console.
+  @retval EFI_SUCCESS     Console initialized.
+**/
+EFI_STATUS
+EFIAPI
+ConsoleInitFrameBuffer (
```
mauricema: I think Output console init needs to be in ConsoleOutLib instead of Stage2 core code.

Platform/QemuBoardPkg/BoardConfig.py
```diff
@@ -51,6 +51,11 @@ def __init__(self, *args, **kwargs):
 
 		self.ENABLE_FWU           = 1
 
+		# To enable graphics console, uncomment the following lines
```
myguitar: Verified on qemu. How is it on APL board?

BootloaderCommonPkg/Include/Library/ConsoleOutLib.h
```diff
@@ -0,0 +1,59 @@
...
+  Perform a blanking backspace on the virtual console.
+
+  @param  Count     Number of backspaces.
+
+  @retval >=0       Number of backspaces performed.
+
+**/
+UINTN
+EFIAPI
+ConsoleBackspace (
```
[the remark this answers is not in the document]
mborgerson: Yeah, I can process `\b` characters inside `ConsoleWrite` and perform the backspace appropriately.

BootloaderCommonPkg/Library/BootloaderDebugLib/DebugLib.c
mborgerson: If serial and console are enabled, and console out has serial enabled, all messages sent via DebugLib will be delivered on the serial port twice.

BootloaderCommonPkg/Library/ConsoleOutLib/ConsoleOutLibNull.c
mborgerson: Will remove the null library instance.

BootloaderCommonPkg/Library/ConsoleOutLib/ConsoleOutLib.c
```diff
@@ -0,0 +1,218 @@
...
+  PcdSet64(PcdFrameBufferConsoleBase, (UINT64)(UINTN)Console);
+
+  // If the console is configured to output messages to the serial port, also
+  // print out any messages that may be stored in output log buffer as the
+  // framebuffer console is not initialized until much later in the boot flow.
+  if ((PcdGet32 (PcdConsoleOutDeviceMask) & ConsoleOutSerialPort)
+      && (PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_LOG_BUFFER)) {
+    LogBufHdr = (DEBUG_LOG_BUFFER_HEADER *) GetDebugLogBufferPtr ();
+    if (LogBufHdr != NULL) {
+      ConsoleWrite (LogBufHdr->Buffer, LogBufHdr->UsedLength - LogBufHdr->HeaderLength);
```
myguitar: To avoid duplicated message in serial port, how about this?
Unmask ConsoleOutSerialPort in PcdConsoleOutDeviceMask
ConsoleWrite
Mask again

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -600,6 +605,25 @@ InitConsole (
...
   }
 
+  if (PcdGet32 (PcdConsoleOutDeviceMask) & ConsoleOutFrameBuffer) {
+    FrameBuffer = (FRAME_BUFFER_INFO *)GetGuidHobData (NULL, NULL, &gLoaderFrameBufferInfoGuid);
+    if (FrameBuffer != NULL) {
+      Width  = FrameBuffer->HorizontalResolution;
+      Height = FrameBuffer->VerticalResolution;
+      OffX   = 0;
+      OffY   = 0;
+      if (Width > 800) {
```
mauricema: Maybe we can use fixed PCD to control the resolution later on.

mauricema:
It looks good to me now.

mborgerson:
Please note that I've made a few improvements to this since opening the PR. Specifically, I've made some simplifications of the interfaces and I've added additional levels of text buffer caching which significantly improve performance (especially with regard to scrolling).

PR #23, opened 2018-10-24
Support ELF format payload

myguitar:
In order to support ELF format payload in Stage2, ElfLib is
separated from MultibootLib.

Change-Id: Iac45e8511101e2d12cb9a9f7f9bb2394a1e7be39

### DIFF
--- a/BootloaderCommonPkg/Include/Library/ElfLib.h
+++ b/BootloaderCommonPkg/Include/Library/ElfLib.h
@@ -1,0 +1,50 @@
+/** @file
+  ELF library
+
+  Copyright (c) 2018, Intel Corporation. All rights reserved.<BR>
+  This program and the accompanying materials
+  are licensed and made available under the terms and conditions of the BSD License
+  which accompanies this distribution.  The full text of the license may be found at
+  http://opensource.org/licenses/bsd-license.php.
+
+  THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+  WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+
+**/
+
+#ifndef __ELF_LIB_H__
+#define __ELF_LIB_H__
+
+#include <Uefi/UefiBaseType.h>
+
+/**
+  Check if the image is a bootable ELF image.
+
+  @param[in]  ImageBase      Memory address of an image
+
+  @retval     TRUE           Image is a bootable ELF image
+  @retval     FALSE          Not a bootable ELF image
+**/
+BOOLEAN
+EFIAPI
+IsElfImage (
+  IN  VOID                   *ImageBase
+  );
+
+/**
+  Load the ELF image to specified address in ELF header.
+
+  This function load ELF image section by section into memory address specified
+  in ELF program header, and also load ELF symbols.
+
+  @param[in]  ImageBase      Memory address of an image
+
+  @retval Image entry point  The entry point of ELF image if load image success
+  @retval NULL               Error with loading ELF image
+**/
+UINT32 *
+LoadElfImage (
+  IN  VOID                   *ImageBase
+  );
+
+#endif /* __ELF_LIB_H__ */
--- a/BootloaderCommonPkg/Library/ElfLib/ElfLib.c
+++ b/BootloaderCommonPkg/Library/ElfLib/ElfLib.c
@@ -1,0 +1,213 @@
+/** @file
+
+  Copyright (c) 2018, Intel Corporation. All rights reserved.<BR>
+  This program and the accompanying materials
+  are licensed and made available under the terms and conditions of the BSD License
+  which accompanies this distribution.  The full text of the license may be found at
+  http://opensource.org/licenses/bsd-license.php.
+
+  THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+  WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+
+**/
+/*-
+ * Copyright (c) 1998 Michael Smith <msmith@freebsd.org>
+ * Copyright (c) 1998 Peter Wemm <peter@freebsd.org>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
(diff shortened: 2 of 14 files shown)


BootloaderCommonPkg/Library/ElfLib/ElfLib.c
```diff
@@ -0,0 +1,228 @@
...
+
+  Eaddr = 0;
+  for (Index = 0; Index < Ehdr->E_Phnum ; Index ++) {
+    if (Phdr[Index].P_Type != PT_LOAD) {
+      continue;
+    }
+    Eaddr = MAX (Eaddr, Phdr[Index].P_Paddr + Phdr[Index].P_Memsz);
+  }
+
+  MbInfo->Flags       |= MULTIBOOT_INFO_HAS_ELF_SYMS;
```
mborgerson: I think the `MULTIBOOT_INFO_HAS_ELF_SYMS` name is a bit misleading and does not exactly line up with [the spec](https://www.gnu.org/software/grub/manual/multiboot/multiboot.html) which defines this bit as:
```
     /* is there an ELF section header table? */
     #define MULTIBOOT_INFO_ELF_SHDR                 0X00000020
```

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  Shdr = (ELF32_SHDR *) (UINTN) Eaddr;
+  Size = Ehdr->E_Shnum * Ehdr->E_Shentsize;
+  CopyMem (Shdr, (VOID *) ((UINT32)ImageBase + Ehdr->E_Shoff), Size);
+  Eaddr += Size;
+
+  // Copy sections that are part of the symbol table.
+  for (Index = 0 ; Index < Ehdr->E_Shnum ; Index += 1) {
+    if (Shdr[Index].Sh_Type != SHT_SYMTAB) {
```
mborgerson: Two questions here: If we are loading the section header table...why not load the entire thing instead of just the symbol table? If we are loading just the symbol table..why should we load it at all? Does any kernel actually depend on the symbol table being loaded?

BootloaderCommonPkg/Library/ElfLib/ElfLib.inf
```diff
@@ -0,0 +1,43 @@
+## @file
+#
+#  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
```
mborgerson: Year

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
```diff
@@ -0,0 +1,228 @@
...
+  ELF32_PHDR                 *Phdr;
+  UINT32                      Eaddr;
+  UINT32                      Index;
+  UINT32                      Msize;
+  UINT32                      Fsize;
+  UINT32                      Mpos;
+  UINT32                      Ipos;
+
+  //
+  // TODO: Check if image, load region overlap; if yes, move image away.
```
mborgerson: Perhaps this can be deferred to another PR, but we should really do some bounds checking to make sure we can actually load the ELF segments into memory at the specified locations. Do we have enough memory? Have the ranges been reserved for something else? Etc.

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
myguitar: Yes, it seems. Let me include the change in next commit rather than new PR.

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
myguitar: All necessary things for execution binary are loaded thru program header. Only symbol information needs to be loaded thru section header. If we do not use program header, we can load the entire thing thru section header.
- Code from program header
- Symbol from section header
This is a very minimal way to load executable code and symbol because program header is much simpler than section header.

I am not sure kernel depends on the symbol table, but it may be optional for debugging. If we don't want to load the information, we can strip ELF.

Someone has any idea?

BootloaderCommonPkg/Library/ElfLib/ElfLib.inf
myguitar: Let me change it.

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
myguitar: Yes, agree and let's defer this. PE format image relocation as well.

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
myguitar: I double checked it. It may need another PR since other bits need to be updated as well. Let me use the name as it is in this commit.

BootloaderCommonPkg/Include/Library/ElfLib.h
```diff
@@ -0,0 +1,70 @@
...
+  THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+  WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+
+**/
+
+#ifndef __ELF_LIB_H__
+#define __ELF_LIB_H__
+
+#include <Uefi/UefiBaseType.h>
+#include <Library/MultibootLib.h>
```
mauricema: can we decouple Multiboot and ELF?   Why ELF needs to include MultiBoot?   Shouldn't be the other way ?

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
mauricema: dynamic symbols could potentially be used at RT for elf.

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
```diff
@@ -0,0 +1,228 @@
...
+
+  @param[in]        ImageBase   Memory address of an image
+  @param[in, out]   MbInfo      Pointer to MULTIBOOT_INFO which needs to be updated
+
+  @retval RETURN_SUCCESS        MULTIBOOT_INFO successfully updated.
+  @retval RETURN_NOT_FOUND      No ELF Symbol information found.
+
+**/
+RETURN_STATUS
+UpdateElfMultibootInfo (
```
mauricema: I think we should try to move MULTBOOT related code into Multiboot library.  Here if required, create new function to return ELF info to be populated into MULTIBOOT_INFO.

myguitar: Okay, let me create a new function for ELF info and use it in MultibootLib.

BootloaderCommonPkg/Include/Library/ElfLib.h
myguitar: I will do that.

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
myguitar: SHT_DYNSYM(dynamic symbols) is a smaller version of SHT_SYMTAB(symbols). SYMTAB already includes DYNSYM.

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -208,11 +208,12 @@ SetupBootImage (
   MultiBoot = &LoadedImage->Image.MultiBoot;
   if (IsElfImage (BootFile->Addr)) {
     DEBUG ((DEBUG_INFO, "Boot image is ELF format...\n"));
-    EntryPoint = LoadElfImage (MultiBoot);
+    EntryPoint = LoadElfImage (BootFile->Addr);
     if (EntryPoint != NULL) {
       if (IsMultiboot (BootFile->Addr)) {
         DEBUG ((DEBUG_INFO, "and Image is Multiboot format\n"));
         SetupMultibootInfo (MultiBoot);
+        UpdateElfMultibootInfo (BootFile->Addr, &MultiBoot->MbInfo);
```
gdong1: Prefer add UpdateElfMultibootInfo() into SetupMultibootInfo() to make this file clean. both parameters in new function could be got from MultiBoot parameter of SetupMultibootInfo().

BootloaderCommonPkg/Library/ElfLib/ElfLib.c
mborgerson: What I mean here is that we are providing ELF section information through the multiboot interface, but only the symbol section. It could be the case that the OS depends on other section headers as well (.text/.data/etc). Should we not also load these if we are doing this?

PayloadPkg/OsLoader/OsLoader.c
gdong1: I am OK with that. For all multiple boot support to ELF image, we put it in SetupMultibootInfo(). we could enhance this function later if we really need it.

myguitar:
Windows build broken - one more patch is coming.

PR #25, opened 2018-10-26
Enable graphics framebuffer as WriteCombining

mauricema:
On APL platform, all PCI MMIO range is set to UC in current
implementation. It includes graphics framebuffer MMIO. It
caused the system performance issue due to large mount of
framebuffer write access. This patch set framebuffer as
WC (WriteCombining) per recommendation to enhance system
performance.

### DIFF
--- a/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -556,6 +556,44 @@ RestoreOtgRole (
 }
 
 /**
+  Set framebuffer range as WC using MTRR to improve performance.
+
+  The BSP MTRR needs to be programmed before FspSiliconInit() API so that
+  all APs' MTRRs will be syned up during FspSiliconInit() call.
+
+**/
+VOID
+SetFrameBufferWriteCombining (
+  VOID
+)
+{
+  UINT32             MsrIdx;
+  UINT32             MsrMax;
+  UINT32             Base;
+
+  // Enable Framebuffer as WC.
+  MsrMax = EFI_MSR_CACHE_VARIABLE_MTRR_BASE +
+           (2 * (UINT32)(AsmReadMsr64(EFI_MSR_IA32_MTRR_CAP) & B_EFI_MSR_IA32_MTRR_CAP_VARIABLE_SUPPORT));
+  for (MsrIdx = EFI_MSR_CACHE_VARIABLE_MTRR_BASE; MsrIdx < MsrMax; MsrIdx += 2) {
+    // Try to find a free MTRR pair
+    if ((AsmReadMsr64(MsrIdx + 1) & B_EFI_MSR_CACHE_MTRR_VALID) == 0) {
+      break;
+    }
+  }
+
+  if (MsrIdx < MsrMax) {
+    // Framebuffer belongs to PMEM32 in PCI resource allocation.
+    // The 1st 256MB from PcdPciResourceMem32Base will be consumed by MEM32 resource.
+    // And framebuffer should be allocated to the next 256MB aligned address.
+    Base = (PcdGet32 (PcdPciResourceMem32Base) + SIZE_256MB)  &  ~(SIZE_256MB - 1);
+    AsmWriteMsr64 (MsrIdx,     Base | CACHE_WRITECOMBINING);
+    AsmWriteMsr64 (MsrIdx + 1, 0xF00000000ULL + B_EFI_MSR_CACHE_MTRR_VALID + (UINT32)(~(SIZE_256MB - 1)));
+  } else {
+    DEBUG ((DEBUG_WARN, "Failed to find a free MTRR pair for framebuffer!\n"));
+  }
+}
+
+/**
   Clear FSP HOB data
 
 **/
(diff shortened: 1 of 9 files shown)


Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -606,6 +642,10 @@ BoardInit (
       VariableConstructor (VarBase, VarSize);
     }
     SaveOtgRole();
+    if (PcdGetBool (PcdSplashEnabled)) {
```
mborgerson: Framebuffer console output may be turned on while splash is disabled. In this case I think we should also enable WC. Can we add that to this condition? Or perhaps a new generic PCD which is simple "Framebuffer Enabled"

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -555,6 +555,42 @@ RestoreOtgRole (
...
+  }
+
+  if (MsrIdx < MsrMax) {
+    // Framebuffer belongs to PMEM32 in PCI resource allocation.
+    // The 1st 256MB from PcdPciResourceMem32Base will be consumed by MEM32 resource.
+    // And framebuffer should be allocated to the next 256MB aligned address.
+    Base = (PcdGet32 (PcdPciResourceMem32Base) + SIZE_256MB)  &  ~(SIZE_256MB - 1);
+    AsmWriteMsr64 (MsrIdx,     Base | CACHE_WRITECOMBINING);
+    AsmWriteMsr64 (MsrIdx + 1, 0xF00000000ULL + B_EFI_MSR_CACHE_MTRR_VALID + (UINT32)(~(SIZE_256MB - 1)));
+  }
```
mborgerson: Should we emit a warning message if a VMTRR could not be allocated?

mauricema: We can add a debug message for it.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: That is a good point.  I think it is better to add a new PCD to control if the GFX should be initialized by bootloader.  Then we can use this PCD to control.

mauricema:
@mborgerson  Update per the review comments, please take a look.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -886,7 +928,11 @@ UpdateFspConfig (
   FspsConfig->HDAudioClkGate              = 1;
   FspsConfig->HdAudioDspUaaCompliance     = 1;
 
-  FspsConfig->GraphicsConfigPtr = PcdGet32 (PcdGraphicsVbtAddress);
+  if (PcdGetBool (PcdFramebufferInitEnabled)) {
```
gdong1: should it be PcdGetBool (PcdIntelGfxEnabled) here?

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -614,8 +656,8 @@ BoardInit (
   case PostSiliconInit:
     // To prevent from generating MCA for CLFLUSH flash region
     AsmMsrAnd32 (IA32_MC4_CTL, (UINT32)~BIT4);
-    // Enable GFX PCI command register if splash screen is required.
-    if (PcdGetBool (PcdSplashEnabled)) {
+    // Enable GFX PCI command register if framebuffer init is required.
+    if (PcdGetBool (PcdFramebufferInitEnabled)) {
```
gdong1: PcdGetBool (PcdIntelGfxEnabled) && PcdGetBool (PcdFramebufferInitEnabled)?

mauricema: I had thought about it.  The platform config itself should deal with the PCD dependency settings properly.   EX:  if Splash is enabled or FramebuferInit is enabled or Graphics console output is enabled,    then the IntelGfx must be enabled.  We cannot check all the dependencies everywhere.  I would like to centralize the check in platform config instead.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: GFX device enable is to determin if the PCI device should be visible on PCI bus.   FramebufferInit is to determine if the GFX should be initialized by FSP.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
gdong1: Got it. looks good to me.

mborgerson:
Looks good, thanks for the changes!

PR #27, opened 2018-10-26
Add shell command to print contents of log buffer

mborgerson:
This patch adds a new shell command `log` to print out the contents of
the boot log buffer.

### DIFF
--- a/BootloaderCommonPkg/Include/Library/ShellLib.h
+++ b/BootloaderCommonPkg/Include/Library/ShellLib.h
@@ -84,6 +84,7 @@ extern CONST SHELL_COMMAND ShellCommandPerf;
 extern CONST SHELL_COMMAND ShellCommandBoot;
 extern CONST SHELL_COMMAND ShellCommandMmcDll;
 extern CONST SHELL_COMMAND ShellCommandCdata;
+extern CONST SHELL_COMMAND ShellCommandDmesg;
 
 #define SHELL_COMMANDS_STANDARD \
   &ShellCommandHob, \
@@ -93,6 +94,7 @@ extern CONST SHELL_COMMAND ShellCommandCdata;
   &ShellCommandBoot, \
   &ShellCommandMmcDll, \
   &ShellCommandCdata, \
+  &ShellCommandDmesg, \
 
 // x86 Architectural Commands
 extern CONST SHELL_COMMAND ShellCommandCpuid;
--- a/BootloaderCommonPkg/Library/ShellLib/CmdDmesg.c
+++ b/BootloaderCommonPkg/Library/ShellLib/CmdDmesg.c
@@ -1,0 +1,107 @@
+/** @file
+  Shell command `dmesg` to print the contents of the log buffer.
+
+  Copyright (c) 2018, Intel Corporation. All rights reserved.<BR>
+  This program and the accompanying materials
+  are licensed and made available under the terms and conditions of the BSD License
+  which accompanies this distribution.  The full text of the license may be found at
+  http://opensource.org/licenses/bsd-license.php.
+
+  THE PROGRAM IS DISTRIBUTED UNDER THE BSD LICENSE ON AN "AS IS" BASIS,
+  WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
+
+**/
+
+#include <Library/ShellLib.h>
+#include <Library/DebugLib.h>
+#include <Library/HobLib.h>
+#include <Guid/LoaderFspInfoGuid.h>
+#include <Library/ConsoleInLib.h>
+#include <Library/ConsoleOutLib.h>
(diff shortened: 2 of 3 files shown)


BootloaderCommonPkg/Library/ShellLib/CmdLog.c
```diff
@@ -0,0 +1,104 @@
...
+EFI_STATUS
+EFIAPI
+ShellCommandLogFunc (
+  IN SHELL  *Shell,
+  IN UINTN   Argc,
+  IN CHAR16 *Argv[]
+  );
+
+CONST SHELL_COMMAND ShellCommandLog = {
+  L"log",
```
mauricema: should we use "dmsg" so that it is similar to Linux ?

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
```diff
@@ -0,0 +1,104 @@
...
+EFIAPI
+ShellCommandLogFunc (
+  IN SHELL  *Shell,
+  IN UINTN   Argc,
+  IN CHAR16 *Argv[]
+  )
+{
+  DEBUG_LOG_BUFFER_HEADER *LogBufHdr;
+  CONST UINTN              LinesPerPage = 30;
+  UINTN                    PageLineCount, Index;
```
mauricema: Split the declaration into two lines per coding style.

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
```diff
@@ -0,0 +1,104 @@
...
+EFI_STATUS
+EFIAPI
+ShellCommandLogFunc (
+  IN SHELL  *Shell,
+  IN UINTN   Argc,
+  IN CHAR16 *Argv[]
+  )
+{
+  DEBUG_LOG_BUFFER_HEADER *LogBufHdr;
+  CONST UINTN              LinesPerPage = 30;
```
mauricema: Should we define a PCD for this? 
Ideally it should match the console row number dynamically.

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
mborgerson: Ok

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
mborgerson: I prefer we keep it like this for now and add / replace this with a generic paging routine in console lib in a follow-up PR which can handle single line feeding and page down functions. I find it useful to use fewer than total screen rows so the reader has prior context around the new messages.

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
mborgerson: Sure, that will probably be more familiar to users. I'll change it.

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
mauricema: OK, I am OK with it.  
To conform to coding style, can you move this const into global const variable instead ?

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
```diff
@@ -0,0 +1,104 @@
...
+
+  PageLineCount = 0;
+  LogBufHdr = (DEBUG_LOG_BUFFER_HEADER *) GetDebugLogBufferPtr ();
+  for (Index = 0; Index < (LogBufHdr->UsedLength - LogBufHdr->HeaderLength); Index++) {
+    ConsoleWrite ((UINT8 *)&LogBufHdr->Buffer[Index], 1);
+
+    // Page out the log contents if requested
+    if (Paged && (LogBufHdr->Buffer[Index] == '\n') && (++PageLineCount == LinesPerPage)) {
+      ShellPrint (L"[Press any key to continue...]");
+      ConsoleRead (Buf, 1);
```
gdong1: Could we support exiting the command by pressing a specific key (e.g. by pressing ESC key)? this way, we don't need press keys all the way to the end of log.

mborgerson: Good idea. Added!

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
mborgerson: Fixed

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
mborgerson: Fixed

BootloaderCommonPkg/Library/ShellLib/CmdLog.c
mborgerson: Fixed

PR #31, opened 2018-10-29
Enable USB keyboard console support for QEMU

### LINKED ISSUE #30: USB keyboard console does not work reliably in QEMU
When trying to enable USB keyboard console for QEMU with Slim Bootloader, sometimes it works.  But it is not reliable, and lots of error message will be printed on console in QEMU.
For example:
XhcPeiCheckUrbResult: Transfer Default Error Occur! Completecode = 0x13! 
XhcPeiSetTrDequeuePointer: Set TR Dequeue Pointer Failed, Status = Device Error 
XhcPeiDequeueTrbFromEndpoint: Set Dequeue Pointer Failed, Status = Device Error 
XhcPeiBulkTransfer: XhcPeiDequeueTrbFromEndpoint failed

mauricema:
Previous USB keyboard console support commit worked fine on APL real
platform, but it has issue on QEMU. This patch further enables USB
keyboard console support for QEMU. A new PcdUsbKeyboardPollingTimeout
is added. It will be used to control the USB keyboard interrupt
transfer polling timeout. For QEMU, it needs a larger number due to
timing issue. As part of it, booting from USB device is also enabled
by this patch. This patch fixes #30.

To test USB keyboard console in QEMU, please first change
CONSOLE_IN _DEVICE_MASK in BoardConfig.py to 3, and then add the
following in the QEMU command line to add XHCI controller and USB KB:
-device qemu-xhci,id=xhci,bus=pcie.0,addr=4 -device usb-kbd,bus=xhci.0


BootloaderCommonPkg/Library/UsbBusLib/UsbIoPeim.c
```diff
@@ -194,6 +195,14 @@ PeiUsbBulkTransfer (
     return EFI_INVALID_PARAMETER;
   }
 
+  if ((PeiUsbDev->EndpointDesc[EndpointIndex]->Attributes & USB_ENDPOINT_TYPE_MASK) == USB_ENDPOINT_INTERRUPT) {
+    // Since USB HC PPI has no interrupt transfer support, reuse bulk transfer intead.
```
gdong1: a typo in comments: intead should be instead.

mauricema: will fix it.

mauricema: updated the comments.

Platform/QemuBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -340,6 +340,17 @@ UpdateOsBootMediumInfo (
   OsBootOptionList->OsBootOptionCount++;
   BootOption++;
 
+  // Boot file image from USB
```
mborgerson: I think it would be better to decouple the the USB keyboard logic from USB image boot structure. Even if they are the same USB controller, I think users may find this confusing.

mauricema: This requires a generic mechanism to provide device map information.   An initial thought, we could provide a system device mapping table which can be consumed by many drivers.

mborgerson: @mauricema I think that's a good idea! Perhaps something akin to Device Trees.

PR #36, opened 2018-11-01
Add basic SMM rebasing support

mauricema:
SBL has no intention to support SMI. However, on many hardware
platform, there is no reliable way to prevent SMI from generating
through SMI SPI.  In case it occurs, CPU will jump to the default
0x38000 location for execution, which exposes huge security issues.

The recommended solution is to do basic SMM base relocation and put
a dummy SMI handler (RSM) there. In this way, the SMI will be ignored,
and it also closes the security concerns. This patch implemented
basic SMM relocation.

It is under the control of a new PCD PcdSmmRebaseEnabled. By default,
it is disabled.  To enable it, please set ENABLE_SMM_REBASE in
BoardConfig.py. As part of it, platform library needs to set
PcdSmramTsegBase and PcdSmramTsegSize properly in PreSiliconInit board
hook. Please take APL platform for reference.

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dec
+++ b/BootloaderCorePkg/BootloaderCorePkg.dec
@@ -117,14 +117,21 @@
   gPlatformModuleTokenSpaceGuid.PcdSeedListBufferSize     | 0x00000400 | UINT32 | 0x200000E2
 
 [PcdsFixedAtBuild, PcdsPatchableInModule]
-  # For patchable PCDs, try to set the default as none-zero
-  # It is to prevent it from being put into BSS section, thus cause patching issue
+  #
+  # For module patchable PCDs, if it is required to do static patching during the build
+  # process using PatchFv.py script, then they should be set to non-zero default values.
+  # It is to prevent it from being put into BSS section, thus cause issue during pacting
+  # because the PCD symbol won't be available in MAP file.
+  # If static patching is not required, zero default value can be used with no issue.
+  #
   gPlatformModuleTokenSpaceGuid.PcdFSPSBase               | 0xFF000000 | UINT32 | 0x20000100
   gPlatformModuleTokenSpaceGuid.PcdAcpiTablesMaxEntry     |         32 | UINT32 | 0x2000000C
   gPlatformModuleTokenSpaceGuid.PcdAcpiTablesRsdp         | 0x00000000 | UINT32 | 0x2000000D
   gPlatformModuleTokenSpaceGuid.PcdAcpiTablesAddress      | 0xFF000000 | UINT32 | 0x20000110
   gPlatformModuleTokenSpaceGuid.PcdAcpiGnvsAddress        | 0xFF000000 | UINT32 | 0x20000112
   gPlatformModuleTokenSpaceGuid.PcdGraphicsVbtAddress     | 0xFF000000 | UINT32 | 0x20000113
+  gPlatformModuleTokenSpaceGuid.PcdSmramTsegBase          | 0xFF000000 | UINT32 | 0x20000120
+  gPlatformModuleTokenSpaceGuid.PcdSmramTsegSize          | 0x00000000 | UINT32 | 0x20000121
   gPlatformModuleTokenSpaceGuid.PcdKeyStoreBase           | 0xFF000000 | UINT32 | 0x20000181
   gPlatformModuleTokenSpaceGuid.PcdDeviceTreeBase         | 0xFF000000 | UINT32 | 0x20000183
   gPlatformModuleTokenSpaceGuid.PcdSplashLogoAddress      | 0xFF000000 | UINT32 | 0x20000184
@@ -153,4 +160,4 @@
   gPlatformModuleTokenSpaceGuid.PcdFlashMapEnabled        | FALSE      | BOOLEAN | 0x2000020C
   gPlatformModuleTokenSpaceGuid.PcdS3DebugEnabled         | FALSE      | BOOLEAN | 0x2000020D
   gPlatformModuleTokenSpaceGuid.PcdPsdBiosEnabled         | FALSE      | BOOLEAN | 0x2000020E
-
+  gPlatformModuleTokenSpaceGuid.PcdSmmRebaseEnabled       | FALSE      | BOOLEAN | 0x2000020F
--- a/BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
+++ b/BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
@@ -12,9 +12,9 @@
 ; Module Name:
 ;
 ;   MpFuncs.asm
-; 
+;
 ; Abstract:
-; 
+;
 ;   This is the assembly code for initializing the APs in IA32 Protected Mode
 ;
 ;-------------------------------------------------------------------------------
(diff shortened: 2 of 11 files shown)


BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
```diff
@@ -23,23 +23,32 @@ SECTION .text
...
 RendezvousFunnelProcStart:
 ; At this point CS = 0x(vv00) and ip= 0x0.
-  
+  db 0fh,  20h, 0d0h            ; mov        eax, cr2
+  db 66h,  09h, 0c0h            ; or         eax, eax
+
+  ; For SMM rebase flow, CR2 will be pre-filled with SMBASE
+  jz StartIpi
+  db 0eah                       ; jmp        0x3800:StartIpi
+  dw StartIpi - RendezvousFunnelProcStart
```
rprangar: Maybe a comment on why the jump is necessary. When CPU enters SMM, the CS will be initialized to 0x3000 and the code is written to use a base of 0x3800.

BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
```diff
@@ -145,10 +184,10 @@ MpInit (
       AsmGetHotAddCodeAddressMap (&mAddressMap);
```
rprangar: Can we rename this function?

BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
```diff
@@ -145,10 +184,10 @@ MpInit (
...
 
       //
-      // Allocate backup Buffer
-      //
-      mBackupBuffer = AllocatePages (EFI_SIZE_TO_PAGES (mAddressMap.CodeSize + mAddressMap.MpDataSize));
-      CopyMem (mBackupBuffer, ApBuffer, (mAddressMap.CodeSize + mAddressMap.MpDataSize));
+      // Allocate backup Buffer fpr MP waking up
```
rprangar: typo in "for"

BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
mauricema: Sure, will add that.

BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
mauricema: OK.  I can rename it.  It is not part of my original code change.

BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
mauricema: will fix typo

mauricema:
Updated the patch per code review comments

BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
```diff
@@ -23,23 +23,36 @@ SECTION .text
...
 SPIN_LOCK_ACQUIRED   EQU   1
 
- 
+
 global ASM_PFX(RendezvousFunnelProc)
 ASM_PFX(RendezvousFunnelProc):
 RendezvousFunnelProcStart:
 ; At this point CS = 0x(vv00) and ip= 0x0.
-  
+  db 0fh,  20h, 0d0h            ; mov        eax, cr2
```
mborgerson: Why not use the real instructions here?

BootloaderCorePkg/BootloaderCorePkg.dec
```diff
@@ -125,6 +125,8 @@
   gPlatformModuleTokenSpaceGuid.PcdAcpiTablesAddress      | 0xFF000000 | UINT32 | 0x20000110
   gPlatformModuleTokenSpaceGuid.PcdAcpiGnvsAddress        | 0xFF000000 | UINT32 | 0x20000112
   gPlatformModuleTokenSpaceGuid.PcdGraphicsVbtAddress     | 0xFF000000 | UINT32 | 0x20000113
+  gPlatformModuleTokenSpaceGuid.PcdSmramTsegBase          | 0xFF000000 | UINT32 | 0x20000120
+  gPlatformModuleTokenSpaceGuid.PcdSmramTsegSize          | 0x00000000 | UINT32 | 0x20000121
```
mborgerson: Comment above mentions not to have default value of 0 here as it may cause the PCD to be put into the BSS section.

BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
mauricema: At this point, the CPU is still in real mode.  But the assembler will try to create 32bit instruction which might cause compatibility issue.  That is why the opcode is used directly until the CPU is fully in 32bit flat mode.

BootloaderCorePkg/BootloaderCorePkg.dec
mauricema: That is good point.  
But it is OK to keep the PCD in BSS section as long as we don't do static patching using PatchFv.py because the symbol won't be generated in MAP file.   So I will update the comment to make it more clear.

mauricema:
Updated per Matt's comments.

BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
```diff
@@ -70,13 +83,19 @@ FLAT32_JUMP:
   dd 0h                         ; 32-bit offset
   dw 0h                         ; 16-bit selector
 
-ProtModeStart:                 ; protected mode entry point
+SmmRebase:
+  mov         edi,  0x3fef8     ; SMBASE offset
+  mov         dword [edi], eax  ; change to new  SMBASE
+  mov         dword [eax + 0x8000], 0x9090AA0F  ; write 'RSM' at new SMM handler entry
+  rsm
+  jmp         $-2
```
myguitar: This is for infinite rsm call just in case?

mauricema: This is to change the default SMBASE from 0x30000 to new location in TSEG.   And at the new SMM entry location, we put a "RSM" as the 1st instruction.

myguitar: My question is about a line - jmp $-2. It shouldn't be reachable.

BootloaderCorePkg/Library/MpInitLib/Ia32/MpFuncs.nasm
mborgerson: NASM supports assembling/emitting 16-bit instructions if you specify "bits 16" just before the instructions (then back to 32b with a "bits 32" line). I think this is fine for now but we might want to change it to make it more maintainable down the road.

mauricema: Yes, agree.   I can try to explore a little bit on this to see if it can be changed to use mixed ASM code.  If so, we can do it in a new PR.

mauricema: BTW, submitted a new PR #37 to convert opcode into instructions.

mborgerson: Awesome, thx @mauricema!

PR #40, opened 2018-11-08
Fix a compile error when VTD is disabled on APL

myguitar:
When VTD_ENABLE sets to '0', a compile error occurs because some .inf
files refer to VtdLib. Therefore, VtdLib is included to default
libraries - dsc_libs['IA32'].

Additionaly,
- Rename VTD_ENABLED to ENABLE_VTD to make feature enabling name
  consistent
- Remove hard-coded UART2 BAR for ACRN

Change-Id: I22969c0e4509ab0a0dc21d50e6afbdfb893f25a8

### DIFF
--- a/Platform/ApollolakeBoardPkg/BoardConfig.py
+++ b/Platform/ApollolakeBoardPkg/BoardConfig.py
@@ -1,7 +1,7 @@
 ## @file
 # This file is used to provide board specific image information.
 #
-#  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
+#  Copyright (c) 2017 - 2018, Intel Corporation. All rights reserved.<BR>
 #
 #  This program and the accompanying materials
 #  are licensed and made available under the terms and conditions of the BSD License
@@ -152,7 +152,7 @@ class Board(BaseBoard):
 			'ShellExtensionLib|Platform/$(BOARD_PKG_NAME)/Library/ShellExtensionLib/ShellExtensionLib.inf',
 			'BootMediaLib|Silicon/ApollolakePkg/Library/BootMediaLib/BootMediaLib.inf',
 			'FlashDescriptorLib|Silicon/ApollolakePkg/Library/FlashDescriptorLib/FlashDescriptorLib.inf',
-			'VTdLib|Silicon/$(SILICON_PKG_NAME)/Library/VTdLib/VTdLib.inf'
+			'VtdLib|Silicon/$(SILICON_PKG_NAME)/Library/VtdLib/VtdLib.inf'
 		]
 		return dsc_libs
 
--- a/Platform/ApollolakeBoardPkg/BoardConfig.py
+++ b/Platform/ApollolakeBoardPkg/BoardConfig.py
@@ -52,7 +52,7 @@ class Board(BaseBoard):
 
 		self.ENABLE_FSP_LOAD_IMAGE    = 0
 		self.ENABLE_CRYPTO_SHA_NI     = 1
-		self.VTD_ENABLED              = 1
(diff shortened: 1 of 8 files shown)


huang-jin-intel:
For commit messages, can we follow conventions for changes that are board / platform specific? E.g.

* [APL]
* [UP2]
* [QEMU]
...

Platform/ApollolakeBoardPkg/BoardConfig.py
```diff
@@ -152,10 +152,8 @@ def GetDscLibrarys (self, BuildPkgName = "BootLoaderCorePkg"):
 			'ShellExtensionLib|Platform/$(BOARD_PKG_NAME)/Library/ShellExtensionLib/ShellExtensionLib.inf',
 			'BootMediaLib|Silicon/ApollolakePkg/Library/BootMediaLib/BootMediaLib.inf',
 			'FlashDescriptorLib|Silicon/ApollolakePkg/Library/FlashDescriptorLib/FlashDescriptorLib.inf',
+			'VTdLib|Silicon/$(SILICON_PKG_NAME)/Library/VTdLib/VTdLib.inf'
```
huang-jin-intel: Why VTd ('T') is uppercase in naming? Maybe clean it up too?

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -410,34 +410,6 @@ AssignPciIrqs (
...
-    PciBase = MmPciBase (DEFAULT_PCI_BUS_NUMBER_SC, \
-                         PCI_DEVICE_NUMBER_LPSS_HSUART, \
-                         (PCI_FUNCTION_NUMBER_LPSS_HSUART0 + 2));
-
-    MmioAnd32   (PciBase + PCI_COMMAND_OFFSET, (UINT32)~ (EFI_PCI_COMMAND_MEMORY_SPACE));
-    MmioWrite32 (PciBase + PCI_BASE_ADDRESSREG_OFFSET, Base);
-    MmioOr32    (PciBase + PCI_COMMAND_OFFSET, EFI_PCI_COMMAND_MEMORY_SPACE);
-  }
-}
-
```
huang-jin-intel: If you remove this unused function here, it would mean ACRN may not able to work anymore. Can we keep it for now?

Platform/ApollolakeBoardPkg/BoardConfig.py
myguitar: It's VT-d actually. I think changing to Vtd would require directory name change as well, so not that much benefit.

huang-jin-intel: The camel case really not very expressive but I understand now.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
huang-jin-intel: I understand.

Platform/ApollolakeBoardPkg/BoardConfig.py
mauricema: According to EDKII coding style,  it should be VtdLib and VtdLib.inf.  :)

myguitar: Okay, coding convention. Let me change whole VTd to Vtd including directories.

myguitar:
Added one more commit to rename VTd to Vtd.

PR #47, opened 2018-11-21
Update file system support

gdong1:
The patch gives the platform the configuration capability to only build
required file system into final image.
This helps to have a smaller image for fast boot.


BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h
```diff
@@ -71,8 +71,8 @@ typedef enum {
 typedef enum  {
   EnumFileSystemTypeFat,
   EnumFileSystemTypeExt2,
-  EnumFileSystemTypeAuto,
-  EnumFileSystemMax
+  EnumFileSystemMax,
+  EnumFileSystemTypeAuto = (1 < EnumFileSystemTypeFat) | (1< EnumFileSystemTypeExt2)
```
mauricema: I think it might be be better to use the bits instead to make it straightforward.
typedef enum  {
  EnumFileSystemTypeFat   = BIT0,
  EnumFileSystemTypeExt2 = BIT1,
  EnumFileSystemTypeAll    = BIT0 | BIT1,
  EnumFileSystemMax	    = EnumFileSystemTypeAll + 1
}

Max should be the last item.

BootloaderCommonPkg/Library/FileSystemLib/FileSystemLib.inf
```diff
@@ -42,3 +42,5 @@
   FatLib
   Ext23Lib
 
+[FixedPcd]
```
mauricema: why not use [Pcd] instead

BootloaderCommonPkg/Library/FileSystemLib/FileSystemLib.c
```diff
@@ -83,17 +83,32 @@ InitFileSystem (
...
+      mFileSystemFuncs[Type].GetFileByName      = ExtGetFileByName;
+    }
   }
 
-  if (EFI_ERROR (Status)) {
-    if ((FsType == EnumFileSystemTypeExt2) || (FsType == EnumFileSystemTypeAuto)) {
-      Status = ExtInitFileSystem (SwPart, PartHandle, FsHandle);
+  for (Type = EnumFileSystemTypeFat; Type < EnumFileSystemMax; Type++) {
+    if (((FsType & Type) != 0) && (mFileSystemFuncs[Type].InitFileSystem != NULL)) {
+      mCurrentFsType = Type;
```
mauricema: better to move this to after line 100 so that the init will only do one time.

BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h
mauricema: Sorry,  I meant the style needs to be consistent.  If sequential number is used, then bit mask should not be used at the same time.    You can also use something like below:
ypedef enum {
EnumFileSystemTypeFat,
EnumFileSystemTypeExt2,
EnumFileSystemTypeAuto,
EnumFileSystemMax
}

gdong1: Thanks for the comments, will update the code

BootloaderCommonPkg/Library/FileSystemLib/FileSystemLib.c
gdong1: will update the code

PR #52, opened 2018-12-04
Clean-up build clean/distclean command

myguitar:
When BaseTools is not ready, 'python BuildLoader.py clean' initiates
building BaseTools even if it is 'clean' command. Therefore, move
rebuild_basetools() into pre_build().

Additionaly, all unstaging files in BaseTools directory will be cleared
to re-compile BaseTools when 'distclean' is executed.


BuildLoader.py
```diff
@@ -1222,13 +1222,24 @@ def cmd_clean(args):
 			files.extend ([
 			])
 
+			cmds.extend ([
```
huang-jin-intel: Do not assume this is a git directory.

PR #53, opened 2018-12-05
Set PYTHON_HOME env variable for Windows build

mauricema:
The latest SBL source code does not build on certain environment.
For example, when PYTHON_HOME is not set manually, it will cause
build failure due to invalid path for python.exe.  This patch will
set the PYTHON_HOME env variable if it has not been set already.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -44,12 +44,12 @@ def rebuild_basetools ():
 			ret = subprocess.call(['make', '-C', 'BaseTools'])
 
 	elif os.name == 'nt':
+		if 'PYTHON_HOME' not in os.environ:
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -44,12 +44,12 @@ def rebuild_basetools ():
 			ret = subprocess.call(['make', '-C', 'BaseTools'])
 
 	elif os.name == 'nt':
+		if 'PYTHON_HOME' not in os.environ:
+			os.environ['PYTHON_HOME'] = 'C:\\Python27'
```
mborgerson: Assumes this directory exists. I think this is a sane default however.

mauricema: The problem is PYTHON_HOME might not have been set by default.  It was my case on my platform.

PR #54, opened 2018-12-05
[APL] Make EPAYLOAD size dynamic

mauricema:
On APL platform, default EPAYLOAD size is more than 1MB to enable
UEFI payload. However, if UEFI payload is not used, the space is
still reserved in flash. A more flexible way is to reserve the
EPAYLOAD size only when it is specified in the build command line.
This patch added this capability.

### DIFF
--- a/Platform/ApollolakeBoardPkg/BoardConfig.py
+++ b/Platform/ApollolakeBoardPkg/BoardConfig.py
@@ -95,7 +95,14 @@ class Board(BaseBoard):
 			self.STAGE1B_SIZE += 0x2000
 		self.STAGE2_SIZE          = 0x00032000
 		self.PAYLOAD_SIZE         = 0x0001F000
-		self.EPAYLOAD_SIZE        = 0x00120000
+
+		if len(self._PAYLOAD_NAME.split(';')) > 1:
+			# EPAYLOAD is specified
+			self.EPAYLOAD_SIZE      = 0x00120000
+		else:
+			# EPAYLOAD does not exist, create a dummy one
+			self.EPAYLOAD_SIZE      = 0x1000
+
 		if self.FSPDEBUG_MODE == 1:
 			self.STAGE1B_SIZE += 0x00009000
 			self.STAGE2_SIZE  += 0x0000F000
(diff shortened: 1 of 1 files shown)


Platform/ApollolakeBoardPkg/BoardConfig.py
```diff
@@ -95,7 +95,14 @@ def __init__(self, *args, **kwargs):
...
 		self.STAGE2_SIZE          = 0x00032000
 		self.PAYLOAD_SIZE         = 0x0001F000
-		self.EPAYLOAD_SIZE        = 0x00120000
+
+		if len(self._PAYLOAD_NAME.split(';')) > 1:
+			# EPAYLOAD is specified
+			self.EPAYLOAD_SIZE      = 0x00120000
+		else:
+			# EPAYLOAD does not exist, create a dummy one
+			self.EPAYLOAD_SIZE      = 0x1000
```
myguitar: If dummy EPAYLOAD is required for stitching tool, why don't you remove 'if self.EPAYLOAD_SIZE > 0:' in GetImageLayout()?

mauricema: good point.  let me remove that.

mauricema: updated the patch to remove it.

Platform/ApollolakeBoardPkg/BoardConfig.py
```diff
@@ -95,7 +95,14 @@ def __init__(self, *args, **kwargs):
 			self.STAGE1B_SIZE += 0x2000
 		self.STAGE2_SIZE          = 0x00032000
 		self.PAYLOAD_SIZE         = 0x0001F000
-		self.EPAYLOAD_SIZE        = 0x00120000
+
+		if len(self._PAYLOAD_NAME.split(';')) > 1:
+			# EPAYLOAD is specified
+			self.EPAYLOAD_SIZE      = 0x00120000
```
huang-jin-intel: The title says _dynamic_ size, but we are allocating a _maxium_ size. Maybe revise the commit message?

mauricema: It is dynamic according to the payload argument (use epayload or not).

PR #56, opened 2018-12-12
[APL] Adjust flash layout padding for stitching

mauricema:
SBL supports open source stitching mechanism based on an existing
IFWI binary. It can also support FIT stitching mechanism. For FIT
stitching, it is required to add certain padding in the layout so
that each component in IBB and OBB partition can be 4KB aligned.
Current padding component IPAD/OPAD size needs to be updated to
match the number of the component inside IBB and OBB partition.
This patch adjusted the padding size accordingly.

### DIFF
--- a/Platform/ApollolakeBoardPkg/BoardConfig.py
+++ b/Platform/ApollolakeBoardPkg/BoardConfig.py
@@ -209,14 +209,14 @@ class Board(BaseBoard):
 		img_list  = []
 		img_list.extend ([
 					# Padding to ensure all other components in OBB partition will be aligned at 4KB boundary
-					# 0xB80 assumes 7 files in BpdtIBB (IBBP.man, BPM.met) + (IPAD, IBBL, IBBM,  OBB, CFGDATA)
-					# 0xF40 assumes 5 files in BpdtOBB (OPAD,  PLD,  VAR, MRCD, PROV)
+					# 0xB00 assumes (IBBP.man, BPM.met) + (IPAD, IBBL, IBBM,  OBB, FWUP, CFGD, PLD, VAR, MRCD) in BpdtIBB
+					# 0x180 assumes (OPAD,  PROV, EPLD) in BpdtOBB 
 					# If more files are added, the offset needs to be adjusted accordingly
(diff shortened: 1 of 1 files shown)


Platform/ApollolakeBoardPkg/BoardConfig.py
```diff
@@ -209,14 +209,14 @@ def GetImageLayout (self):
 		img_list  = []
 		img_list.extend ([
 					# Padding to ensure all other components in OBB partition will be aligned at 4KB boundary
-					# 0xB80 assumes 7 files in BpdtIBB (IBBP.man, BPM.met) + (IPAD, IBBL, IBBM,  OBB, CFGDATA)
-					# 0xF40 assumes 5 files in BpdtOBB (OPAD,  PLD,  VAR, MRCD, PROV)
+					# 0xB00 assumes (IBBP.man, BPM.met) + (IPAD, IBBL, IBBM,  OBB, FWUP, CFGD, PLD, VAR, MRCD) in BpdtIBB
+					# 0x180 assumes (OPAD,  PROV, EPLD) in BpdtOBB
```
huang-jin-intel: How is the actual padding size calculated based on the number of files?

mauricema: Unfortunately there is no clear published document  to know that since the alignment is arranged internally by the FITC tool.  Current approach is to try FITC stitching, and then check the layout alignment in the final IFWI.  From the offset, we will know the gap to pad.  Using this information we can decide the IPAD/OPAD size.   
This is only used by FITC stitching.  For open source stitching, the layout is fully controlled by the LoaderStitch.py script.  IPAD/OPAD will not be used at all.

huang-jin-intel:
> IFWI binary. It can also support FITC stitching mechanism. For FITC

To be accurate, it is FIT tool instead of FITC tool.

> stitching, it is required to add certain padding in the layout so
> that each component in IBB and OBB partition can be 4KB aligned.

What is the consequence without this change?

mauricema:
Updated the commit message.

PR #61, opened 2019-01-03
Fix firmware update bug to load capsule from USB

huang-jin-intel:
File system is not initialized when invalid parameter is
given to InitFileSystem(). This bug causes FWU payload
fails to load capsule using shell command fwupdate.

Signed-off-by: Huang Jin <huang.jin@intel.com>

### DIFF
--- a/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
+++ b/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
@@ -271,7 +271,7 @@ LoadCapsuleImage (
   }
 
   FsHandle = NULL;
-  Status = InitFileSystem (CapsuleInfo->SwPart, EnumFileSystemTypeFat, HwPartHandle, &FsHandle);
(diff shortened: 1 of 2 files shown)


PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -271,7 +271,7 @@ LoadCapsuleImage (
   }
 
   FsHandle = NULL;
-  Status = InitFileSystem (CapsuleInfo->SwPart, EnumFileSystemTypeFat, HwPartHandle, &FsHandle);
+  Status = InitFileSystem (CapsuleInfo->SwPart, EnumFileSystemTypeAuto, HwPartHandle, &FsHandle);
```
gdong1: should we use CapsuleInfo->FsType instead of hardcode?

huang-jin-intel: The patch now reworked to fix the real bug in FileSystem library.  Will cleanup and test API consumer code in follow-up changes.

PR #68, opened 2019-01-08
[APL] Enlarge payload heap size from 64MB to 128MB

huang-jin-intel:
This change makes more space for SBL paylaod to load
larger IAS image (e.g. Ubuntu kernel + initrd ~ 33MB)

The payload heap memory is reusable as long as crash mode
is disabled.

TEST=Loaded and booted IAS image containing Ubuntu kernel and
     initrd from USB flash drive on UP2 board.

Signed-off-by: Huang Jin <huang.jin@intel.com>


huang-jin-intel:
@mauricema This is a simple change I have tested with Ubuntu to support w/ IAS flow

PR #74, opened 2019-01-26
Fix firmware update failure

raghavag:
When UEFI payload or epayload is enabled and firmware update mode
is enabled, current code in prepare payload function will consider
firmware update payload as multi payload. This is causing failure
and end up as halting cpu.

This patch will add an additional check for firmware update mode
and do hash verification

TEST=Verified that firmware update is working when UEFI payload
	 is enabled.


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -119,8 +119,9 @@ PreparePayload (
       }
     }
 
-    if (IsNormalPld || (Stage2Hob->PayloadId != 0)) {
+    if (IsNormalPld || (Stage2Hob->PayloadId != 0) || (BootMode == BOOT_ON_FLASH_UPDATE)) {
```
mauricema: I think it is better to set IsNomralPld to TRUE when firmware update is detected.

PR #79, opened 2019-01-31
Refactor LiteFvLib/LitePeCoffLib implementation

mauricema:
This patch did some clean up for LiteFvLib and LitePeCoffLib.  It
also moved LoadFvImage() API from Stage2 core code into the LiteFvLib.

### DIFF
--- a/BootloaderCommonPkg/Include/Library/LitePeCoffLib.h
+++ b/BootloaderCommonPkg/Include/Library/LitePeCoffLib.h
@@ -2,7 +2,7 @@
   Base PE/COFF loader supports loading any PE32/PE32+ or TE image, but
   only supports relocating IA32, x64, IPF, and EBC images.
 
-  Copyright (c) 2006 - 2008, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2006 - 2019, Intel Corporation. All rights reserved.<BR>
   Portions copyright (c) 2008-2009 Apple Inc. All rights reserved.<BR>
   This program and the accompanying materials
   are licensed and made available under the terms and conditions of the BSD License
@@ -43,8 +43,6 @@ IsTePe32Image (
   Retrieves the entry point to the PE/COFF image specified by Pe32Data and returns this entry
   point in EntryPoint.  If the entry point could not be retrieved from the PE/COFF image, then
   return RETURN_INVALID_PARAMETER.  Otherwise return RETURN_SUCCESS.
-  If Pe32Data is NULL, then ASSERT().
-  If EntryPoint is NULL, then ASSERT().
 
   @param  Pe32Data                  The pointer to the PE/COFF image that is loaded in system memory.
   @param  EntryPoint                The pointer to entry point to the PE/COFF image to return.
@@ -60,19 +58,6 @@ PeCoffLoaderGetEntryPoint (
   OUT VOID  **EntryPoint
   );
 
-/**
-  Performs an specific relocation fpr PECOFF images. The caller needs to
-  allocate enough buffer at the PreferedImageBase
-
-  @param  CurrentImageBase        Pointer to the current image base.
-
-  @return Status code.
-
-**/
-RETURN_STATUS
-PeCoffRelocateImage (
-  IN     UINT32   CurrentImageBase
-  );
 
 /**
   Performs an specific relocation fpr PECOFF images. The caller needs to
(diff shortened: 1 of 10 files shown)


BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
```diff
@@ -56,6 +81,10 @@ GetFfsFileByName (
   // Get the first file
   //
   CurrentFile = (EFI_FFS_FILE_HEADER *) ((UINTN)FvHeader + FvHeader->HeaderLength);
+  if (FvHeader->ExtHeaderOffset != 0) {
+    FvExHeader  = (EFI_FIRMWARE_VOLUME_EXT_HEADER *)(((UINT8 *)FvHeader) + FvHeader->ExtHeaderOffset);
```
mborgerson: May want to put this in a separate function as it's duplicated below

BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
```diff
@@ -87,6 +116,95 @@ GetFfsFileByName (
...
+        *File = CurrentFile;
+        return EFI_SUCCESS;
+      }
+      Count++;
+    }
+
+    //
+    // Get next file, compensate for 8 byte alignment if necessary.
+    //
+    NextFile = (EFI_FFS_FILE_HEADER *) ((((UINTN)CurrentFile - (UINTN)FvHeader + \
```
mborgerson: Can we simplify this to something like `ALIGN_UP(CurrentFile + GET_FFS_LENGTH(CurrentFile), 8)`

BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
```diff
@@ -161,3 +275,70 @@ GetSectionByType (
...
+  SecCoreImageBase = (UINTN)Section;
+  Status = PeCoffGetPreferredBase ((VOID *)SecCoreImageBase, &PreferredBase);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  // Move FV to preferred base if required
+  Gap = (UINT32)(UINTN)PreferredBase - SecCoreImageBase;
+  if (Gap != 0) {
+    CopyMem ((UINT8 *)FvHeader + Gap, FvHeader, FvLength);
```
mborgerson: We should probably check for destination/source partial overlap in this case (or use something which does so like a `memmove`)

BootloaderCommonPkg/Include/Library/LitePeCoffLib.h
```diff
@@ -74,40 +73,40 @@ PeCoffRelocateImage (
...
+  Performs an specific relocation fpr PECOFF images. The caller needs to
+  allocate enough buffer at the PreferedImageBase
+
+  @param  CurrentImageBase        Pointer to the current image base.
+
+  @return Status code.
+
+**/
+RETURN_STATUS
+PeCoffRelocateImage (
```
gdong1: Add EFIAPI for this function.

BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
```diff
@@ -87,6 +116,95 @@ GetFfsFileByName (
...
+  EFI_FIRMWARE_VOLUME_EXT_HEADER  *FvExHeader;
+
+  //
+  // Verify library has been initialized.
+  //
+  if (FvHeader == NULL || File == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  if (FvHeader->Signature != EFI_FVH_SIGNATURE) {
```
gdong1: Using IsValidFvHeader () to check if it is a valid FV?

BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
gdong1: Prefer using ALIGN_POINTER (). e.g.   FfsFileHeader = ALIGN_POINTER (FfsFileHeader, 8);

mauricema:
Pushed new patch and it addressed all the comments.

BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
```diff
@@ -48,31 +105,97 @@ GetFfsFileByName (
...
     }
 
     //
     // Get next file, compensate for 8 byte alignment if necessary.
     //
-    NextFile = (EFI_FFS_FILE_HEADER *) ((((UINTN)CurrentFile - (UINTN)FvHeader + \
-                                          GET_FFS_LENGTH (CurrentFile) + 0x07) & (~ (UINTN)7)) + (UINTN)FvHeader);
-
+    Offset   = (UINTN)CurrentFile - (UINTN)FvHeader + GET_FFS_LENGTH (CurrentFile);
+    NextFile = (EFI_FFS_FILE_HEADER *)((UINTN)FvHeader + ALIGN_UP (Offset, 8));
```
gdong1: How about using "NextFile = (EFI_FFS_FILE_HEADER *) ALIGN_UP ((UINTN)CurrentFile + GET_FFS_LENGTH (CurrentFile), 8));" instead of 2 lines above?

mauricema: The logic is not the same.   The original logic does not assume the FV header is 16 byte aligned in the location.  It means the function will work even though the FV is misaligned.

gdong1: Make sense in case FvHeader is not in aligned address.

PR #81, opened 2019-02-05
Add duplicated CFGDATA check in GenCfgData.py

mauricema:
CFGDATA name should be unique in the whole CFGDATA database. Current
tool does not check this and will keep silent for this error. This
brings issue into later development stage. The build process should
be enhanced to check this and error out if duplicated CFGDATA item is
found. This patch added this check.  It fixed #81 .

### DIFF
--- a/BootloaderCorePkg/Tools/GenCfgData.py
+++ b/BootloaderCorePkg/Tools/GenCfgData.py
@@ -1538,6 +1538,32 @@ EndList
 
         return 0
 
+    def CheckCfgData (self):
+        # Check if CfgData contains any duplicated name
+        def  AddItem (Item, ChkList):
+            Name = Item['cname']
+            if Name in ChkList:
+                return Item
+            if Name not in ['Dummy', 'Reserved', 'CfgHeader', 'CondValue']:
+                ChkList.append(Name)
+            return None
+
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/GenCfgData.py
```diff
@@ -1538,6 +1538,32 @@ def GenerateDataIncFile (self, DatIncFileName, BinFile = None):
 
         return 0
 
+    def CheckCfgData (self):
+        # Check if CfgData contains any duplicated name
+        def  AddItem (Item, ChkList):
+            Name = Item['cname']
+            if Name in ChkList:
+                return Item
+            if Name not in ['Dummy', 'Reserved', 'CfgHeader', 'CondValue']:
```
stalamudupula: Can we ignore fields named 'Rsvd*' also? There might be multiple fields with name 'Rsvd' that might be a hit here.

mauricema: Let us scandalize the keyboards used in CFGDATA.  Rsvd* normally is for field in a sub region.  So it could be duplicated within different sub region scope.   I have verified the current change worked fine for QEMU and APL.

PR #86, opened 2019-02-07
SBL - Measured boot support in UEFI payload

lsubash:
Added support for TPM event log info HOB and updated Loader FSP info.


BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
```diff
@@ -382,4 +382,51 @@ TpmLogLocalityEvent (
...
+  UINT32                Laml;            //LogAreaMinimumLength
+  UINT32                EventSize;
+  TCG_PCR_EVENT2_HDR    *EmptySlot;
+  TCG_PCR_EVENT_HDR     *FirstEvent;
+
+  GetTCGLasa (&Lasa, &Laml);
+
+  if (Lasa == 0 || Laml == 0 ) {
+    DEBUG ((DEBUG_WARN, "Unable to get log area for TCG 2.0 format events !!\n"));
+    *Tcg2Lasa = 0;
```
mauricema: Is this function an external interface ?  If so, we need to check if Tcg2Lasa/Tcg2LasaEven is NULL.

BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
```diff
@@ -382,4 +382,51 @@ TpmLogLocalityEvent (
...
+  Get the TCG log Lasa and LasaEven info.
+
+  @param[out] Tcg2Lasa  Pointer to the Log area start address .
+  @param[out] Tcg2LasaEven  Pointer to the Last event logged.
+
+**/
+VOID
+GetTCGLogInfo (
+  OUT  UINT32  *Tcg2Lasa,
+  OUT  UINT32  *Tcg2LasaEven
```
mauricema: Should we use Tcg2LasaLastEvent instead according to the function comments ?

BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
lsubash: Yes, its an external interface. I would add NULL check for Tcg2Lasa and Tcg2LasaEven.

BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
lsubash: Yeah.. I could change to Tcg2LasaLastEvent instead of Tcg2LasaEven.

lsubash:
Thanks Sachin. I have added in UEFI payload functionality to retrieve TPM Event log info by accessing TPM2 ACPI table.  Able to get from events logged in Slim boot loader.

TPM Event log HOB need not be added to SBL.

BootloaderCorePkg/Stage2/Stage2Support.c
```diff
@@ -390,6 +390,9 @@ BuildBaseInfoHob (
   if (LoaderFspInfo != NULL) {
     LoaderFspInfo->FspsBase   = PCD_GET32_WITH_ADJUST (PcdFSPSBase);
     LoaderFspInfo->FspHobList = LdrGlobal->FspHobList;
+    LoaderFspInfo->VerifiedBoot = FeaturePcdGet (PcdVerifiedBootEnabled);
+    LoaderFspInfo->MeasuredBoot = FeaturePcdGet (PcdMeasuredBootEnabled);
```
gdong1: we could not just use these

















Even we build Verified boot and Measured boot (FeaturePcds are set to true in SBL), but the boot guard profile could disable them. should we use LOADER_PLATFORM_INFO->SecurityFlags instead?

lsubash: Sure, I would use interface LOADER_PLATFORM_INFO->SecurityFlags to update the security features.

lsubash:
- Alternate implementation is done in UEFI payload to get TCG Event log info by reading ACPI tables.
- using the interface LOADER_PLATFORM_INFO to update security flags.

These patches would create an new HOB's which are no longer required.

PR #89, opened 2019-02-11
Enhance build script for clean command

gdong1:
BuildLoader.py will failure to run git command if Slim bootloader
source code is not in a git repo. Update Buildloader.py to check
if git repo could not be detected, it will not run git command to
clean un-tracked git files.

TEST=Tested to build/clean APL success without git repo.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -1218,7 +1218,6 @@ def main():
 			os.path.join (sbl_dir, 'BootloaderCorePkg/Platform.dsc'),
 			os.path.join (workspace, 'Report.log')
 		]
-		cmds  = []
 
 		if args.distclean:
 			dirs.extend ([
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -1242,7 +1237,8 @@ def cmd_clean(args):
 				print 'Removing %s' % file
 				os.remove(file)
 
-		for cmd in cmds:
+		if os.path.exists(os.path.join (sbl_dir, '.git')):
+			cmd = 'git clean -xdf BaseTools'
 			x = subprocess.call(cmd.split(' '), cwd=sbl_dir)
```
huang-jin-intel: 'distclean' does not completely remove all generated files in 'gitless' tree, in which what 'git clean' does. This is slightly different behavior compared to a git tree. If we don't plan to fix it the same way, provide a comment to say so.

gdong1: I am not sure if you want to add comment to code or if you want to print some info. And could you give an example of comments you want to add?
The BuildLoader.py will call edk2 script to build edk2 base tool, but it doesn't provide script to clean up them. I think it is fine not to clean up base tool for 'GitLess' case since it has no other base tool source to build.

PR #103, opened 2019-02-27
Add GRUB config file support to boot Ubuntu

mauricema:
This patch added a simple parser for grub.cfg to make it easy to boot
Ubuntu ISO image using OsLoader payload. Without it, it is required to
copy vmlinuz/initrd to root directory and create a config.cfg to list
the kernel boot command line in order to boot the ISO image. This patch
makes it possible to boot the original Ubuntu ISO (16.04 or 18.04)
directly. It provides better user experience for people who wants to
try out SBL.

Please note, same as before, when verified boot is enabled, only debug
build will support this feature. Release build will disable this feature
due to security concern, please use IAS image boot mechnism instead.


mauricema:
Added minor change to enable it for APL by default.

mauricema:
Enabled this feature QEMU and APL by default.

PR #105, opened 2019-03-03
Allow more flexible config.cfg format

mauricema:
Since grub.cfg is already supported in SBL. It makes sense to support
the similar syntax in config.cfg. This patch enabled config.cfg to
follow grub.cfg format to provide multiple boot options. The old
config.cfg format will still be supported. This patch also added
support for single/double quote around boot menu entry name.

### DIFF
--- a/PayloadPkg/OsLoader/LoadImage.c
+++ b/PayloadPkg/OsLoader/LoadImage.c
@@ -350,12 +350,24 @@ GetTraditionalLinux (
   }
 
   if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "Could not find configuration file!\n"));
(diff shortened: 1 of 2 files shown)


PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -355,7 +355,15 @@ GetTraditionalLinux (
   }
```
gdong1: Do we need support the case where there is no config.cfg file but having kernel and initrd?
Remember previously we support this case and Yocto image doesn't have config.cfg file.

mauricema: OK, will consider to add this support.

mauricema:
Made minor changes to support the case where config.cfg does not exist.

PR #114, opened 2019-03-13
Disable deprecated EDKII API interfaces

mauricema:
This patch enabled DISABLE_NEW_DEPRECATED_INTERFACES build option by
default so that the deprecated APIs cannot be used in SBL source tree.
It is to enhance the coding for security.


mauricema:
Fixed Linux GCC build errors.

PR #115, opened 2019-03-15
[APL] Update StitchLoader.py to check Boot Guard status

mauricema:
The APL StitchLoader.py script has its own limitation. For example,
if Boot Guard is enabled in the base image, the final stitched image
with this script might not boot due to security reason. It is better
to check this during stitching process so that it can error out earlier
to notify the user.


huang-jin-intel:
@sagraw2 Can you take a look too?

sagraw2:
This change looks ok to me.

PR #116, opened 2019-03-19
This patch add code to handle fsp requested reset after silicon init API call

raghavag:
This patch add code to handle fsp requested reset after silicon init API call

### DIFF
--- a/BootloaderCorePkg/Stage2/Stage2.c
+++ b/BootloaderCorePkg/Stage2/Stage2.c
@@ -424,6 +424,18 @@ SecStartup (
   AddMeasurePoint (0x3030);
   ASSERT_EFI_ERROR (Status);
 
+  //
+  // Reset the system if FSP API returned FSP_STATUS_RESET_REQUIRED status
+  //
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -424,6 +424,14 @@ SecStartup (
   AddMeasurePoint (0x3030);
   ASSERT_EFI_ERROR (Status);
 
+  //
+  // Reset the system if FSP API returned FSP_STATUS_RESET_REQUIRED status
+  //
+  if ((Status >= FSP_STATUS_RESET_REQUIRED_COLD) && (Status <= FSP_STATUS_RESET_REQUIRED_8)) {
```
gdong1: do we need handle warm reset case separately?
and need put these newly added code before  ASSERT_EFI_ERROR (Status);

mauricema: Yes, let us handle cold and warm reset separately although I did not see a reason why FSP needs to do warm reboot.   
Please do the same for both FspMemoryInit and FspSiliconInit.

raghavag: Handled cold and warm resets separately in the new patch

PR #122, opened 2019-04-03
Add SBL logos for community usage

jgutbub:
Slim Bootloader has its own set of approved logos. Add them to
the repo for community usage along with guidelines on logo usage.


jgutbub:
> Can we put the logo into Platform\CommonBoardPkg\Logo instead of root directory ?

Sure, fixed in latest commit.

mauricema:
Also there is a logo.bmp under BootlaoderCorePkg.  Do we need to move that one as well ?

jgutbub:
> Also there is a logo.bmp under BootlaoderCorePkg. Do we need to move that one as well ?

We can move it, logo.bmp is being used for splash screen. The new logo files being uploaded are JPEG format so aren't for splash screen without converting to BMP format. Do we want to move logo.bmp?

mauricema:
I think it might be better to have all logo in a single place.

jgutbub:
> I think it might be better to have all logo in a single place.

Moved in latest push and updated FDF logo.bmp path.

mauricema:
Looks good to me.

PR #128, opened 2019-04-08
Adding Pre-OS boot option support

jgutbub:
Slim Bootloader needs to add support for booting
pre-OS images that will execute before booting to
the actual OS; Slim Bootloader will load both the
pre-OS image and OS image into memory and pass the
OS information to the pre-OS image so that after
the pre-OS image finishes execution it will launch
the OS itself without returning to Slim Bootloader.

### DIFF
--- a/BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h
+++ b/BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h
@@ -25,6 +25,7 @@
 #define BOOT_FLAGS_CRASH_OS        BIT1
 #define BOOT_FLAGS_TRUSTY          BIT2
 #define BOOT_FLAGS_EXTRA           BIT3
+#define BOOT_FLAGS_PRE_OS_BOOT     BIT4
 
 // This bit is used dynamically.
 #define LOAD_IMAGE_FROM_BACKUP     BIT7
--- a/PayloadPkg/OsLoader/OsLoader.c
+++ b/PayloadPkg/OsLoader/OsLoader.c
@@ -485,32 +485,90 @@ StartBooting (
 }
 
 /**
-  Boot from OsBootOption
+  Start pre-OS boot image
+
+  This function will jump to pre-OS boot image entry point.
+
+  @param[in]  PreOsBootLoadedImage Loaded pre-OS boot image information.
+  @param[in]  OsLoadedImage        Loaded OS boot image information.
+
+  @retval  RETURN_SUCCESS       boot image is return after boot
+  @retval  Others               There is error when checking boot image
+**/
+EFI_STATUS
+StartBootingPreOs (
+  IN LOADED_IMAGE            *PreOsBootLoadedImage,
+  IN LOADED_IMAGE            *OsLoadedImage
+  )
+{
+  EFI_STATUS                 Status;
+  PRE_OS_BOOT_ENTRY          PreOsBootMain;
+  PRE_OS_BOOT_PARAMS         PreOsBootParams;
+
+  DEBUG_CODE_BEGIN();
+  PrintStackHeapInfo ();
+  DEBUG_CODE_END();
+
+  PreOsBootParams.EaxGpr = 0x0;
+  PreOsBootParams.EcxGpr = 0x0;
+  PreOsBootParams.EdxGpr = 0x0;
+  PreOsBootParams.EbxGpr = 0x0;
+  PreOsBootParams.EspGpr = 0x0;
+  PreOsBootParams.EbpGpr = 0x0;
+  PreOsBootParams.EsiGpr = 0x0;
+  PreOsBootParams.EdiGpr = 0x0;
+
+  Status = RETURN_SUCCESS;
+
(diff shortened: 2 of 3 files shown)


PayloadPkg/Include/PreOsBoot.h
```diff
@@ -0,0 +1,33 @@
...
+**/
+typedef struct
+{
+  UINT32 eaxGpr;    /**< EAX General Purpose Register. */
+  UINT32 ecxGpr;    /**< ECX General Purpose Register. */
+  UINT32 edxGpr;    /**< EDX General Purpose Register. */
+  UINT32 ebxGpr;    /**< EBX General Purpose Register. */
+  UINT32 espGpr;    /**< ESP General Purpose Register. */
+  UINT32 ebpGpr;    /**< EBP General Purpose Register. */
+} PRE_OS_BOOT_PARAMS;
```
jgutbub: This structure is not finalized yet; we may need to pass in additional registers (e.g. ESI, EDI, EIP, etc.).

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -705,7 +827,11 @@ PayloadMain (
   //
   mCurrentBoot = GetCurrentBootOption (OsBootOptionList, mCurrentBoot);
   while  (mCurrentBoot < OsBootOptionList->OsBootOptionCount) {
-    BootOsImage (&OsBootOptionList->OsBootOption[mCurrentBoot]);
+    if((OsBootOptionList->OsBootOption[mCurrentBoot].BootFlags & BOOT_FLAGS_PRE_BOOT) != 0) {
+      BootPreBootImage (&OsBootOptionList->OsBootOption[mCurrentBoot], &OsBootOptionList->OsBootOption[mCurrentBoot+1]);
```
jgutbub: The pre-OS and OS image might be stored on different media devices so here we pass in the boot options individually if BOOT_FLAGS_PRE_BOOT type is detected. Also, the pre-OS and OS image are expected to be at index=n and index=n+1 respectively.

BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h
```diff
@@ -25,6 +25,7 @@
 #define BOOT_FLAGS_CRASH_OS        BIT1
 #define BOOT_FLAGS_TRUSTY          BIT2
 #define BOOT_FLAGS_EXTRA           BIT3
```
jgutbub: BOOT_FLAGS_EXTRA type does not seem to support loading images from different media devices (e.g. all images & extra images must be loaded from the same device type). Needed to add a new flag to allow loading from different media types.

PayloadPkg/Include/PreOsBoot.h
mauricema: I would suggest to add these.  Aslo we can consider to match PUSHAD/POPAD.  So that the preOS can just use a simple POPAD/RET to jump to real OS image entry.

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -484,6 +484,128 @@ StartBooting (
...
+  boot image to launch after pre-OS boot image finishes execution.
+
+  @param[in]  InPreBootOption     Pre-OS boot option to boot
+  @param[in]  InOsBootOption      OS boot option to boot
+
+  @retval  RETURN_SUCCESS       Image returns after boot into its entrypoint
+  @retval  Others               There is error to boot from this boot option
+**/
+EFI_STATUS
+BootPreBootImage (
```
mauricema: This code is very similar to BootOsImage().   Can we consider to enable BootOsImage() to support both OS and PreOS image?

PayloadPkg/Include/PreOsBoot.h
mauricema: And if it is just a single structure,  I think we can define it in existing header file.

gdong1: How about using CPU_BOOT_STATE that we ever used for trusty boot?  Of cause we need put it to a common place, maybe have a more generic name for this solution.

PayloadPkg/OsLoader/OsLoader.c
gdong1: Yes. I prefer doing similar with trusty boot. Adding a new OS flag with minor code change.

PayloadPkg/OsLoader/OsLoader.c
gdong1: Do we really have requirement to boot pre-OS and OS in different boot media? By our current design, our boot option doesn't support this case. It would be easier to have them both in same boot media.

jgutbub: Yes. There is a request for the option to integrate and launch an IFWI/SPI IAS pre-OS image and load an OS image from another storage media device that the pre-OS image will launch after it finishes execution.

jgutbub:
Updated to address most of the comments until tomorrow's discussion.

PR #139, opened 2019-04-13
Add SmBios infrastructure

stalamudupula:
Get the bare-bones ready to add Smbios tables and
report to the payload via SystemTableInfo.

### DIFF
--- a/BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
+++ b/BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
@@ -214,6 +214,7 @@ typedef struct {
   VOID             *PcdDataPtr;
   VOID             *LogBufPtr;
   VOID             *DeviceTable;
+  VOID             *SmbiosPtr;
   UINT8             PlatformName[PLATFORM_NAME_SIZE];
   UINT32            LdrFeatures;
   BL_PERF_DATA      PerfData;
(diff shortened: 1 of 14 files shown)


BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
```diff
@@ -214,6 +214,7 @@ typedef struct {
   VOID             *PcdDataPtr;
   VOID             *LogBufPtr;
   VOID             *DeviceTable;
+  VOID             *SmbiosPtr;
```
mauricema: Global data structure is more for cross-stage communications.  
For SMBIOS table, since it is built in Stage2 only,  I think the flow can be simpler by doing an allocation in Stage2, and then call SmbiosInit() to update the buffer.   I think it can be handled in the same way as ACPI table.
Putting SMBIOS pointer into Global structure does not seem to be necessary.

BootloaderCorePkg/Library/SmbiosInitLib/SmbiosInitLib.c
```diff
@@ -0,0 +1,49 @@
...
+  SmbiosEntryPoint->AnchorString[3]               =  '_';
+  SmbiosEntryPoint->EntryPointLength              = sizeof (SMBIOS_TABLE_ENTRY_POINT);
+  SmbiosEntryPoint->MajorVersion                  = 0;
+  SmbiosEntryPoint->MinorVersion                  = 1;
+  SmbiosEntryPoint->TableLength                   = SMBIOS_TABLE_MAX_LENGTH_SBL;
+  SmbiosEntryPoint->TableAddress                  = (UINT32)SmbiosEntryPoint + sizeof (SMBIOS_TABLE_ENTRY_POINT) + sizeof (UINT8);
+  SmbiosEntryPoint->EntryPointStructureChecksum   = CalculateCheckSum8 ((UINT8 *)SmbiosEntryPoint, sizeof (SMBIOS_TABLE_ENTRY_POINT));
+
+  *SmbiosEntry = (UINT32) (UINTN) SmbiosEntryPoint;
+}
```
mauricema: we might want to add some basic data type into SMBIOS table in the common init function.  And then allow platform to append new records if required.

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -529,6 +530,10 @@ SecStartup (
     PlatformService->ResetSystem = ResetSystem;
   }
 
+  // SMBIOS Initialization
```
myguitar: Can we use PCD to control this feature enabling? We don't have to add empty PlatformUpdateSmbios() and etc. if SmBios is not mandatory.

mauricema:
Abandon this pull request.  The submitter will create a new one for it.

PR #144, opened 2019-04-23
Add payload binary file check in BuildUtility.py

mauricema:
This patch added the additional existence check for payload binary.
If the file does not exist, it will error out.  Also if the binary
is already under build FV folder, skip the unnecessary copy.

### DIFF
--- a/BootloaderCorePkg/Tools/BuildUtility.py
+++ b/BootloaderCorePkg/Tools/BuildUtility.py
@@ -571,7 +571,11 @@ def gen_payload_bin (fv_dir, pld_list, pld_bin, priv_key, brd_name = None):
 		else :
 			dst_path = os.path.join(fv_dir, os.path.basename(src_file))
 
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -571,7 +571,11 @@ def gen_payload_bin (fv_dir, pld_list, pld_bin, priv_key, brd_name = None):
 		else :
 			dst_path = os.path.join(fv_dir, os.path.basename(src_file))
 
-		shutil.copy (src_file, dst_path)
+		if not os.path.exists(src_file):
+			raise Exception ("Can not fine payload file '%s' !" % src_file)
```
jgutbub: I think it should be "Cannot find" instead of "Can not fine"

mauricema: Good catch.  :)  Will change it.

PR #149, opened 2019-04-30
Update the pre-OS payload/checker parameter

jgutbub:
For the pre-OS payload/checker pass in a parameter
that contains additional information for heap space,
heap address, and additional CPU registers that may
need to be set before the pre-OS payload/checker
passes control to the OS.

### DIFF
--- a/PayloadPkg/Library/TrustyBootLib/TrustyBoot.h
+++ b/PayloadPkg/Library/TrustyBootLib/TrustyBoot.h
@@ -43,6 +43,15 @@ typedef enum {
 #pragma pack (1)
 
 typedef struct {
+  UINT32 Eip;
+  UINT32 Eax;
+  UINT32 Ebx;
+  UINT32 Esi;
+  UINT32 Edi;
+  UINT32 Ecx;
+} CPU_BOOT_STATE;
+
+typedef struct {
   // use size as version. In future, we can only add members in the structure
   UINT32         SizeOfThisStruct;
   UINT32         Version;
--- a/PayloadPkg/OsLoader/OsLoader.c
+++ b/PayloadPkg/OsLoader/OsLoader.c
@@ -478,11 +478,11 @@ StartBooting (
   IN LOADED_IMAGE            *LoadedImage
   )
 {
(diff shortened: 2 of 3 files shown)


PayloadPkg/Library/TrustyBootLib/TrustyBoot.h
```diff
@@ -42,6 +42,15 @@ typedef enum {
...
 #pragma pack (1)
 
+typedef struct {
+  UINT32 Eip;
+  UINT32 Eax;
+  UINT32 Ebx;
+  UINT32 Esi;
+  UINT32 Edi;
+  UINT32 Ecx;
+} CPU_BOOT_STATE;
```
jgutbub: Moving this structure back into TrustyBoot.h since it cannot be used for pre-OS payload support (the registers used are different).

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -497,10 +497,15 @@ StartBooting (
       EntryPoint = (PRE_OS_CHECKER_ENTRY) (UINTN)mPreOsCheckerEntry;
       BeforeOSJump ("Starting Pre-OS Checker ...");
 
-      OsBootState.Esi = (UINT32) BootParams;
-      OsBootState.Eip = BootParams->Hdr.Code32Start;
+      PreOsParams.Version  = 0x1;
+      PreOsParams.HeapSize = EFI_SIZE_TO_PAGES (0);
+      PreOsParams.HeapAddr = (UINT32) AllocatePages (PreOsParams.HeapSize);
```
gdong1: Just want to confirm that the memory space for preOs heap could be used by normal OS later, right? 
If not, we need update normal OS memory map.

jgutbub: That should be correct, all of the memory used by the pre-OS payload/checker can be reclaimed/used by the OS, it should not require any reserved memory allocation/map-entries.

jgutbub:
Please do not check this in until everyone has had time to review it. Will check-in later next week after other approvers provide their input.

PR #153, opened 2019-05-08
Optimize CopyMem to use XMM registers

jgutbub:
The current implementation for CopyMem will
move data in DWORD quantities at a time. By
using the XMM registers we can move 128 bytes
of data at a time to improve performance.

Tested using APL-based platform.

### DIFF
--- a/MdePkg/Library/BaseMemoryLibRepStr/Ia32/CopyMem.nasm
+++ b/MdePkg/Library/BaseMemoryLibRepStr/Ia32/CopyMem.nasm
@@ -38,6 +38,33 @@ ASM_PFX(InternalMemCopyMem):
     mov     esi, [esp + 16]             ; esi <- Source
     mov     edi, [esp + 12]             ; edi <- Destination
     mov     edx, [esp + 20]             ; edx <- Count
+    shr     edx, 7
+    jz      @CopyRest
+@Copy128:
+    movdqu  xmm0, [esi + 0 ]
+    movdqu  xmm1, [esi + 16]
+    movdqu  xmm2, [esi + 32]
+    movdqu  xmm3, [esi + 48]
+    movdqu  xmm4, [esi + 64]
+    movdqu  xmm5, [esi + 80]
+    movdqu  xmm6, [esi + 96]
+    movdqu  xmm7, [esi + 112]
+    movdqu  [edi + 0 ], xmm0
+    movdqu  [edi + 16], xmm1
+    movdqu  [edi + 32], xmm2
+    movdqu  [edi + 48], xmm3
+    movdqu  [edi + 64], xmm4
+    movdqu  [edi + 80], xmm5
+    movdqu  [edi + 96], xmm6
+    movdqu  [edi + 112], xmm7
+    add     esi, 128
+    add     edi, 128
+    dec     edx
+    jnz     @Copy128                    ; Copy as many 128-byte blocks as possible
+@CopyRest:
+    mov     edx, [esp + 20]
+    and     edx, 7Fh
+    jz      @End
     lea     eax, [esi + edx - 1]        ; eax <- End of Source
     cmp     esi, edi
     jae     .0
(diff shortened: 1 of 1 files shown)


jgutbub:
I am having some issue to get release build APL to work in my system but I did notice that splash screen performance increased (went from 76ms to 70ms) with this enhancement.

Not sure if all platforms will have XMM registers and also not sure if this would need to go into an Override package instead of MdePkg

mauricema:
James,  there is a different memory library instance in EDK2 MdePkg called BaseMemoryLibMmx.
I think we should add that lib into SBL instead of implementing our own.   What do you think?

mauricema:
> I am having some issue to get release build APL to work in my system but I did notice that splash screen performance increased (went from 76ms to 70ms) with this enhancement.
> 
> Not sure if all platforms will have XMM registers and also not sure if this would need to go into an Override package instead of MdePkg

XMM is supported by all current processors.  And it is in the original EDK2 MdePkg.  We just need to add it into SBL MdePkg (subset of the EDK2 MdePkg).

jgutbub:
> James, there is a different memory library instance in EDK2 MdePkg called BaseMemoryLibMmx.
> I think we should add that lib into SBL instead of implementing our own. What do you think?

I didn't realize this already existed : )
As a test I tried to copy the routine from BaseMemoryLibMmx to test on APL and I am getting a hang, need to debug some more. I replaced my InternalMemCopyMem with the one from BaseMemoryLibMmx:
 https://github.com/tianocore/edk2/blob/master/MdePkg/Library/BaseMemoryLibMmx/Ia32/CopyMem.nasm

jgutbub:
I was looking at the wrong file, I got MMX and XMM confused. For XMM I have also tried to copy the routine from here but also getting some issues when using APL, I had to change one instruction 'movntdq' to be 'movdqu' or I get a hang around Stage1B but even after fixing this I get the same hang from before when using MMX CopyMem, it will get to 'PCI Enum' message in Stage2 and hang:

https://github.com/tianocore/edk2/blob/3806e1fd139775610d8f2e7541a916c3a91ad989/MdePkg/Library/BaseMemoryLibSse2/Ia32/CopyMem.nasm

mauricema:
We might need some debug to understand what is the issue.  BTW, to use SSE instructions, it is required to set CR4 bit9.  I believe FSP should have this set already in TempRamInit().

mauricema:
I just checked the SSE2 lib implementation.  It used MOVNTDQ instruction, which could cause issue in NEM mode since it uses WC memory type to write to memory directly without going through cache hierarchy.  Maybe we can try to use SSE2 memory library only after Stage1B.

jgutbub:
> I just checked the SSE2 lib implementation. It used MOVNTDQ instruction, which could cause issue in NEM mode since it uses WC memory type to write to memory directly without going through cache hierarchy. Maybe we can try to use SSE2 memory library only after Stage1B.

As a test I modified the DSC file to use BaseMemoryLib | MdePkg/Library/BaseMemoryLibSse2/BaseMemoryLibSse2.inf every where except for for Stage1A & Stage1B which use BaseMemoryLib | MdePkg/Library/BaseMemoryLibRepStr/BaseMemoryLibRepStr.inf. Looks like this solve part of the issue but I am still seeing some hang in Stage2 at 'PCI Enum' string.

jgutbub:
> As a test I modified the DSC file to use BaseMemoryLib | MdePkg/Library/BaseMemoryLibSse2/BaseMemoryLibSse2.inf every where except for for Stage1A & Stage1B which use BaseMemoryLib | MdePkg/Library/BaseMemoryLibRepStr/BaseMemoryLibRepStr.inf. Looks like this solve part of the issue but I am still seeing some hang in Stage2 at 'PCI Enum' string.

Traced down the 'PCI Enum' hang some more, looks like when it calls "ProgramResource (Root, PciBarTypeMem32);" it is hanging in the first while loop in ProgramResource routine. Tracing for further data.

jgutbub:
Need to debug some more offline, may push another review later to use the EDK2 XMM library.

PR #155, opened 2019-05-13
Enhance BMP image splashing

jgutbub:
Current implementation of splashing a BMP image to the
display requires the allocation of a temporary buffer
to store the BMP image into a GOP Blt friendly format
that can then be used to splash the image to the display.
Instead of using a buffer for the whole image we can
allocate a smaller buffer to splash one line at a time
since splashing is already performed one line at a time.

Also fix a small print message regarding frame buffer
format.

### DIFF
--- a/BootloaderCommonPkg/Library/GraphicsLib/BmpFormat.c
+++ b/BootloaderCommonPkg/Library/GraphicsLib/BmpFormat.c
@@ -16,17 +16,17 @@
 #include <IndustryStandard/Bmp.h>
 #include <Protocol/GraphicsOutput.h>
 #include <Library/MemoryAllocationLib.h>
+#include <Guid/GraphicsInfoHob.h>
 
 /**
-  Convert a *.BMP graphics image to a GOP blt buffer. If a NULL Blt buffer
+  Display a *.BMP graphics image to the frame buffer. If a NULL GopBlt buffer
   is passed in a GopBlt buffer will be allocated by this routine. If a GopBlt
   buffer is passed in it will be used if it is big enough.
 
   @param  BmpImage      Pointer to BMP file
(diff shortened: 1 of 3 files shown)


BootloaderCorePkg/Stage2/Stage2Support.c
```diff
@@ -419,7 +403,7 @@ BuildBaseInfoHob (
       GfxMode = &BlGfxHob->GraphicsMode;
       DEBUG ((DEBUG_INFO, "Graphics Info: %d x %d x 32 @ 0x%08X\n",GfxMode->HorizontalResolution,\
         GfxMode->VerticalResolution, BlGfxHob->FrameBufferBase));
-      if ((GfxMode->PixelFormat != PixelRedGreenBlueReserved8BitPerColor) ||
+      if ((GfxMode->PixelFormat != PixelRedGreenBlueReserved8BitPerColor) &&
```
mauricema: good catch!  :)

BootloaderCommonPkg/Library/GraphicsLib/BmpFormat.c
```diff
@@ -233,11 +245,11 @@ ConvertBmpToGopBlt (
         // Other bit format BMP is not supported.
         //
         if (IsAllocated) {
-          // FreePool (*GopBlt);
+          FreePool (*GopBlt);
```
mauricema: can we free the buffer at one place?   Maybe change the flow a little bit to reuse the code at line 268-271 ?

jgutbub: Added a Status variable and goto statement, if there is a better way to do it please let me know : )

PR #156, opened 2019-05-14
Add BaseMemoryLibSse2 library

jgutbub:
Copying the MdePkg BaseMemoryLibSse2 library into
SBL for faster CopyMem and other routines to improve
boot performance during splash screen displaying
and OS Loader payload execution.

The BaseMemoryLibSse2 folder was copied from EDK2
GitHub based on the following commit ID:

9344f0921518309295da89c221d10cbead8531aa

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dsc
+++ b/BootloaderCorePkg/BootloaderCorePkg.dsc
@@ -59,7 +59,7 @@
   PciExpressLib|MdePkg/Library/BasePciExpressLib/BasePciExpressLib.inf
   DebugPrintErrorLevelLib|$(PLATFORM_PACKAGE)/Library/DebugPrintErrorLevelLib/DebugPrintErrorLevelLib.inf
   PrintLib|MdePkg/Library/BasePrintLib/BasePrintLib.inf
-  BaseMemoryLib|MdePkg/Library/BaseMemoryLibRepStr/BaseMemoryLibRepStr.inf
+  BaseMemoryLib|MdePkg/Library/BaseMemoryLibSse2/BaseMemoryLibSse2.inf
   SynchronizationLib|MdePkg/Library/BaseSynchronizationLib/BaseSynchronizationLib.inf
   TimeStampLib|BootloaderCommonPkg/Library/TimeStampLib/TimeStampLib.inf
   ExtraBaseLib|BootloaderCommonPkg/Library/ExtraBaseLib/ExtraBaseLib.inf
@@ -324,6 +324,7 @@
     <PcdsFeatureFlag>
       gPlatformCommonLibTokenSpaceGuid.PcdMinDecompression | TRUE
     <LibraryClasses>
+      BaseMemoryLib| MdePkg/Library/BaseMemoryLibRepStr/BaseMemoryLibRepStr.inf
       SocInitLib   | Silicon/$(SILICON_PKG_NAME)/Library/Stage1ASocInitLib/Stage1ASocInitLib.inf
       BoardInitLib | Platform/$(BOARD_PKG_NAME)/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.inf
 !if $(SKIP_STAGE1A_SOURCE_DEBUG)
@@ -333,6 +334,7 @@
 
   $(PLATFORM_PACKAGE)/Stage1B/Stage1B.inf {
     <LibraryClasses>
+      BaseMemoryLib| MdePkg/Library/BaseMemoryLibRepStr/BaseMemoryLibRepStr.inf
       SocInitLib   | Silicon/$(SILICON_PKG_NAME)/Library/Stage1BSocInitLib/Stage1BSocInitLib.inf
       BoardInitLib | Platform/$(BOARD_PKG_NAME)/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.inf
   }
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
@@ -88,6 +88,9 @@ PciDevicePresent (
   )
 {
   UINT32      Address;
+  UINT32      Loop;
+  UINT32      *Src;
+  UINT32      *Dst;
 
   //
   // Create PCI address map in terms of Bus, Device and Func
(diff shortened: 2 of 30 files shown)


MdePkg/Library/BaseMemoryLibSse2/BaseMemoryLibSse2.inf
```diff
@@ -0,0 +1,99 @@
...
+  Ia32/CompareMem.nasm
+  Ia32/ZeroMem.nasm
+  Ia32/SetMem64.nasm
+  Ia32/SetMem32.nasm
+  Ia32/SetMem16.nasm
+  Ia32/SetMem.nasm
+  Ia32/CopyMem.nasm
+  Ia32/IsZeroBuffer.nasm
+
+[Sources.X64]
```
mauricema: Let us remove all X64 files since SBL does not use it.

jgutbub: Removed X64 references and files

BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -343,8 +343,14 @@
...
   }
 
+  BootloaderCommonPkg/Library/GraphicsLib/GraphicsLib.inf {
+    <LibraryClasses>
+      BaseMemoryLib| MdePkg/Library/BaseMemoryLibSse2/BaseMemoryLibSse2.inf
+  }
+
   PayloadPkg/OsLoader/OsLoader.inf {
     <LibraryClasses>
+      BaseMemoryLib       | MdePkg/Library/BaseMemoryLibSse2/BaseMemoryLibSse2.inf
```
myguitar: Do we have perf benefit in OsLoader as well? How about Stage2?

jgutbub: In OS Loader I see it saving time in the "Parse partition info" section (hundreds of milliseconds but unfortunately this is using debug build on APL, having issue with release build boot). Stage2 is having some issue in PCI enumeration if I use it for all of Stage2 code (but when overriding for the PciEnumerationLib specifically and not all of Stage2 it works so the issue is somewhere else in Stage2). Will try APL release build boot again to get some numbers.

jgutbub: Needed to use an IAS image to boot (forgot that traditional Linux is debug only). Will post some numbers soon.

mauricema: BTW, I just looked into PCI enumeration issue with SSE2 memory lib.   It turns out it is using the lib to copy from MMIO to normal memory using SSE2 instruction.   PCI CFG MMIO needs to be accessed as DWORD.  
By changing the code as below,  it can work fine .  I think we can consider it and switch to use SSE2 lib for full Stage2.
PciDevicePresent ()
{
   // CopyMem (...)
    Src = (UINT32 *) ((UINTN)PcdGet64 (PcdPciExpressBaseAddress) + PCI_EXPRESS_LIB_ADDRESS (Bus, Device, Func, 0));
    Dst = (UINT32 *) Pci;
    for (Loop = 0; Loop < sizeof(PCI_TYPE00); Loop += 4) {
      *Dst++ = *Src++;
    }                         PCI_EXPRESS_LIB_ADDRESS (Bus, Device, Func, 0)), sizeof (PCI_TYPE00));
    return EFI_SUCCESS;
  }
}

jgutbub: Just tested the patch, it is working, will add the change to the review, thanks Maurice!

For DSC file change, do you prefer to override Stage2 & OS Loader or update BaseMemoryLib to SSE2 library and override Stage1A & Stage1B to use non-SSE2 library?

mauricema: I think it might be better to use SSE2 lib by default for performance reason.
And then only override it for Stage1A/Stage1B.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -100,8 +103,11 @@ PciDevicePresent (
   if (PciExpressRead16 (Address) == 0xFFFF) {
     return EFI_NOT_FOUND;
   } else {
-    CopyMem (Pci, (VOID *) ((UINTN)PcdGet64 (PcdPciExpressBaseAddress) + \
-                            PCI_EXPRESS_LIB_ADDRESS (Bus, Device, Func, 0)), sizeof (PCI_TYPE00));
+    Src = (UINT32 *) ((UINTN)PcdGet64 (PcdPciExpressBaseAddress) + PCI_EXPRESS_LIB_ADDRESS (Bus, Device, Func, 0));
```
mauricema: Maybe we need to add comments to explain why we did not use MemCopy().

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -100,8 +103,11 @@ PciDevicePresent (
...
+    Dst = (UINT32 *) Pci;
+    for (Loop = 0; Loop < sizeof (PCI_TYPE00); Loop += 4) {
```
mauricema: change '4' to sizeof(UINT32)

jgutbub:
Performance for release build does improve but not too significantly; here are some numbers (attached).
Before(left)vsAfter(right)_SSE2_CopyMem_APL_eMMC_Boot

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
jgutbub: Added a comment in the code and mentioned in commit message.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
jgutbub: Done.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -100,8 +103,17 @@ PciDevicePresent (
   if (PciExpressRead16 (Address) == 0xFFFF) {
     return EFI_NOT_FOUND;
   } else {
-    CopyMem (Pci, (VOID *) ((UINTN)PcdGet64 (PcdPciExpressBaseAddress) + \
-                            PCI_EXPRESS_LIB_ADDRESS (Bus, Device, Func, 0)), sizeof (PCI_TYPE00));
+    //
+    // We perform a memory copy here manually to ensure DWORD alignment
+    // as required for PCI config space MMIO accessing. Using the
+    // BaseMemoryLibSse2 CopyMem routine can cause issue since it may
+    // fall back to byte-level copying once the DQWORD copying finishes.
```
mauricema: I think the issue is more on access unit greater than DWORD.   BYTE access for PCI space should be OK.
Since SSE2 might use 128bit access,  this might be the issue.

jgutbub: Updated comment, maybe it is a more accurate description now?

PR #157, opened 2019-05-14
Add HobListPtr to pre-OS payload param struct

jgutbub:
It might be useful for the pre-OS payload/checker
to have a handle to the HOB list data for additional
support or data checking that the HOB list contains.


PayloadPkg/Include/Library/PayloadLib.h
```diff
@@ -56,6 +56,7 @@ typedef struct {
   UINT32        Version;
   UINT32        HeapSize;
   UINT32        HeapAddr;
+  UINT32        HobListPtr;
```
rprangar: Can you move this after OS State variable? That way its backwards compatible.

jgutbub: Moved to the end as requested.

PR #158, opened 2019-05-15
Add build support for PseFw image

bteo1:
Integrate Pse firmware into Slimboot.

Change-Id: I89875cd9ef03d38027e7ead6206661522d0bd273
Signed-off-by: Teo, Boon Tiong <boon.tiong.teo@intel.com>


jgutbub:
The changes look good, the only thing I'm wondering about is if the this FW binary is included in Stage2 if we want to do FWU it will require updating all of Stage2. Any thoughts anyone has on that?

mauricema:
James has a good point.  The IP FW should be able to be updated separately.   So it should not be part of the Stage2.  Instead, it needs a new region for that so that it can be authenticated and updated independly of SBL Stages and Payloads.

bteo1:
Thank for  the feedback. And yes, we can bring it back for further discussion.

bteo1:
Need to re-architect the implementation.

PR #166, opened 2019-05-20
Abstraction FatLib/ExtLib into FileSystemLib

myguitar:
FileSystemLib is a single interface to access filesystem and files.
No direct access to a specific filesystem like FAT/Ext.

- Added CloseFileSystem
  If there are opened files in a specific filesystem, those files
  will be de-allocated in CloseFileSystem.
- Added OpenFile/GetFileSize/ReadFile/CloseFile
  The ReadFile() does not allocate any memory for the file content.
  The caller of OpenFile() MUST allocate necessary memory before
  calling ReadFile().


gdong1:
Looks we put FAT and EXT into same library. Have a little concern on the size.
In the previous design, we could build in FAT, EXT or both by boardConfig to save size. now if we put into a same library, how to still support flexibility? later potentially add more file system support, need think about the size optimization.

myguitar:
> Looks we put FAT and EXT into same library. Have a little concern on the size.
> In the previous design, we could build in FAT, EXT or both by boardConfig to save size. now if we put into a same library, how to still support flexibility? later potentially add more file system support, need think about the size optimization.

No concern on size. It's same mechanism which is controlled by BoardConfig. Both FAT and Ext are in the same FileSystemLib, but those are in different files. Linker will not link each FAT/Ext files depending on BoardConfig mask.
If self.FILE_SYSTEM_SUPPORT_MASK  = 3 -> 1, Ext related files will not be linked into FileSystemLib.

gdong1:
> > Looks we put FAT and EXT into same library. Have a little concern on the size.
> > In the previous design, we could build in FAT, EXT or both by boardConfig to save size. now if we put into a same library, how to still support flexibility? later potentially add more file system support, need think about the size optimization.
> 
> No concern on size. It's same mechanism which is controlled by BoardConfig. Both FAT and Ext are in the same FileSystemLib, but those are in different files. Linker will not link each FAT/Ext files depending on BoardConfig mask.
> If self.FILE_SYSTEM_SUPPORT_MASK = 3 -> 1, Ext related files will not be linked into FileSystemLib.

That's great there is no size impact.
Still not get the point why we merge actual file system lib into FileSystemLib?

myguitar:
> > > Looks we put FAT and EXT into same library. Have a little concern on the size.
> > > In the previous design, we could build in FAT, EXT or both by boardConfig to save size. now if we put into a same library, how to still support flexibility? later potentially add more file system support, need think about the size optimization.
> > 
> > 
> > No concern on size. It's same mechanism which is controlled by BoardConfig. Both FAT and Ext are in the same FileSystemLib, but those are in different files. Linker will not link each FAT/Ext files depending on BoardConfig mask.
> > If self.FILE_SYSTEM_SUPPORT_MASK = 3 -> 1, Ext related files will not be linked into FileSystemLib.
> 
> That's great there is no size impact.
> Still not get the point why we merge actual file system lib into FileSystemLib?

There are some reasons, but 1st one is abstraction. Users do not have to know what current file system is. And, it gives same interfaces to access a file system or files regardless of file system.

gdong1:
> > > > Looks we put FAT and EXT into same library. Have a little concern on the size.
> > > > In the previous design, we could build in FAT, EXT or both by boardConfig to save size. now if we put into a same library, how to still support flexibility? later potentially add more file system support, need think about the size optimization.
> > > 
> > > 
> > > No concern on size. It's same mechanism which is controlled by BoardConfig. Both FAT and Ext are in the same FileSystemLib, but those are in different files. Linker will not link each FAT/Ext files depending on BoardConfig mask.
> > > If self.FILE_SYSTEM_SUPPORT_MASK = 3 -> 1, Ext related files will not be linked into FileSystemLib.
> > 
> > 
> > That's great there is no size impact.
> > Still not get the point why we merge actual file system lib into FileSystemLib?
> 
> There are some reasons, but 1st one is abstraction. Users do not have to know what current file system is. And, it gives same interfaces to access a file system or files regardless of file system.

We have already an file system abstraction layer in FileSystemLib.h to have same interface to access different actual file system. And EnumFileSystemTypeAuto is supported if user doesn't know current file system. If we want to add more API to file system, we could extend FileSystemLib and actual file system impl, my question is why we put actual file system impl into FileSystemLib instead of keep them as-is. BTW, we have similar methods in MediaAccessLib.h to access different boot media.

myguitar:
> > > > > Looks we put FAT and EXT into same library. Have a little concern on the size.
> > > > > In the previous design, we could build in FAT, EXT or both by boardConfig to save size. now if we put into a same library, how to still support flexibility? later potentially add more file system support, need think about the size optimization.
> > > > 
> > > > 
> > > > No concern on size. It's same mechanism which is controlled by BoardConfig. Both FAT and Ext are in the same FileSystemLib, but those are in different files. Linker will not link each FAT/Ext files depending on BoardConfig mask.
> > > > If self.FILE_SYSTEM_SUPPORT_MASK = 3 -> 1, Ext related files will not be linked into FileSystemLib.
> > > 
> > > 
> > > That's great there is no size impact.
> > > Still not get the point why we merge actual file system lib into FileSystemLib?
> > 
> > 
> > There are some reasons, but 1st one is abstraction. Users do not have to know what current file system is. And, it gives same interfaces to access a file system or files regardless of file system.
> 
> We have already an file system abstraction layer in FileSystemLib.h to have same interface to access different actual file system. And EnumFileSystemTypeAuto is supported if user doesn't know current file system. If we want to add more API to file system, we could extend FileSystemLib and actual file system impl, my question is why we put actual file system impl into FileSystemLib instead of keep them as-is. BTW, we have similar methods in MediaAccessLib.h to access different boot media.

The 2nd reason is because it's hard to track the use of heap memory. We have FatLib/ExtLib and FileSystemLib which are using FatLib/ExtLib. If users use FatLib/ExtLib/FileSystemLib separately, it's hard to manage memory usage. This change explicitly does not allow users to use FatLib and ExtLib directly to avoid complexity. Instead, FileSystemLib has two sub-directories FatLib/Ext23Lib which has original implementation. FYI, FileSystemLib manages opened files of each file system and those files will be closed at CloseFileSystem() if they are not closed properly to avoid memory leak.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -344,15 +401,41 @@ GetTraditionalLinux (
...
+      continue;
+    }
+
+    ConfigFile = AllocatePool (ConfigFileSize);
+    if (ConfigFile == NULL) {
+      CloseFile (FileHandle);
+      return EFI_OUT_OF_RESOURCES;
+    }
+
+    Status = ReadFile (FileHandle, &ConfigFile, &ConfigFileSize);
```
gdong1: we should not allocate memory every time when reading file. In the file system instance (e.g. ExtFsReadFile), the memory was allocated there.  or just allocate in caller, not allocate memory in file system instance.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -299,13 +328,40 @@ LoadLinuxFile (
...
+    goto Done;
+  }
+
+  FileBuffer = AllocatePool (FileSize);
+  if (FileBuffer == NULL) {
+    Status = EFI_OUT_OF_RESOURCES;
+    goto Done;
+  }
+
+  Status = ReadFile (FileHandle, &FileBuffer, &FileSize);
```
gdong1: is it possible to make the APIs exactly same for FileSystemLib and specific file system (e.g. Fat)? look currently FsHandle is not exposed in FileSystemLib.

PayloadPkg/OsLoader/LoadImage.c
myguitar: Good catch. This is ExtFsReadFile bug. Let me remove memory allocation in ExtFsReadFile.

PayloadPkg/OsLoader/LoadImage.c
myguitar: This is known issue. Currently, FatReadFile requires FsHandle as a parameter. Let me change FatReadFile in separate commit.

PayloadPkg/OsLoader/LoadImage.c
gdong1: It is fine to fix it in another patch.

PR #167, opened 2019-05-20
[PartitionLib] Added ClosePartitions to clean-up partition data

myguitar:
The FindPartitions() allocates memory for PART_BLOCK_DEVICE instance.
This allocated memory needs to be de-allocated if no more necessary
to avoid memory leak.

TBD: Current partition info needs to be cleared before going to next
boot option.

### DIFF
--- a/BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
+++ b/BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
@@ -596,3 +596,30 @@ FindPartitions (
 
   return Status;
 }
+
+/**
+  Clean-up allocated memory/etc. used for partitions
+
+  @param[in]  PartHandle      The partition handle to clean-up
+
+  @retval                     none
+
+**/
+VOID
+EFIAPI
+ClosePartitions (
+  IN   EFI_HANDLE             PartHandle
+  )
+{
+  PART_BLOCK_DEVICE           *PartBlockDev;
+
+  PartBlockDev = NULL;
+  if (PartHandle != NULL) {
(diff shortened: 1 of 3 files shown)


BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -596,3 +596,30 @@ FindPartitions (
...
+{
+  PART_BLOCK_DEVICE           *PartBlockDev;
+
+  PartBlockDev = NULL;
+  if (PartHandle != NULL) {
+    PartBlockDev = (PART_BLOCK_DEVICE *)PartHandle;
+    ASSERT (PartBlockDev->Signature == PART_INFO_SIGNATURE);
+  }
+
+  if (PartBlockDev != NULL && PartBlockDev->Signature == PART_INFO_SIGNATURE) {
```
mauricema: Maybe add bracket to make the condition easier to understand.

mauricema: Actually, I don't think it is required check (PartBlockDev->Signature == PART_INFO_SIGNATURE)
We can always try to free PartBlockDev if it is not NULL, right?

myguitar: Added Bracket.

PartBlockDev->Signature == PART_INFO_SIGNATURE
This is to avoid invalid memory de-allocation in Release build when caller gives a pointer which is not PartHandle. Caller must guarantee correct PartHandle, but double check it just in case.

mauricema: OK, make sense.

PR #168, opened 2019-05-22
Package multiple firmware Images into single capsule

raghavag:
This patch will update GenCapsuleFirmware.py tool to package
multiple firmware update images into single capsule.

Example:

python GenCapsuleFirmware.py -p BIOS pld1.bin -p <GUID> pld2.bin
	   -p CSME Pld3.bin -p CSMD Pld4.bin -o img.bin -k  TestSigningPrivateKey.pem

The following are the changes made

1) Capsule header format is changed to align with
   EFI_FIRMWARE_MANAGEMENT_CAPSULE_HEADER
2) Each firmware image in the capsule will have
   EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER
3) BIOS, CSME, CSME Update Driver (CSMD) and CFGData payloads
   can be passed to the tool using their names as suggested in
   example.
4) Unknown payloads can also be passed to the tool with associated
   guid as input.

### DIFF
--- a/BootloaderCorePkg/Tools/GenCapsuleFirmware.py
+++ b/BootloaderCorePkg/Tools/GenCapsuleFirmware.py
@@ -15,14 +15,371 @@ import os
 import argparse
 import subprocess
 import sys
-from ctypes import *
 import re
 import struct
 import uuid
-import ctypes
+from ctypes import *
 
 sys.dont_write_bytecode = True
 from BuildUtility import rsa_sign_file, gen_pub_key
+
+
+class FmpCapsuleImageHeaderClass(object):
+    # typedef struct {
+    #   UINT32   Version;
+    #
+    #   ///
+    #   /// Used to identify device firmware targeted by this update. This guid is matched by
+    #   /// system firmware against ImageTypeId field within a EFI_FIRMWARE_IMAGE_DESCRIPTOR
+    #   ///
+    #   EFI_GUID UpdateImageTypeId;
+    #
+    #   ///
+    #   /// Passed as ImageIndex in call to EFI_FIRMWARE_MANAGEMENT_PROTOCOL.SetImage ()
+    #   ///
+    #   UINT8    UpdateImageIndex;
+    #   UINT8    reserved_bytes[3];
+    #
+    #   ///:
+    #   /// Size of the binary update image which immediately follows this structure
+    #   ///
+    #   UINT32   UpdateImageSize;
+    #
+    #   ///
+    #   /// Size of the VendorCode bytes which optionally immediately follow binary update image in the capsule
+    #   ///
+    #   UINT32   UpdateVendorCodeSize;
+    #
+    #   ///
+    #   /// The HardwareInstance to target with this update. If value is zero it means match all
+    #   /// HardwareInstances. This field allows update software to target only a single device in
+    #   /// cases where there are more than one device with the same ImageTypeId GUID.
+    #   /// This header is outside the signed data of the Authentication Info structure and
+    #   /// therefore can be modified without changing the Auth data.
+    #   ///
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -15,105 +15,532 @@
...
-
-	os.remove(pubkey_file)
-	os.remove(fwupdate_bin_file)
+    _fields_ = [
+        ('FileGuid', ARRAY(c_uint8, 16)), ('HeaderSize', c_uint32),
+        ('FirmwreVersion', c_uint32), ('CapsuleFlags', c_uint32),
+        ('PubKeyOffset', c_uint32), ('PubKeySize', c_uint32),
+        ('ImageOffset', c_uint32), ('ImageSize', c_uint32),
+        ('SignatureOffset', c_uint32), ('SignatureSize', c_uint32),
+        ('Reserved', ARRAY(c_uint32, 3))
```
gdong1: Had better have one field per line as before for better readable.

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -15,105 +15,532 @@
...
+    ]
+
+
+def SignImage(RawData, OutFile, PrivKey):
```
gdong1: It is difficult to get change difference since SignImage() was removed and then added here.
Is it caused by CR or space or other unnecessary change?

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -15,105 +15,532 @@
...
-
-  #
-  # Create command line argument parser object
-  #
-  parser = argparse.ArgumentParser()
-  parser.add_argument('-b',  '--image', dest='Image', type=str, help='Raw image into firmware update capsule image', required = True)
-  parser.add_argument('-t',  '--type', dest='Type', choices=['b','c'], help='Image type b:Bootloader c:CfgData', default = 'b')
-  parser.add_argument('-k',  '--priv_key', dest='PrivKey', type=str, help='Private RSA 2048 key in PEM format to sign image', required = True)
-  parser.add_argument('-o',  '--output', dest='NewImage', type=str, help='Output file for signed image', required = True)
-  parser.add_argument("-q", "--quiet", dest='Quiet', action="store_true", help="without output messages or temp files")
```
gdong1: If removed the arguments, how it could support the parameter "-p", "-o" and "-k" in your commit message?

raghavag: If you see what are removed, these are all unused args, Look from line# 513 in the new file, args are parsed there

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
raghavag: I believe when I correct your above comment, this will be resolved as well.

BTW, it is not removed and added here, it is just alignment issue

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
raghavag: Will Fix this

raghavag: Fixed this in latest release

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
gdong1: sorry I missed the one you added since code added/removed are not back to back in the patch.
That's why we prefer only changing necessary code in patch.  The new format is not aligned in different parameter. the old one is more readable.

PR #173, opened 2019-06-03
Clean up PCI MM address

gdong1:
Common MM_PCI_ADDRESS() provided PCI device BDF to PCIE MMIO base address.
So remove MmPciAddress() and MmPciBase () defined in platform and update
code to use MM_PCI_ADDRESS().
Add TO_PCI_LIB_ADDRESS() in common library.

### DIFF
--- a/BootloaderCommonPkg/Include/Library/BootloaderCommonLib.h
+++ b/BootloaderCommonPkg/Include/Library/BootloaderCommonLib.h
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2017 - 2018, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2019, Intel Corporation. All rights reserved.<BR>
   This program and the accompanying materials
   are licensed and made available under the terms and conditions of the BSD License
   which accompanies this distribution.  The full text of the license may be found at
@@ -63,6 +63,9 @@ typedef struct {
 
 #define TO_MM_PCI_ADDRESS( Addr ) \
   MM_PCI_ADDRESS ( ((Addr) >> 16) & 0xFF, ((Addr) >> 8) & 0xFF, (Addr) & 0xFF, 0)
+
+#define TO_PCI_LIB_ADDRESS( Addr ) \
+  PCI_LIB_ADDRESS ( ((Addr) >> 16) & 0xFF, ((Addr) >> 8) & 0xFF, (Addr) & 0xFF, 0)
 
 //
 // Enumeration of stages of bootloader execution.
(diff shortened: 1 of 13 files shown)


BootloaderCommonPkg/Library/UsbDeviceLib/UsbDeviceLib.c
```diff
@@ -42,13 +42,13 @@ UsbDeviceDriverSupported (
 
   UsbInfo = &mUsbDeviceInfo;
 
-  Address = MmPciAddress ( 0, UsbInfo->XdciDeviceAddress.Bus, UsbInfo->XdciDeviceAddress.Device,
+  Address = MM_PCI_ADDRESS (UsbInfo->XdciDeviceAddress.Bus, UsbInfo->XdciDeviceAddress.Device,
```
mauricema: why not just  define MmPciAddress as macro.  In this way it needs much less changes.  Also it is easy to compare with original EDKII code.

BootloaderCommonPkg/Include/Library/BootloaderCommonLib.h
```diff
@@ -64,6 +64,9 @@ typedef struct {
 #define TO_MM_PCI_ADDRESS( Addr ) \
   MM_PCI_ADDRESS ( ((Addr) >> 16) & 0xFF, ((Addr) >> 8) & 0xFF, (Addr) & 0xFF, 0)
 
+#define TO_PCI_LIB_ADDRESS( Addr ) \
```
mauricema: why do we need both TO_MM_PCI_ADDRESS  and TO_PCI_LIB_ADDRESS ?

BootloaderCommonPkg/Library/UsbDeviceLib/UsbDeviceLib.c
gdong1: By the coding style of EDK2: Use all capital letters for both #define and typedef declarations.
This clearly differentiates static declarations from dynamic data types. So just not use MmPciAddress().

BootloaderCommonPkg/Include/Library/BootloaderCommonLib.h
gdong1: Before MMIO access is available, PCI lib address is required in early phase, e.g. to enable SPI flash access.

mauricema: Since they are the same,  any reason for duplication with different macro names ?

BootloaderCommonPkg/Library/UsbDeviceLib/UsbDeviceLib.c
mauricema: Understand the coding style requirement.  But these changes make it difficult to compare with the origin files from EDK2.  If it is a concern,  we can add this to the common lib as a real function.

gdong1: Roll back changes to common USB lib. only update platform code for MmPciAddress() in this patch.

BootloaderCommonPkg/Include/Library/BootloaderCommonLib.h
gdong1: They are different. one is for MMIO access, one is for CF8/CFC legacy access.

PR #182, opened 2019-06-12
[Script] Skip CopyFspBins for unsupported platforms

myguitar:
Instead of Exception, need to skip FSP copying for all unsupported platforms


BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -120,7 +93,7 @@ def CopyFspBins (fsp_dir, sbl_dir, platform):
           ('CoffeeLakeFspBinPkg/SampleCode/Vbt/Vbt.bsf', 'Platform/CoffeelakeBoardPkg/VbtBin/Vbt.bsf')
         ])
     else:
-        Fatal ('%s platform is not supported yet !' % platform)
```
jgutbub: Can you consider leaving this line here but just change it to 'print' instead of 'Fatal' ? I also have some build issue because of this one.

jgutbub: It's okay to just remove it if preferred, I don't have a strong opinion on this one really, just want this fix.

PR #186, opened 2019-06-13
Update BSD license to BSD+Patent license

gdong1:
To align with EDK2, update file license
to use BSD+Patent license


mauricema:
Thank you for making this change. It looks good.

PR #189, opened 2019-06-13
Fix boot performance issue for BP1

### LINKED ISSUE #188: [WHL] Boot performance is very slow when booting from BP1
On WHL with Boot Guard profile 0, if booting from SBL boot partition 1,  the boot time is significantly longer than booting from boot partition 0.

mauricema:
On WHL, if using Boot Guard profile 0, booting from BP1 will be
significantly slower than BP0. It is because some code region in BP1
is not covered by MTRR cache settings. This patch adjusted MTRR
settings during PostTempRamInit notification to cover full flash
code region if Boot Guard profile 0 is used.

It fixed #188.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
@@ -15,15 +15,14 @@
 #include <Library/DebugLib.h>
 #include <GpioPinsCnlLp.h>
 #include <RegAccess.h>
+#include <Register/CpuRegs.h>
 #include <FsptUpd.h>
 #include <PlatformData.h>
 
 #define UCODE_REGION_BASE   FixedPcdGet32(PcdUcodeBase)
 #define UCODE_REGION_SIZE   FixedPcdGet32(PcdUcodeSize)
-// PcdPayloadBase is too big. using following value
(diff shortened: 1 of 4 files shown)


Platform/CoffeelakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
```diff
@@ -160,6 +163,24 @@ BoardInit (
...
       SetCurrentBootPartition (BootPartition == BootPartition2 ? 1 : 0);
     }
+
+    // Enlarge the code cache region to cover full flash for non-BootGuard case only
+    if ((AsmReadMsr64 (MSR_BOOT_GUARD_SACM_INFO) & B_BOOT_GUARD_SACM_INFO_NEM_ENABLED) == 0) {
+       // WHL FSP-T does not allow to enable full flash code cache due to cache size restriction.
+       // Here, MTRR is patched to enable full flash region cache to avoid performance penalty.
+       // However, the SBL code flow should ensure only limited flash regions will be accessed
+       // before FSP TempRamExit() is called. The combined DATA and CODE cache size should satisfy
+       // the BWG requirement.
```
sindhuragrandhi: By BWG, do you mean BIOS Writer's Guide?

mauricema: Yes, it is.

raghavag:
Changes look good to me

PR #194, opened 2019-06-15
Exclude board notifications for built-in payloads

mauricema:
Previous commit 4061d47f3041e95a23c9307d38ea07c4dd62b9cd missed one
condition while handling the board notification calls. For built-in
payload such as OsLoader and FwUpdate, it will issue notifications
from payload through SBL platform services. So for these payloads,
board notification should not be called in SBL Stage2. This patch
fixed this.

### DIFF
--- a/BootloaderCorePkg/Stage2/Stage2.c
+++ b/BootloaderCorePkg/Stage2/Stage2.c
@@ -225,6 +225,7 @@ NormalBootPath (
   LOADER_GLOBAL_DATA             *LdrGlobal;
   EFI_STATUS                      Status;
   BOOLEAN                         CallBoardNotify;
+  UINT32                          PayloadId;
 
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -270,12 +271,19 @@ NormalBootPath (
 
   BoardInit (EndOfStages);
 
-  CallBoardNotify = TRUE;
-  if ((GetPayloadId () == UEFI_PAYLOAD_ID_SIGNATURE) && (Dst[0] != 0)) {
+  PayloadId = GetPayloadId ();
```
myguitar: What about adding a field in BuildLoader.py -p option?

mauricema: If it is a build option, it might be easily missed by end user.  For OsLoader, we need to enforce this to avoid some issues.   If we have more payloads need to be handled specially,  then we can consider adding build option.

PR #198, opened 2019-06-18
[APL] Fix boot failure in OsLoader

### LINKED ISSUE #197: [APL] Boot failure when board notifications are called in SBL Stage2
When SBL calls board notification ReadyToBoot and EndOfFirmware in Stage2,  OsLoader will assert on APL platform.  
Debug log as below:
...                                                                                                    
Payload normal heap: 0x4000000 (0xE3A000 used)                                                      
Payload reserved heap: 0x900000 (0x0 used)                                                          
Payload stack: 0x10000 (0x1C00 used)                                                                
                                                                                                    
ASSERT [Stage2] w:\slimboot\slimfork\MdePkg\Library\BaseMemoryLibSse2\ZeroMemWrapper.c(47): Buffer != ((void *) 0)                                                                                      
                                                                                                    
STAGE_2: System halted!

mauricema:
When SBL calls board notification ReadyToBoot and EndOfFirmware in
Stage2, OsLoader will assert on APL platform. It was caused by NULL
pointer access in ClearFspHob(). This patch added NULL pointer check
before zeroing out memory. It fixed #197.

### DIFF
--- a/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -387,7 +387,7 @@ AssignPciIrqs (
   // Need to program interrupt register for add-in cards
   Shift = 0;
   for (Function = PCI_FUNCTION_NUMBER_PCIE_ROOT_PORT_5; Function <= PCI_FUNCTION_NUMBER_PCIE_ROOT_PORT_6; Function++) {
-    PciBase = MM_PCI_ADDRESS (DEFAULT_PCI_BUS_NUMBER_SC, PCI_DEVICE_NUMBER_SC_PCIE_DEVICE_2, Function, 0); 
+    PciBase = MM_PCI_ADDRESS (DEFAULT_PCI_BUS_NUMBER_SC, PCI_DEVICE_NUMBER_SC_PCIE_DEVICE_2, Function, 0);
     if (MmioRead32(PciBase) == 0xFFFFFFFF) {
       continue;
     }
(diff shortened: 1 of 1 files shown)


Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -578,9 +578,11 @@ ClearFspHob (
 
   LdrGlobal  = (LOADER_GLOBAL_DATA *)GetLoaderGlobalDataPointer ();
   HandOffHob = (EFI_HOB_HANDOFF_INFO_TABLE  *) LdrGlobal->FspHobList;
-  Length     = (UINT8 *) (UINTN) HandOffHob->EfiEndOfHobList - (UINT8 *)HandOffHob;
-  ZeroMem (HandOffHob, Length);
-  LdrGlobal->FspHobList = NULL;
+  if (HandOffHob != NULL) {
```
gdong1: Do we know why FspHobList could be NULL? we might need update logic to make sure only free it once.

mauricema: FspHobClear() will be called in board notification.  
And FspHobList will be set to NULL.   When in OsLoader, it tries to notify again,  this will happen.

gdong1: Could we also avoid to call FspHobClear() twice?

mauricema: I think this patch will handle it.  When FspHobClear() is called more than once, it will just return.

gdong1: Agree this patch could WA the issue.  At same time, should we avoid it is invoked twice?

mauricema: I think it is better to handle it inside ClearHobClear() to make it work in either way.
ClearHobClear() will be called from board notification.  The board notification should be called only once per notify code.   This is not enforced for now.

gdong1: Some notifications, e.g. EndOfStages, EndOfFirmware should be invoked only once.
I am ok to check this fix firstly and update others code later.

mauricema: OK, let us handle it separately.
For this fix, it is required anyway since it can potentially access NULL pointer.

PR #200, opened 2019-06-19
Store GPT entry info for later usage

jgutbub:
It may be useful to store the GPT entry
data when it is read from the GPT partitioned
block device during the partition location
routines. The partition GUIDs may be needed
for other purposes during the boot setup.

### DIFF
--- a/BootloaderCommonPkg/Include/Library/PartitionLib.h
+++ b/BootloaderCommonPkg/Include/Library/PartitionLib.h
@@ -11,6 +11,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define _PARTITION_LIB_H_
 
 #include <BlockDevice.h>
+#include <Uefi/UefiGpt.h>
 
 #define  PART_INFO_SIGNATURE      SIGNATURE_32 ('P', 'A', 'R', 'T')
 
@@ -100,6 +101,21 @@ VOID
 EFIAPI
 ClosePartitions (
   IN   EFI_HANDLE             PartHandle
+  );
+
+/**
+  Get the GPT entries array from the last GPT block
+  device read.
+
+  @param  None
+
+  @retval                   Pointer to the GPT entry data
+
+**/
+EFI_PARTITION_ENTRY *
+EFIAPI
+GetGptEntryData (
+  VOID
   );
 
 #endif
--- a/BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
+++ b/BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
@@ -9,7 +9,6 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #include <PiPei.h>
 #include <IndustryStandard/Mbr.h>
-#include <Uefi/UefiGpt.h>
 #include <Library/MemoryAllocationLib.h>
 #include <Library/BaseMemoryLib.h>
 #include <Library/Crc32Lib.h>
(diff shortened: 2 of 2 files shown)


BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -388,13 +406,22 @@ FindGptPartitions (
   ReadSize = Gpt->NumberOfPartitionEntries * Gpt->SizeOfPartitionEntry;
   ReadSize = (ReadSize % DevBlockInfo->BlockSize) == 0 ? ReadSize : DevBlockInfo->BlockSize * ((
                ReadSize / DevBlockInfo->BlockSize) + 1);
-  GptEntries = (EFI_PARTITION_ENTRY *) AllocatePool (ReadSize);
-  if (GptEntries == NULL) {
-    Status = EFI_OUT_OF_RESOURCES;
-    goto Done;
+
+  if (mGptEntriesSize < ReadSize) {
```
mauricema: Can we avoid using module global variable for this?   We might have multiple partition instance opened at the same time, it will cause variable sharing between different instance.   Since each instance already has its own private data PART_BLOCK_DEVICE,  the GptEntries can be one member inside this structure.

BootloaderCommonPkg/Include/Library/PartitionLib.h
```diff
@@ -102,4 +103,19 @@ ClosePartitions (
...
+  device read.
+
+  @param  None
+
+  @retval                   Pointer to the GPT entry data
+
+**/
+EFI_PARTITION_ENTRY *
+EFIAPI
+GetGptEntryData (
```
mauricema: Considering PartitionLib.h is a library class.    All interfaces for this lib should be partition type independent.  
So we should not provide GPT specific API interface in the library class header.   Either rename this API to be more generic or move it into private header file.

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -62,6 +63,25 @@ GetPartitionTypeName (
...
+  device read.
+
+  @param  None
+
+  @retval                   Pointer to the GPT entry data
+
+**/
+EFI_PARTITION_ENTRY *
+EFIAPI
+GetGptEntryData (
```
mauricema: Is anybody using this new API ?
And for what purpose ?

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
myguitar: I agree with @mauricema.

jgutbub: I am not aware of any scenarios where we will have multiple block devices mounted at the same time (if one boot option fails then we get the next block device's partition info and update the GPT data to the latest block device's entry data).

How to get access to PART_BLOCK_DEVICE data during kernel command line parameter setup?

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
jgutbub: Some Linux kernel command line parameters require GPT partition UUID information.

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
mauricema: Our code does not do it currently.  But the design allows to open multiple partition handles.  The implementation should consider that.
The partition handle should be available in OsLoader.  It needs to be passed back into the API as a parameter.  And inside the API, from the handle, you can get its private data by typecasting.

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
mauricema: Shouldn't the kernel command line parameter be provided by user or cfg file ?   Why do we need to know this in OsLoader ?
If this type of function is really required,  we can consider some API like  GetPartitionInfo() to abstract it.  But we still need to understand what data is necessary for bootloader to consume.

PR #201, opened 2019-06-19
[APL] Add Payload ID selection through GPIO pin

mauricema:
This patch allows to use one GPIO pin to select different Payload ID.
When enabled, if GEN_CFG_DATA.PayloadId is set to 'AUTO', the GPIO pin
low level will select OsLoader payload, and high level will select UEFI
payload. If GEN_CFG_DATA.PayloadId is set to other values, the GPIO pin
low level will always select OsLoader payload, and high level will
select the Payload ID specified by GEN_CFG_DATA.PayloadId.

On APL LeafHill, OxbowHill and JuniperHill boards, the current GPIO is
set to GP48 (Pin 10 on J6B2 connector). This GPIO pin should be used as
input signal, and can be reconfigured for alternative functions after
the Payload ID selection is done.


Platform/ApollolakeBoardPkg/CfgData/CfgData_CapsuleInformation.dsc
```diff
@@ -5,17 +5,21 @@
...
   # !BSF TYPE:{Combo}
   # !BSF OPTION:{0:SATA, 1:SD, 2:EMMC, 3:UFS, 4:SPI, 5:USB, 6:NVME, 7:MAX}
   # !BSF HELP:{Specify boot device}
   gCfgData.DevType                      |      * | 0x01 | 5
 
-  # !BSF NAME:{Boot Device instance}
+  # !BSF NAME:{Device instance for Capsule}
   # !BSF TYPE:{Combo}
   # !BSF OPTION:{0:Device 0, 1:Device 1, 2:Device 2, 3:Device 3}
   # !BSF HELP:{Specify boot device instance when then are multple instances}
```
gdong1: Also update help info to remove boot and add capsule?

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -627,6 +627,46 @@ PlatformCpuInit (
...
+  GEN_CFG_DATA       *GenericCfgData;
+  PLATFORM_CFG_DATA  *PlatCfgData;
+  UINT32              PayloadId;
+  UINT32              GpioLevel;
+
+  PayloadId = GetPayloadId ();
+  GenericCfgData = (GEN_CFG_DATA *)FindConfigDataByTag (CDATA_GEN_TAG);
+  if (GenericCfgData != NULL) {
+    if (GenericCfgData->PayloadId == AUTO_PAYLOAD_ID_SIGNATURE) {
+      PayloadId = 0;
```
gdong1: Looks PayloadId is got in line 644. 
In this case, if the value is auto, should it be 0 or keep the original value?

Platform/ApollolakeBoardPkg/CfgData/CfgData_CapsuleInformation.dsc
mauricema: OK, will update this.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: this is the first time we set Payload ID value.  The original should be the same as 0.
But I think we can remove this to simplify the code.  Will change the code.

mauricema: Just revisited the code logic, I think it is better to set it to 0 to boot OsLoader if PayloadId is AUTO in CFGDATA.  This is to enforce we will boot to a good known PayloadId in case that the original value is not valid for some reason. This makes the behavior more deterministic.

mauricema:
Updated a new patch.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
gdong1: Should we update Line 644 to be PayloadId = 0; if this is the first time to set PayloadId?
It is a little confusing if we get PayloadId in Line 644, and then update to 0 in line 648.

mauricema: OK, updated the code.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -627,6 +627,44 @@ PlatformCpuInit (
...
+
+  GpioLevel = 1;
+  PlatCfgData = (PLATFORM_CFG_DATA *)FindConfigDataByTag (CDATA_PLATFORM_TAG);
+  if ((PlatCfgData != NULL) && (PlatCfgData->PayloadSelGpio.Enable != 0)) {
+    // The default GPIOSet to Pull Up 20K
+    GpioGetInputValue (PlatCfgData->PayloadSelGpio.PadInfo, 0x0C, &GpioLevel);
+    if (GpioLevel == 0) {
+      PayloadId = 0;
+    } else {
+      if ((GenericCfgData != NULL) && (GenericCfgData->PayloadId == AUTO_PAYLOAD_ID_SIGNATURE)) {
```
gdong1: It will be more clear if moving line 660 before line 653. 
"Set PayloadId by Gpio pin" is required only when PayloadId is auto.
In current logic, if PayloadId is set to auto, but config data CDATA_PLATFORM_TAG is not found, the final PayloadId would be auto, this is not expected.

mauricema: In the case you mentioned,  the payload ID will be 0.

gdong1: oh, I had thought PayloadId will be auto. It is OK.

PR #213, opened 2019-06-26
Handle empty and NULL strings in Smbios types

stalamudupula:
If the platform provides an empty or a NULL string,
Smbios type might end abruptly and the Types are
reported incorrectly.

### DIFF
--- a/BootloaderCommonPkg/Library/SmbiosInitLib/SmbiosInitLib.c
+++ b/BootloaderCommonPkg/Library/SmbiosInitLib/SmbiosInitLib.c
@@ -244,7 +244,10 @@ GetSmbiosString (
 
   for (Index = 0; Index < PcdGet16 (PcdSmbiosStringsCnt); ++Index) {
     if (Type == SmbiosStrings[Index].Type && StringId == SmbiosStrings[Index].Idx) {
-      return SmbiosStrings[Index].String;
+      if (SmbiosStrings[Index].String != NULL && AsciiStrLen (SmbiosStrings[Index].String) != 0) {
+        return SmbiosStrings[Index].String;
+      }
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/SmbiosInitLib/SmbiosInitLib.c
```diff
@@ -244,7 +244,15 @@ GetSmbiosString (
 
   for (Index = 0; Index < PcdGet16 (PcdSmbiosStringsCnt); ++Index) {
     if (Type == SmbiosStrings[Index].Type && StringId == SmbiosStrings[Index].Idx) {
-      return SmbiosStrings[Index].String;
+      if (SmbiosStrings[Index].String != NULL) {
```
myguitar: @stalamudupula There are multiple "unknown" in this function. Can you do more clean-up?

BootloaderCommonPkg/Library/SmbiosInitLib/SmbiosInitLib.c
```diff
@@ -244,11 +245,26 @@ GetSmbiosString (
...
+          StringFound = TRUE;
+          break;
+        } else {
+          StringFound = FALSE;
+          break;
+        }
+      } else {
+        StringFound = FALSE;
+        break;
+      }
```
gdong1: The patch and logic would be very simple if using below code to replace "return SmbiosStrings[Index].String;"
if ((SmbiosStrings[Index].String != NULL) && (AsciiStrLen (SmbiosStrings[Index].String) != 0)) {
    return AsciiStrLen (SmbiosStrings[Index].String;
}
break;

PR #221, opened 2019-07-02
Update current version in ACPI ESRT table

raghavag:
This patch will update Slim Bootloader current
firmware version to ESRT table and for all other
firmware, platform hook routine will be called to
identify the firmware and update current version for it.


BootloaderCorePkg/Include/Library/BoardInitLib.h
```diff
@@ -118,6 +118,23 @@ PlatformUpdateHobInfo (
...
+
+  @param  Guid          The GUID to identify the firmware.
+  @param  Version       Pointer to the version information.
+
+**/
+VOID
+PlatformUpdateCurrentVersion (
+  IN CONST EFI_GUID   *Guid,
+  IN OUT UINT32       *Version
+  );
```
gdong1: Add EFIAPI modifier for lib API.
is it possible to use component signature instead of GUID for the first parameter?
In our internal flash map, we use signature to identify component, it will be consistent with flash map. And we will support more components to update, it would be better if we could avoid to add more and more guids for them.

raghavag: I will add EFIAPI.

Firmwares inside the capsule are identified with the GUID, this is according to the UEFI specification. It is unnecessary to add a conversion from GUID to name again.

gdong1: This API is used to get current version, not the version inside capsule image, right? 
if so, it is not related with GUID inside capsule image.

raghavag: we update the firmware from capsule  image, we identify using the GUID, to identify we need to declare a GUID, correct ? I reused this same GUID to update current version as well.

Do you suggest creating a Mapping this GUID to string again?

gdong1: we should separate "getting firmware version from current SBL" and "getting firmware version from capsule". For the first one, since there is GUID in current SBL to identify a component, so it is nature to use Signature instead of a GUID.
In the firmware update payload, currently we use a GUID to identify a component, so we could do convert in firmware update payload. Later, we might update capsule layout to directly use signature to identify a component in new capsule image. so we don't need change this API at that time.

BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
```diff
@@ -420,6 +424,13 @@ UpdateFwst (
     FwstAcpiTablePtr->EsrtTableEntry[Count].FwType = ESRT_FW_TYPE_SYSTEMFIRMWARE;
     FwstAcpiTablePtr->EsrtTableEntry[Count].LastAttemptVersion = FwUpdCompStatus->LastAttemptVersion;
     FwstAcpiTablePtr->EsrtTableEntry[Count].LastAttemptStatus = FwUpdCompStatus->LastAttemptStatus;
+    if (CompareGuid(&(FwUpdCompStatus->FirmwareId), &gSblFWUpdateImageFileGuid) == TRUE) {
+      Version = ((VerInfoTbl->ImageVersion.CoreMajorVersion << 24) | (VerInfoTbl->ImageVersion.CoreMinorVersion << 16) | \
+                (VerInfoTbl->ImageVersion.ProjMajorVersion << 8) | (VerInfoTbl->ImageVersion.ProjMinorVersion));
+      FwstAcpiTablePtr->EsrtTableEntry[Count].FwVersion = Version;
+    } else {
+      PlatformUpdateCurrentVersion(&(FwUpdCompStatus->FirmwareId), &(FwstAcpiTablePtr->EsrtTableEntry[Count].FwVersion));
```
myguitar: All platforms must have this function even if this is not used. What about creating a VersionLib? Each platforms set version info for each fw components and UpdateFwst gets version thru VersionLib.

PR #226, opened 2019-07-09
[APL] Offer full GPIO list for PlatformID selection

stalamudupula:
Config data should offer a full list of GPIOs for the
end-user to be able to select for PlatformID.


Platform/ApollolakeBoardPkg/CfgData/CfgData_GpioPinOption.dsc
```diff
@@ -30,4 +30,4 @@
 # !BSF OPTION:{+0x018C4:GPIO_190, 0x020C4:GPIO_191, 0x028C4:GPIO_192, 0x030C4:GPIO_193, 0x038C4:GPIO_194, 0x040C4:GPIO_195, 0x048C4:GPIO_196, 0x050C4:GPIO_197, }
 # !BSF OPTION:{+0x058C4:GPIO_198, 0x060C4:GPIO_199, 0x068C4:GPIO_200, 0x070C4:GPIO_201, 0x078C4:GPIO_202, 0x080C4:GPIO_203, 0x088C4:GPIO_204, 0x000C0:GPIO_205, }
 # !BSF OPTION:{+0x008C0:GPIO_206, 0x010C0:GPIO_207, 0x018C0:GPIO_208, 0x0D0C7:GPIO_209, 0x0D8C7:GPIO_210, 0x0E0C7:GPIO_211, 0x0E8C7:GPIO_212, 0x0D0C4:GPIO_213, }
-# !BSF OPTION:{+0x0D8C4:GPIO_214, 0x0E0C4:GPIO_215}
+# !BSF OPTION:{+0x0D8C4:GPIO_214, 0x0E0C4:GPIO_215, 0x0F0C4:GPIO_PMIC_STDBY,            0x0C8C4:GPIO_PMIC_RESET_B,          0x0C0C4:GPIO_PMIC_PWRGOOD}
```
myguitar: Remove unnecessary spaces

Platform/ApollolakeBoardPkg/CfgData/Template_CfgData.dsc
```diff
@@ -58,7 +58,30 @@
 
       # !BSF NAME:{Pid Pad Info}
       # !BSF TYPE:{Combo}
-      # !BSF OPTION:{0x0F0C4:GPIO_PMIC_STDBY, 0x0D0C4:GPIO_213, 0x0C8C4:GPIO_PMIC_RESET_B, 0x0C0C4:GPIO_PMIC_PWRGOOD}
+      # !BSF OPTION:{ 0x000C5:GPIO_0, 0x008C5:GPIO_1, 0x010C5:GPIO_2, 0x018C5:GPIO_3, 0x020C5:GPIO_4, 0x028C5:GPIO_5, 0x030C5:GPIO_6, 0x038C5:GPIO_7, }
```
myguitar: It looks like every options are the same as PinOption's one.  Can we make a single one and re-use it? Like using include GpioOption file? Or better idea?

gdong1: Agree with Aiden. should use "!include CfgData_GpioPinOption.dsc" here. 
but it looks there is other bug to support include file in template. so we could use include after tool issue is fixed.

PR #229, opened 2019-07-18
[CFL] Restore regsiters on S3 resume

stalamudupula:
Restore the register information saved during the
normal boot, on S3 resume path for UEFI payload only.

### DIFF
--- a/BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
+++ b/BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
@@ -120,22 +120,6 @@ typedef struct {
   UINT8         BootPartition;
 } S3_DATA;
 
-typedef struct {
-  UINT32  ApicId;
-  UINT32  SmmBase;
-} CPU_SMMBASE;
-
-typedef struct {
-  UINT32       Signature;
-  UINT8        CpuEntry;
-  UINT8        Unused;
-  UINT8        SmiCtrlRegType;
-  UINT8        SmiCtrlRegWidth;
-  UINT32       SmiCtrlRegAddr;
-  UINT32       SmiCtrlRegVal;
-  CPU_SMMBASE  SmmBase[];
-} SMMBASE_INFO;
-
 #pragma pack()
 
 //
--- a/BootloaderCorePkg/Include/Library/MpInitLib.h
+++ b/BootloaderCorePkg/Include/Library/MpInitLib.h
@@ -6,6 +6,7 @@
 **/
 
 #ifndef _MP_INIT_LIB_H_
+#define _MP_INIT_LIB_H_
 
 typedef enum {
   EnumMpInitNull   = 0x00,
--- a/BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
+++ b/BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
@@ -7,6 +7,7 @@
 **/
 #include <MpInitLibInternal.h>
 #include <Library/BootloaderCoreLib.h>
+#include <Library/S3SaveRestoreLib.h>
 
 MP_ASSEMBLY_ADDRESS_MAP            mAddressMap;
 ALL_CPU_INFO                       mSysCpuInfo;
@@ -74,18 +75,14 @@ CpuInit (
 
   if (PcdGet8 (PcdSmmRebaseMode) == SMM_REBASE_ENABLE_ON_S3_RESUME_ONLY) {
     if (GetBootMode() == BOOT_ON_S3_RESUME) {
-      SmmBaseInfo = (SMMBASE_INFO *)PcdGet32 (PcdSmramTsegBase);
-      if (SmmBaseInfo->Signature == PLD_TO_LDR_SMM_SIG) {
-        for (CpuIdx = 0; CpuIdx < SmmBaseInfo->CpuEntry; CpuIdx++) {
-          if (ApicId == SmmBaseInfo->SmmBase[CpuIdx].ApicId) {
-            SmmRebase (Index, ApicId, SmmBaseInfo->SmmBase[CpuIdx].SmmBase);
-            break;
-          }
-        }
-        if (CpuIdx == SmmBaseInfo->CpuEntry) {
-          ASSERT (CpuIdx < SmmBaseInfo->CpuEntry);
+      SmmBaseInfo = (SMMBASE_INFO *) FindS3Info (SMMBASE_INFO_COMM_ID);
+      for (CpuIdx = 0; CpuIdx < SmmBaseInfo->SmmBaseHdr.Count; CpuIdx++) {
+        if (ApicId == SmmBaseInfo->SmmBase[CpuIdx].ApicId) {
+          SmmRebase (Index, ApicId, SmmBaseInfo->SmmBase[CpuIdx].SmmBase);
+          break;
         }
       }
+      ASSERT (CpuIdx < SmmBaseInfo->SmmBaseHdr.Count);
     }
   } else if (PcdGet8 (PcdSmmRebaseMode) == SMM_REBASE_ENABLE) {
     SmmRebase (Index, ApicId, 0);
--- a/BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
+++ b/BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
@@ -1,0 +1,185 @@
+/** @file
+
+  Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <PiPei.h>
+#include <Library/DebugLib.h>
+#include <Library/HobLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/S3SaveRestoreLib.h>
+#include <Library/IoLib.h>
+#include <Guid/SmmInformationGuid.h>
+#include <Library/BoardInitLib.h>
+#include <Guid/SmmInformationGuid.h>
+#include <Base.h>
+
+typedef UINT32 (*REG_READ)  (UINTN);
+typedef VOID   (*REG_WRITE) (UINTN, UINT32);
+
+#define NUM_TYPE    2
+#define NUM_WIDTH   3
+
+const REG_WRITE  mRegWrite[NUM_TYPE][NUM_WIDTH] = {
+  { (REG_WRITE) MmioWrite8, (REG_WRITE) MmioWrite16, (REG_WRITE) MmioWrite32 },
+  { (REG_WRITE) IoWrite8, (REG_WRITE) IoWrite16, (REG_WRITE) IoWrite32 }
+};
+
+const REG_READ mRegRead[NUM_TYPE][NUM_WIDTH] = {
+  { (REG_READ) MmioRead8, (REG_READ) MmioRead16, (REG_READ) MmioRead32 },
+  { (REG_READ) IoRead8, (REG_READ) IoRead16, (REG_READ) IoRead32 }
+};
+
+const UINT8 mWidthToIdx[NUM_WIDTH][2] = {
+  { 0, WIDE8  },
+  { 1, WIDE16 },
+  { 2, WIDE32 }
+};
+
+/**
+  This function appends information in TSEG area
+  designated for S3 save/restore purpose.
+
+  @param    DataPtr               Address of the structure to be copied to TSEG
+
+  @retval   EFI_OUT_OF_RESOURCES  If SmmSize is exceeding 4KiB
+  @retval   EFI_OUT_OF_RESOURCES  If appeding new struct exceeds SmmSize
+  @retval   EFI_SUCCESS           Append is successful
+
+**/
+EFI_STATUS
+EFIAPI
+AppendS3Info (
+  IN  VOID    *DataPtr
+  )
+{
+  LDR_SMM_INFO      LdrSmmInfo;
+  UINT8             *SmmBase;
+  UINT32            SmmSize;
+  BL_PLD_COMM_HDR   *CommHdr;
+
+  PlatformUpdateHobInfo (&gSmmInformationGuid, &LdrSmmInfo);
+  SmmBase = (UINT8 *) LdrSmmInfo.SmmBase;
+  if (LdrSmmInfo.Flags & SMM_FLAGS_4KB_COMMUNICATION) {
+    SmmSize = SIZE_4KB;
+  } else {
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  CommHdr = (BL_PLD_COMM_HDR *) SmmBase;
+  while ( ((UINT8 *)CommHdr < SmmBase + SmmSize) &&
+          CommHdr->Signature == BL_PLD_COMM_SIG) {
+    CommHdr = (BL_PLD_COMM_HDR *) ((UINT8 *)CommHdr + CommHdr->TotalSize);
+  }
+
+  if ( (UINT8 *)CommHdr + ((BL_PLD_COMM_HDR*)DataPtr)->TotalSize > SmmBase + SmmSize ) {
+    return EFI_OUT_OF_RESOURCES;
+  }
+  CopyMem ((VOID *)CommHdr, DataPtr, ((BL_PLD_COMM_HDR*)DataPtr)->TotalSize);
+
+  return EFI_SUCCESS;
+}
+
+
+/**
+  Get pointer to a specific struct area in TSEG
+
+  @param      Id                Id of the struct to be retreived
+
+  @retval     NULL              Struct with signature not found
+  @retval     Ptr               Pointer to struct with signature
+
+**/
+VOID *
+EFIAPI
+FindS3Info (
+  IN  UINT8   Id
+  )
+{
+  LDR_SMM_INFO      LdrSmmInfo;
+  UINT8             *SmmBase;
+  UINT32            SmmSize;
+  BOOLEAN           FoundInfo = FALSE;
+  BL_PLD_COMM_HDR   *CommHdr;
+
+  PlatformUpdateHobInfo (&gSmmInformationGuid, &LdrSmmInfo);
+  SmmBase = (UINT8 *) LdrSmmInfo.SmmBase;
+  if (LdrSmmInfo.Flags & SMM_FLAGS_4KB_COMMUNICATION) {
+    SmmSize = SIZE_4KB;
+  } else {
+    DEBUG((DEBUG_INFO, "Invalid Communication area size\n"));
+    return NULL;
+  }
+
+  CommHdr = (BL_PLD_COMM_HDR *) SmmBase;
+  while ( ((UINT8 *)CommHdr < SmmBase + SmmSize) &&
+          CommHdr->Signature == BL_PLD_COMM_SIG) {
+    if (CommHdr->Id == Id) {
+      FoundInfo = TRUE;
+      break;
+    }
+    CommHdr = (BL_PLD_COMM_HDR *) ((UINT8 *)CommHdr + CommHdr->TotalSize);
+  }
+
+  if (!FoundInfo) {
+    return NULL;
+  }
+
+  return (VOID *) CommHdr;
+}
+
+
+/**
+  This function restores the states of the registers that were
+  set to be saved by the bootloader in the normal boot path.
+  After the payload has updated the 'Val' field of the reg info
+  in the normal boot path, we simply OR the updated 'Val'
+  with the existing vale of the register in the S3 resume boot path.
+  This function is only called in the S3 resume path.
+
+  @param    S3SaveReg               S3_SAVE_REG info offset
+    
+  @retval   EFI_INVALID_PARAMETER   Invalid pointer to S3_SAVE_REG in Communicaton region
+  @retval   EFI_INVALID_PARAMETER   Invalid Type and Width
+  @retval   EFI_SUCCESS             Restore successful
+
+**/
+EFI_STATUS
+EFIAPI
+RestoreS3RegInfo (
+  IN  S3_SAVE_REG   *S3SaveReg
+  )
+{
+  UINT8     Index;
+  UINT32    Data32;
+  UINT8     Type;
+  UINT8     Width;
+  UINT8     Idx;
+
+  if (S3SaveReg == NULL || S3SaveReg->S3SaveHdr.Id != S3_SAVE_REG_COMM_ID) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  for (Index = 0; Index < S3SaveReg->S3SaveHdr.Count; Index++) {
+    if (S3SaveReg->RegInfo[Index].Addr != 0x00) {
+      Type = S3SaveReg->RegInfo[Index].Type;
+      Width = S3SaveReg->RegInfo[Index].Width;
+      for (Idx = 0; Idx < NUM_WIDTH; Idx++) {
+        if (mWidthToIdx[Idx][1] == Width) {
+          Width = Idx;
+          break;
+        }
+      }
+      if (Type >= NUM_TYPE || Width >= NUM_WIDTH) {
+        return EFI_INVALID_PARAMETER;
+      }
+      mRegWrite[Type][Width] (S3SaveReg->RegInfo[Index].Addr, S3SaveReg->RegInfo[Index].Val);
+      Data32 = mRegRead[Type][Width] (S3SaveReg->RegInfo[Index].Addr);
+      DEBUG ((DEBUG_INFO, "Value after restore reg @ 0x%08X=0x%08X\n", S3SaveReg->RegInfo[Index].Addr, Data32));
+    }
+  }
+
+  return EFI_SUCCESS;
+}
--- a/BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestoreLib.inf
+++ b/BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestoreLib.inf
@@ -1,0 +1,39 @@
+## @file
+#
+#  Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 0x00010005
+  BASE_NAME                      = S3SaveRestoreLib
+  FILE_GUID                      = 6DBF359B-DC7D-4009-9827-D8C736157706
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = S3SaveRestoreLib
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64
+#
+
+[Sources]
+  S3SaveRestore.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  BootloaderCorePkg/BootloaderCorePkg.dec
+  BootloaderCommonPkg/BootloaderCommonPkg.dec
+
+[LibraryClasses]
+  BaseLib
+  DebugLib
+  HobLib
+  BootloaderCoreLib
+
+[Guids]
+  gSmmInformationGuid
+
+[Pcd]
--- a/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -61,6 +61,7 @@
 #include <Library/SmbiosInitLib.h>
 #include <IndustryStandard/SmBios.h>
 #include <VerInfo.h>
+#include <Library/S3SaveRestoreLib.h>
 
 #define DEFAULT_GPIO_IRQ_ROUTE                      14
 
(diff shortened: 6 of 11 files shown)


BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
```diff
@@ -120,24 +120,80 @@ typedef struct {
   UINT8         BootPartition;
 } S3_DATA;
 
+//
+// Format to share info between bootloader and UEFI payload.
```
mauricema: Let me do not mention UEFI payload here.  The code should be applicable for any payload that will do SMM base initialization.

BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
```diff
@@ -120,24 +120,80 @@ typedef struct {
...
+      //  |       SmmBase[n]       |                    V
+      //  -----------------------------------------------
+
+typedef struct {
+  UINT8         Type:4;
+  UINT8         Width:4;
+  UINT8         Rsvd[3];
+  UINT32        Addr;
+  UINT32        Val;
+} REG_INFO;
```
mauricema: why do we put these in BootloaderCoreLib.h ?   Does not seem to be the right place.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -253,29 +253,81 @@ EnableLegacyRegions (
...
+RestoreS3Regs (
+  VOID
+)
+{
+  S3_SAVE_REG           *S3SaveReg;
+  UINT8                 Index;
+  UINT32                Data32;
+  UINT16                Data16;
+
+  S3SaveReg = (S3_SAVE_REG *) (MmioRead32 (TO_MM_PCI_ADDRESS (0x00000000) + TSEG) & ~0xF);
```
mauricema: you need to verify the SaSaveReg pointer is valid.   Check for NULL, and also verify the signature S3_SAVE_REG_SIG.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -253,29 +253,81 @@ EnableLegacyRegions (
...
+      // Check register type and width
+      //
+      if ( ((S3SaveReg->RegInfo[Index].Type != IO) && (S3SaveReg->RegInfo[Index].Type != MEM)) ||
+           ((S3SaveReg->RegInfo[Index].Width != WIDE32) && (S3SaveReg->RegInfo[Index].Width != WIDE16)) ) {
+        DEBUG ((DEBUG_ERROR, "ERROR: not supported regtype (0x%x) or regwidth (0x%x)\n",\
+        S3SaveReg->RegInfo[Index].Type, S3SaveReg->RegInfo[Index].Width));
+        return ;
+      }
+      //
+      // Restore registers saved by UEFI payload
```
mauricema: should not mention UEFI payload here.

BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
stalamudupula: Corrected.

BootloaderCorePkg/Include/Library/BootloaderCoreLib.h
stalamudupula: Moved to a new file.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
stalamudupula: Added the checks.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
stalamudupula: Corrected.

BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
```diff
@@ -74,16 +75,18 @@ CpuInit (
...
           if (ApicId == SmmBaseInfo->SmmBase[CpuIdx].ApicId) {
             SmmRebase (Index, ApicId, SmmBaseInfo->SmmBase[CpuIdx].SmmBase);
             break;
           }
         }
-        if (CpuIdx == SmmBaseInfo->CpuEntry) {
-          ASSERT (CpuIdx < SmmBaseInfo->CpuEntry);
+        if (CpuIdx == SmmBaseInfo->CpuCnt) {
+          ASSERT (CpuIdx < SmmBaseInfo->CpuCnt);
         }
```
gdong1: Looks it will never ASSERT since it runs to it only when CpuIdx == SmmBaseInfo->CpuCnt.
do you mean ASSERT (FALSE)?

stalamudupula: It can ASSERT when the ApicId for a proc is missing in the ApicIds reported via TSEG region by the payload.

gdong1: Looks there is duplication there are two conditions.
remove If condition or remove condition inside ASSERT would be more clear.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1042,8 +1043,10 @@ BoardInit (
 
     break;
   case EndOfFirmware:
+    ClearSmi ();
     if ((GetBootMode() == BOOT_ON_S3_RESUME) && (GetPayloadId () == UEFI_PAYLOAD_ID_SIGNATURE)) {
-      EnableSmi ();
+      SmmBaseInfo = (SMMBASE_INFO *) (MmioRead32 (TO_MM_PCI_ADDRESS (0x00000000) + TSEG) & ~0xF);
+      S3RestoreRegData ((UINT8 *)SmmBaseInfo);
```
myguitar: How do we handle UpdateSmmInfo()?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,131 @@
...
+#include <Library/DebugLib.h>
+#include <Library/HobLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/SocInitLib.h>
+#include <Library/MpInitLib.h>
+#include <Library/IoLib.h>
+#include <Guid/SmmInformationGuid.h>
+
+
+S3_SAVE_REG   *mS3SaveReg = NULL;
```
myguitar: Can we have better way to avoid global variable in Library?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,131 @@
...
+VOID
+EFIAPI
+S3SaveRegData (
+  IN  UINT8     *SmmBaseInfo,
+  IN  UINT8     Type,
+  IN  UINT8     Width,
+  IN  UINT32    Addr
+  )
+{
+  //S3_SAVE_REG   *S3SaveReg;
```
myguitar: Remove unnecessary line.

BootloaderCorePkg/Include/Library/SocInitLib.h
```diff
@@ -7,6 +7,110 @@
...
+S3SaveRegData (
+  IN  UINT8     *SmmBaseInfo,
+  IN  UINT8     Type,
+  IN  UINT8     Width,
+  IN  UINT32    Addr
+  );
+
+/**
+  This function restores the states of the registers that were
+  set to be saved by the bootloader inthe normal boot path.
```
gdong1: Add a space for "inthe"

BootloaderCorePkg/Include/Library/MpInitLib.h
```diff
@@ -46,6 +46,29 @@ typedef struct {
...
+  UINT32        Signature;
+  UINT8         CpuCnt;
+  UINT8         Unused;
+  UINT16        Rsvd;
+  CPU_SMMBASE   SmmBase[];
+} SMMBASE_INFO;
+
+#pragma pack()
+
+
```
gdong1: Given added a new library S3SaveRestoreLib, why S3SaveRestoreLib.h is missing?
Prefer moving these definition to S3SaveRestoreLib.h

BootloaderCorePkg/Include/Library/SocInitLib.h
```diff
@@ -7,6 +7,110 @@
...
+  @retval   None
+**/
+VOID
+EFIAPI
+S3SaveRegData (
+  IN  UINT8     *SmmBaseInfo,
+  IN  UINT8     Type,
+  IN  UINT8     Width,
+  IN  UINT32    Addr
+  );
```
gdong1: Prefer moving API and definition to S3SaveRestoreLib.h

BootloaderCorePkg/Include/Library/SocInitLib.h
```diff
@@ -7,6 +7,110 @@
...
+  UINT8         Rsvd[2];
+  UINT32        Addr;
+  UINT32        Val;
+} REG_INFO;
+
+typedef struct {
+  UINT32        Signature;
+  UINT32        RegCnt;
+  REG_INFO      RegInfo[];
+} S3_SAVE_REG;
```
gdong1: It is not consistent by defining RegCnt UINT32 and CpuCnt UINT8.
It would be easier to parse a structure using a common header for all the structures in S3 SAVE area.
e.g.:
typedef struct {
  UINT32        Signature;
  UINT16        Count;
  UINT16        TotalSize;
} S3_SAVE_HEADER;

BootloaderCorePkg/Include/Library/MpInitLib.h
stalamudupula: In the previous discussions, it was decided that this was CPU related info and a better place was in MpInitLib.h

BootloaderCorePkg/Include/Library/SocInitLib.h
stalamudupula: Makes sense, will make a common header for S3_SAVE_REG and SMMBASE_INFO strcutres.

BootloaderCorePkg/Include/Library/SocInitLib.h
stalamudupula: Will do.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: Just wanted to reuse the S3SaveReg offset calculated the first time. But assuming there wont be many registers to be saved for S3 resume, if at all, will just calculate this offset each time we call into this function. We can avoid global here.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
stalamudupula: UpdateSmmInfo() is called only on Normal boot path.

BootloaderCorePkg/Include/Library/MpInitLib.h
gdong1: Since now we have S3SaveRestoreLib, does it make more sense to move to S3SaveRestoreLib?

stalamudupula: In that case, MpInitLib has to include S3SaveRestoreLib header. Ideally it does not need to know S3 related information. So, decided to separate S3 related structures from including in MP related files. Please let me know if you have still have a concern.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,160 @@
...
+  @param    SmmBaseInfo   Pointer to TSEG region
+  @param    DataPtr       Address of the structure to be copied to TSEG
+
+  @retval   None
+**/
+VOID
+EFIAPI
+S3SetData (
+  IN  VOID    *SmmBaseInfo,
+  IN  UINT32  SmmSize,
```
gdong1: Could we get SMM communication inside this function instead from caller? This will make caller friendly use.
The communication area could be got using common API PlatformUpdateHobInfo ().

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,160 @@
...
+  IN  VOID    *DataPtr
+  )
+{
+  UINT8             *Ptr;
+
+  if ( ((BL_PLD_COMM_HDR *)SmmBaseInfo)->Signature != SMM_BASE_INFO_SIG) {
+    DEBUG((DEBUG_ERROR, "Add SMMBASE_INFO region first\n"));
+    return;
+  }
```
gdong1: we don't need check there. Let the platform to set SmmBase info firstly if the platform wants to make it as the first record.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: PlatformUpdateHobInfo() doesnt prepare the SMM Hob until BuildExtraInfoHob(), which takes place much later then PrePayLoadLoading phase. Infact there isnt any board phase after BuildExtraInfoHob().

gdong1: PlatformUpdateHobInfo() doesn't depend on BuildExtraInfoHob(). In the library it could call PlatformUpdateHobInfo()  directly with parameter LDR_SMM_INFO *.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,162 @@
...
+  )
+{
+  UINT8             *Ptr;
+  LDR_SMM_INFO      LdrSmmInfo;
+  UINT8             *SmmBase;
+  UINT32            SmmSize;
+
+  PlatformUpdateHobInfo (&gSmmInformationGuid, &LdrSmmInfo);
+  SmmBase = (UINT8 *) LdrSmmInfo.SmmBase;
+  SmmSize = LdrSmmInfo.SmmSize;
```
gdong1: The SmmSize here is the whole SMM range size.
need check if flag SMM_FLAGS_4KB_COMMUNICATION is set in LdrSmmInfo.Flags. if yes. set SmmSize to 4KB. else return.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,162 @@
...
+
+  for (Ptr = SmmBase; Ptr < SmmBase + SmmSize; ) {
+    if ( ((BL_PLD_COMM_HDR *)Ptr)->Signature != 0x00000000) {
```
gdong1: where SMM communication area is cleared?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,162 @@
...
+  UINT32            SmmSize;
+
+  PlatformUpdateHobInfo (&gSmmInformationGuid, &LdrSmmInfo);
+  SmmBase = (UINT8 *) LdrSmmInfo.SmmBase;
+  SmmSize = LdrSmmInfo.SmmSize;
+
+  if ( ((BL_PLD_COMM_HDR *)SmmBase)->Signature != SMM_BASE_INFO_SIG) {
+    DEBUG((DEBUG_ERROR, "Invalid communication area\n"));
+    return NULL;
+  }
```
gdong1: Here it assume SMM_BASE_INFO_SIG is the first record. Better to be decided by platform. Should we remove this check?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,162 @@
...
+
+  for (Ptr = SmmBase; Ptr < SmmBase + SmmSize; ) {
+    if ( ((BL_PLD_COMM_HDR *)Ptr)->Signature != Signature) {
+      Ptr += ((BL_PLD_COMM_HDR *)Ptr)->TotalSize;
+      continue;
+    }
+    break;
+  }
+
+  return (VOID *) Ptr;
```
gdong1: Should it return NULL if it could not find?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: Good catch. Thanks, will fix it.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,162 @@
...
+
+  @param    SmmBaseInfo   TSEG Base
+    
+  @retval   None
+
+**/
+VOID
+EFIAPI
+RestoreS3RegInfo (
+  IN  S3_SAVE_REG   *S3SaveReg
```
gdong1: Parameters in function comment doesn't match actual name.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestoreLib.inf
```diff
@@ -0,0 +1,40 @@
+## @file
+#
+#  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
```
gdong1: change to 2019?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestoreLib.inf
```diff
@@ -0,0 +1,40 @@
...
+  FILE_GUID                      = 6DBF359B-DC7D-4009-9827-D8C736157706  
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = S3SaveRestoreLib
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+#  VALID_ARCHITECTURES           = IA32 X64 IPF
+#
```
gdong1: Remove IPF

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: ZeroMem() by platform code, before copying the first structure.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: OK, will remove it.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: Yes, its an error. Will fix it.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: ok, will change.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -971,6 +969,25 @@ BoardInit (
     if (EFI_ERROR (Status)) {
       DEBUG ((DEBUG_WARN, "VBT not found %r\n", Status));
     }
+    if (GetPayloadId () == UEFI_PAYLOAD_ID_SIGNATURE && GetBootMode() != BOOT_ON_S3_RESUME) {
+      //
+      // Update TSEG region with S3 save/restore info
+      //
+      SmmBaseInfo = (SMMBASE_INFO *) (MmioRead32 (TO_MM_PCI_ADDRESS (0x00000000) + TSEG) & ~0xF);
+      SmmSize = (MmioRead32 (TO_MM_PCI_ADDRESS (0x00000000) + BGSM) & ~0xF) - (UINT32)SmmBaseInfo;
+      ZeroMem ((VOID *)SmmBaseInfo, SmmSize);
```
gdong1: Better not have this zero memory. often SMRAM is 8MB or 16MB. it impact performance.
also it makes the S3 lib depends on platform to zero the library.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -971,6 +969,25 @@ BoardInit (
...
+      //
+      // Set SMMBASE_INFO dummy strucutre in TSEG before others
+      //
+      SmmBaseInfo->SmmBaseHdr.Signature = SMM_BASE_INFO_SIG;
+      SmmBaseInfo->SmmBaseHdr.Count     = (UINT16) MpGetInfo()->CpuCount;
+      SmmBaseInfo->SmmBaseHdr.TotalSize = sizeof(BL_PLD_COMM_HDR) + SmmBaseInfo->SmmBaseHdr.Count * sizeof(CPU_SMMBASE);
```
gdong1: Better to call the Set API to set SMM_BASE_INFO_SIG.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestoreLib.inf
```diff
@@ -0,0 +1,40 @@
...
+
+[Sources]
+  S3SaveRestore.c
+
+[Packages]
+  MdePkg/MdePkg.dec 
+  IntelFsp2Pkg/IntelFsp2Pkg.dec
```
gdong1: why has  IntelFsp2Pkg/IntelFsp2Pkg.dec?

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1042,8 +1056,9 @@ BoardInit (
 
     break;
   case EndOfFirmware:
+    ClearSmi ();
     if ((GetBootMode() == BOOT_ON_S3_RESUME) && (GetPayloadId () == UEFI_PAYLOAD_ID_SIGNATURE)) {
-      EnableSmi ();
+      RestoreS3RegInfo (GetS3Info(S3_SAVE_REG_COMM_ID));
```
mauricema: Add a white space between GetS3Info and '('

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -838,6 +829,16 @@ InitializeSmbiosInfo (
   return EFI_SUCCESS;
 }
 
+
+STATIC SMMBASE_INFO SmmBaseInfo = {
```
mauricema: module global variables need to be  prefixed with "m"

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,175 @@
...
+  PlatformUpdateHobInfo (&gSmmInformationGuid, &LdrSmmInfo);
+  SmmBase = (UINT8 *) LdrSmmInfo.SmmBase;
+  if (LdrSmmInfo.Flags & SMM_FLAGS_4KB_COMMUNICATION) {
+    SmmSize = SIZE_4KB;
+  } else {
+    DEBUG((DEBUG_ERROR, "Invalid Communication area size\n"));
+    return;
+  }
+
+  for (Ptr = SmmBase; Ptr < SmmBase + SmmSize; ) {
```
mauricema: The available size at TSEG is very limited.  I think we only reserved 4KB for it.   So we need to make sure we cannot pass the reserved boundary.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,175 @@
...
+    if ( ((BL_PLD_COMM_HDR *)Ptr)->Signature == BL_PLD_COMM_SIG) {
```
mauricema: For coding, better to define a new pointer for BL_LD_COMM_HDR instead of typecasting everywhere.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
mauricema: Please ignore my comments.   The SmmSize is already set to 4KB.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,175 @@
...
+  This function uploads information in TSEG area
+  designated for S3 save/restore purpose.
+
+  @param    DataPtr       Address of the structure to be copied to TSEG
+
+  @retval   None
+**/
+VOID
+EFIAPI
+SetS3Info (
```
mauricema: More exactly, you are doing appending.
Maybe rename to  AppendS3Info

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,175 @@
...
+  PlatformUpdateHobInfo (&gSmmInformationGuid, &LdrSmmInfo);
+  SmmBase = (UINT8 *) LdrSmmInfo.SmmBase;
+  if (LdrSmmInfo.Flags & SMM_FLAGS_4KB_COMMUNICATION) {
+    SmmSize = SIZE_4KB;
+  } else {
+    DEBUG((DEBUG_ERROR, "Invalid Communication area size\n"));
+    return NULL;
+  }
+
+  for (Ptr = SmmBase; Ptr < SmmBase + SmmSize; ) {
```
mauricema: There is no reason to search the whole SmmSize region.  If as soon as BL_PLD_COMM_SIG is not seen,  the code can break out.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,175 @@
...
+    return;
+  }
+
+  for (Index = 0; Index < S3SaveReg->S3SaveHdr.Count; Index++) {
+    if (S3SaveReg->RegInfo[Index].Addr != 0x00) {
+      //
+      // Check register type and width
+      //
+      if ( ((S3SaveReg->RegInfo[Index].Type != IO) && (S3SaveReg->RegInfo[Index].Type != MEM)) ||
+           ((S3SaveReg->RegInfo[Index].Width != WIDE32) && (S3SaveReg->RegInfo[Index].Width != WIDE16)) ) {
```
mauricema: Any reason why  WIDE8 is not supported ?

stalamudupula: No specific reason, we can add support of 8-bit wide registers if wee need to save one such reg. Currently, we are only saving 32-bit wide registers.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,180 @@
...
+        DEBUG ((DEBUG_ERROR, "ERROR: not supported regtype (0x%x) or regwidth (0x%x)\n",\
+        S3SaveReg->RegInfo[Index].Type, S3SaveReg->RegInfo[Index].Width));
+        return ;
+      }
+      //
+      // Restore registers saved by payload
+      //
+      if (S3SaveReg->RegInfo[Index].Width == WIDE16) {
```
myguitar: Better to use macro.

ex)
#define RESTORE_S3_REGS(TYPE, SIZE, ADDR, VALUE) \
  TYPE##Or##SIZE(ADDR, (UINT##SIZE)VALUE); \
  (VOID)TYPE##Read##SIZE(ADDR)

RESTORE_S3_REGS (type == Mem ? Mmio : Io, Width * 8, Addr, Val);

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,194 @@
...
+  for (Index = 0; Index < S3SaveReg->S3SaveHdr.Count; Index++) {
+    if (S3SaveReg->RegInfo[Index].Addr != 0x00) {
+      //
+      // Check register type and width
+      //
+      Type = S3SaveReg->RegInfo[Index].Type;
+      Width = S3SaveReg->RegInfo[Index].Width;
+      if ( ((Type != IO) && (Type != MEM)) ||
+           ((Width != WIDE32) && (Width != WIDE16)) ) {
+        DEBUG ((DEBUG_ERROR, "ERROR: not supported regtype (0x%x) or regwidth (0x%x)\n", Type, Width));
```
mauricema: since you do have error condition,  better to return Status  instead. 
In general we should not print error message too often since it comes with a size cost.  We can use status to return an error condition.  
Also since you are using array, better to compare with the array size to prevent access overflow.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  return (VOID *) CommHdr;
+}
+
+
+typedef UINT32 (*REG_READ)  (UINTN);
+typedef VOID   (*REG_WRITE) (UINTN, UINT32);
+
+STATIC REG_WRITE  mRegWrite[2][3] = {
```
mauricema: pls follow coding style

PR #230, opened 2019-07-18
Enhance SPI Flash Library

gdong1:
1. Add region type FlashRegionAll support
2. Add Hardware sequence check


gdong1:
> The changes look good.
> BTW, might be better to move this library into Silicon/CommonSocPkg.
Agree to move to CommonSocPkg. let's do it late in a separate patch.

PR #231, opened 2019-07-19
Modify kernel size loading check

jgutbub:
When loading a Linux kernel we should not
limit the size of the kernel being loaded
to 15MB but we should verify that the kernel
will not overwrite the payload which can
some times be located in low memory where
the kernel is being loaded.

### DIFF
--- a/PayloadPkg/OsLoader/Linux.c
+++ b/PayloadPkg/OsLoader/Linux.c
@@ -220,7 +220,7 @@ SetupBzImage (
   KernelBuf  = (VOID *) (UINTN)LINUX_KERNEL_BASE;
   KernelSize = Bp->Hdr.SysSize * 16;
   DEBUG ((DEBUG_INFO, "Src=0x%p Dest=0x%p KernelSize=%d\n", Buffer + BootParamSize, KernelBuf, KernelSize));
-  if (KernelSize > IMAGE_MAX_SIZE) {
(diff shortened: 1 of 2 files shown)


PayloadPkg/OsLoader/OsLoader.inf
```diff
@@ -40,6 +40,7 @@
 
 [Packages]
   MdePkg/MdePkg.dec
+  BootloaderCorePkg/BootloaderCorePkg.dec
```
gdong1: The code change looks good.
Remember we try to avoid to use BootloaderCorePkg in OsLoader. is it still the case now?

jgutbub: Not sure, @mauricema any thoughts on that before we merge?

myguitar: I agree with @gdong1. Accessing BootloaderCorePkg from Payload is not a good option.

mauricema: In general, it is true for payload.  But for OsLoader payload, it is considered as special since it is always built together with SBL core code.  
In this case, it is better if we can avoid any reference to BootloaderCorePkg.dec.  I just looked into the code,  since we passed PldBase to OsLoader, can we use this to do the comparison instead of the PCDs ?

mauricema: Never mind.  Noticed that PldBase is only used for PreOsChecker.
I think maybe we can use the symbol  ModuleEntryPoint to get the OsLoader loading base.

PR #232, opened 2019-07-19
Add EXT2F_INCOMPAT_RECOVER to EXT2F_INCOMPAT_SUPP

jgutbub:
Currently there is some limitation with mounting an
EXT4 partition to retrieve a file for booting when
the partition was not properly umounted during the
previous usage (ex. powering off an OS without
using 'poweroff' command, unplugging power, etc.).

There is recovery support in the Ext23Lib but as
a temporary solution we can ignore the recovery
bit to avoid boot issues when the recovery bit is
flipped due to unexpected power loss.


myguitar:
I have never seen this issue even if power loss happens while I have seen OS fsck running. What's the exact behavior and cause?

PR #233, opened 2019-07-23
Append optional boot params

jgutbub:
There may be additional boot params
specified by a platform that needs to
be added without updating the boot
option data (ex. IAS cmdline.txt).
These optional additional boot params
can be included via CFG data, Shell
command, etc., and applied by setting
the mAppendParams pointer.

### DIFF
--- a/PayloadPkg/OsLoader/BootParameters.c
+++ b/PayloadPkg/OsLoader/BootParameters.c
@@ -9,6 +9,8 @@
 
 CONST CHAR8 *mMmcDllStr = "MMCDLL";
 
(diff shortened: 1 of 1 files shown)


PayloadPkg/OsLoader/BootParameters.c
```diff
@@ -9,6 +9,8 @@
 
 CONST CHAR8 *mMmcDllStr = "MMCDLL";
 
+CONST CHAR8 *mAppendParams = NULL;
```
mauricema: This is only for IAS image, right ?
If so, can we make the variable name more specific. EX:  mAppendCmdLineParams

PR #235, opened 2019-07-26
Add a platform flag for SPI flash SMM write disable

gdong1:
When writing flash in SMM mode, some platform need disable SMM write
protection. when this flag is set, it means this platform need disable
SMM write protection.


BootloaderCommonPkg/Include/Guid/LoaderPlatformInfoGuid.h
```diff
@@ -54,7 +62,8 @@ typedef struct {
   UINT16         PlatformId;
   UINT32         CpuCount;
   UINT16         HwState;
-  UINT8          Reserved1[2];
+  UINT8          Flags;
```
mauricema: maybe just define UINT16 for Flags?

gdong1: OK. let me update to UINT16

PR #236, opened 2019-07-29
Add temporary memory management for MemoryAllocationLib

mauricema:
There are cases where temporary memory is required in stages. Since
MemoryAllocationLib instance for stages has no de-allocation function,
it needs special handling. This patch added temporary memory management
into standard MemoryAllocatoinLib interface.


BootloaderCommonPkg/Library/FullMemoryAllocationLib/FullMemoryAllocationLib.c
```diff
@@ -632,4 +632,36 @@ AllocateRuntimePool (
...
+  This function allocates temporary memory pool.
+
+  @param[in]  AllocationSize    The memory pool size to allocate.
+
+  @retval     A pointer to the allocated temporary buffer or NULL if allocation fails.
+
+**/
+VOID *
+EFIAPI
+AllocateTemporaryMemory (
```
gdong1: Comparing with AllocateRuntimePool (), better to rename it to AllocateBootTimePool () or AllocateTemporaryPool ()?  with Pool in the name it is clear that AllocationSize is in byte unit.

mauricema: AllocateTemporaryMemory is paired with FreeTemporaryMemory .  
It think it is clear that it is byte.   But if you insist in  AllocateTemporaryPool(), I can also make the change.

gdong1: OK. It is fine to use it instead of changing FreeTemporaryMemory()

PR #238, opened 2019-07-29
Add ContainerLib to support sub-region

mauricema:
This patch added a ContainerLib to support load and location a
component from a specified container. It copies the component
from flash to memory, authenticate it, and then decompress it if
required. It can also be used to support load component from flash
map in SBL stage2 or payload, such as payload or e-payload.

### DIFF
--- a/BootloaderCommonPkg/BootloaderCommonPkg.dec
+++ b/BootloaderCommonPkg/BootloaderCommonPkg.dec
@@ -50,6 +50,8 @@
   gPlatformCommonLibTokenSpaceGuid.PcdHeciLibId              |          5 |  UINT8 | 0x20000106
   gPlatformCommonLibTokenSpaceGuid.PcdMmcTuningLibId         |          6 |  UINT8 | 0x20000107
 
+  gPlatformCommonLibTokenSpaceGuid.PcdContainerMaxNumber     |          8 | UINT32 | 0x20000120
+
   gPlatformCommonLibTokenSpaceGuid.PcdCpuLocalApicBaseAddress| 0xFEE00000 | UINT32  | 0x20000186
   gPlatformCommonLibTokenSpaceGuid.PcdSupportedMediaTypeMask | 0xFFFFFFFF | UINT32  | 0x20000187
   gPlatformCommonLibTokenSpaceGuid.PcdMmcTuningLba           | 0x00000040 | UINT32  | 0x20000188
--- a/BootloaderCommonPkg/Include/Library/ContainerLib.h
+++ b/BootloaderCommonPkg/Include/Library/ContainerLib.h
@@ -1,0 +1,128 @@
+/** @file
+  Header file for container library implementation.
+
+  Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+#ifndef _CONTAINER_LIB_H_
+#define _CONTAINER_LIB_H_
+
+#include <Library/PcdLib.h>
+
+#define CONTAINER_LIST_SIGNATURE SIGNATURE_32('C','T','N', 'L')
+
+#define AUTH_TYPE_NONE                0
+#define AUTH_TYPE_SHA2_256            1
+#define AUTH_TYPE_SHA2_384            2
+#define AUTH_TYPE_SIG_RSA2048_SHA256  3
+#define AUTH_TYPE_SIG_RSA3072_SHA384  4
+
+typedef struct {
+  UINT32           Signature;
+  UINT32           HeaderCache;
+  UINT32           Base;
+  UINT32           Reserved;
+} CONTAINER_ENTRY;
+
+typedef struct {
+  UINT32           Signature;
+  UINT32           Reserved;
+  UINT32           TotalLength;
+  UINT32           Count;
+  CONTAINER_ENTRY  Entry[0];
+} CONTAINER_LIST;
+
+typedef struct {
+  UINT32           Signature;
+  UINT16           Version;
+  UINT16           DataOffset;
+  UINT32           DataSize;
+  UINT8            AuthType;
+  UINT8            ImageType;
+  UINT8            Flags;
+  UINT8            Count;
+} CONTAINER_HDR;
+
+typedef struct {
+  UINT32           Name;
+  UINT32           Offset;
+  UINT32           Size;
+  UINT8            Rserved;
+  UINT8            Alignment;
+  UINT8            AuthType;
+  UINT8            HashSize;
+  UINT8            HashData[0];
+} COMPONENT_ENTRY;
+
+
+/**
+  Load a component region information from a container.
+
+  @param[in] ContainerSig    Container signature.
+  @param[in] ComponentName   component name.
+  @param[in] Buffer          Pointer to receive component base.
+  @param[in] Length          Pointer to receive component size.
+
+  @retval EFI_UNSUPPORTED          Unsupported AuthType.
+  @retval EFI_NOT_FOUND            Cannot locate component.
+  @retval EFI_SECURITY_VIOLATION   Authentication failed.
+  @retval EFI_SUCCESS              Authentication succeeded.
+
+**/
+EFI_STATUS
+EFIAPI
+LocateComponentRegion (
+  IN     UINT32    ContainerSig,
+  IN     UINT32    ComponentName,
+  OUT    UINT32   *RegionBase,
+  OUT    UINT32   *RegionLength
+  );
+
+/**
+  Locate a component from a container.
+
+  @param[in]     ContainerSig       Container signature.
+  @param[in]     ComponentName      Component name.
+  @param[in,out] ContainerEntryPtr  Pointer to receive container entry info.
+  @param[in,out] ComponentEntryPtr  Pointer to receive component entry info.
+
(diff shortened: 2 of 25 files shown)


BootloaderCommonPkg/BootloaderCommonPkg.dec
```diff
@@ -50,6 +50,8 @@
   gPlatformCommonLibTokenSpaceGuid.PcdHeciLibId              |          5 |  UINT8 | 0x20000106
   gPlatformCommonLibTokenSpaceGuid.PcdMmcTuningLibId         |          6 |  UINT8 | 0x20000107
 
+  gPlatformCommonLibTokenSpaceGuid.PcdContainerMaxNumber     |          8 | UINT32 | 0x20000120
```
gdong1: I am still thinking it makes code and usage a little complicated by having multiple container support.
a) too many layer: Multiple containers -> multiple regions in container -> one or multiple files in image in region.
b) Caller need to know containers beside region name
c) If different container is verified by same key,  will it bring security concern? if verified by different key, need add multiple slot in hashstore for different container.

BootloaderCommonPkg/Include/Library/ContainerLib.h
```diff
@@ -0,0 +1,128 @@
...
+typedef struct {
+  UINT32           Signature;
+  UINT16           Version;
+  UINT16           DataOffset;
+  UINT32           DataSize;
+  UINT8            AuthType;
+  UINT8            ImageType;
+  UINT8            Flags;
+  UINT8            Count;
+} CONTAINER_HDR;
```
gdong1: container is very good for region support. lately if we added signed image using same header, do you think is it still a good name to be shared?

mauricema: CONTAINER is a generic term, I did not see the issue.

BootloaderCommonPkg/BootloaderCommonPkg.dec
mauricema: There are use cases for multiple containers.  
1. Component in containers might have different level of importance.  If A and B in the same container, but A needs redundancy, B does not.  It will cause issue.  If we put container in redundant region, then B will be duplicated and waste flash space.  If we put container in non-redundant region, then A will have no redundancy and break redundant boot.
2. Container might need to be updated or created by different owners.  Grouping all components in one container makes this difficult to support it.  

It is expected to use container signature to load a component.  Just like loading a file, you have to know which directory.

I did not see issue to sign different container header with the same key.  This is only for header signing. Each component inside can still have its own key.

BootloaderCommonPkg/Library/SecureBootLib/SecureBootHash.c
```diff
@@ -30,22 +33,31 @@ RETURN_STATUS
 DoHashVerify (
   IN CONST UINT8           *Data,
   IN       UINT32           Length,
-  IN       UINT8            ComponentType
+  IN       UINT8            ComponentType,
+  IN OUT   UINT8           *Hash
```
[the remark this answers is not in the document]
mauricema: We can add new API for this.  But I prefer changing the existing one to avoid too many APIs left behind. 

For DoHashVerify,  I think we can add a HashAlgorithm parameter to support different hash type.

BootloaderCommonPkg/Include/Library/ContainerLib.h
```diff
@@ -0,0 +1,128 @@
...
+  );
+
+/**
+  Locate a component from a container.
+
+  @param[in] ContainerSig       Container signature.
+  @param[in] ComponentName      Component name.
+  @param[in] ContainerEntryPtr  Pointer to receive container entry info.
+  @param[in] ComponentEntryPtr    Pointer to receive component entry info.
+
```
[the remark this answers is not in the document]
mauricema: will change it.

BootloaderCommonPkg/Include/Library/ContainerLib.h
```diff
@@ -0,0 +1,128 @@
...
+  OUT    COMPONENT_ENTRY       **ComponentEntryPtr
+  );
+
+/**
+  Load a component from a container to memory.
+
+  @param[in] ContainerSig    Container signature or component type.
+  @param[in] ComponentName   Component name.
+  @param[in] Buffer          Pointer to receive component base.
+  @param[in] Length          Pointer to receive component size.
```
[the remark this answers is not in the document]
mauricema: in to in, out ?

BootloaderCommonPkg/BootloaderCommonPkg.dec
gdong1: Looks one container could support the use cases you listed.
1. For the first use case to support redundancy. If A is boot critical (e.g. CfgData), we could put it in regular redundancy region managed by flashmap. If A is not boot critical, we could put A1 and A2 in container. 
2. We will provide tool to generate capsule image for firmwareupdate. From user viewpoint, it is transparent. they only need provide image components.
Given all the containers are signed by same key, and if user could generate and update container. It means one user of a container could update the other containers. In this case, it looks there is not much help by providing different key to sign components inside a container since the container including the component key could be updated.

BootloaderCommonPkg/Library/SecureBootLib/SecureBootHash.c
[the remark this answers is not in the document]
gdong1: agree changing existing API is a clear way. But at some point, we need make sure SBL could be backward compatible in APIs to make sure customer could upgrade SblOpen smoothly. 
For the API we plan to drop, we could print a warning/usage to let user to use new API. then after some time (e.g. a year) we could bump SBL version and drop old API.

BootloaderCommonPkg/BootloaderCommonPkg.dec
mauricema: That will need to handle different layout.   And if you have more components like that, you will be running out of control pretty soon.  I didn't see why multiple containers are complicated.  It is almost the same as support single container.

BootloaderCommonPkg/Library/SecureBootLib/SecureBootHash.c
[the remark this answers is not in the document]
mauricema: In this case,  It should not be a big issue.   The API is in SecureBootLib.  And all secure boot related flow are controlled by the core code.  As long as the core code is updated properly,  it should just work fine.

BootloaderCommonPkg/Include/Library/ContainerLib.h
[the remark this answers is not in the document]
mauricema: changed in new PR

BootloaderCommonPkg/Include/Library/ContainerLib.h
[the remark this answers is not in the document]
mauricema: changed in new PR

BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
```diff
@@ -36,7 +39,8 @@ DoRsaVerify (
   IN       UINT8            ComponentType,
   IN CONST UINT8           *Signature,
   IN       UINT8           *PubKey,
-  OUT      UINT8           *OutHash
+  IN       UINT8           *PubKeyHash      OPTIONAL,
+  OUT      UINT8           *OutHash         OPTIONAL
```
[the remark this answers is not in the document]
mauricema: Will keep it for now.

mauricema:
Uploaded new changes

mauricema:
> Where the container header was verified?
The header is verified during container registration.    Only verified container will be registered successfully.  Please check logic inside LocateComponent().

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -0,0 +1,523 @@
...
+    ContainerHdr     = (CONTAINER_HDR *)ContainerEntry->HeaderCache;
+    ContainerHdrSize = GetContainerHeaderSize (ContainerHdr);
+    if (ContainerHdrSize > 0) {
+      AuthType = ContainerHdr->AuthType;
+      AuthData = (UINT8 *)ContainerHdr + ALIGN_UP(ContainerHdrSize, AUTH_DATA_ALIGN);
+      if ((AuthType == AUTH_TYPE_NONE) && FeaturePcdGet (PcdVerifiedBootEnabled)) {
+        Status = EFI_SECURITY_VIOLATION;
+      } else {
+        Status = AuthenticateComponent ((UINT8 *)ContainerHdr, ContainerHdrSize,
+                                             AuthType, AuthData, NULL, COMP_TYPE_PUBKEY_FWU);
```
gdong1: To verify the container header, the authenticate method should not depend on container itsself since it is not been verified yet.  In this case, if container st the AuthType to None, container could bypass verification.
I think we could verify the container header by RSA2048 if hashstore is using SHA256 or by RSA3072 if hashstore is using SHA384.

mauricema: We do verify the AuthType separately.  When it is NONE,  it will pass the container header check only if VerifiedBoot is disabled.

gdong1: OK. I checked the DoRsaVerify() and DoHashVerify().  To verify the public key of container, it is still using hardcode sha256 now, not using AuthType.

PR #240, opened 2019-07-31
[CFL] Remove PlatformId checks in platform code

stalamudupula:
PlatformId checks for different board support in
platform code should be avoided. Instead use a different
variable to identify cpu for platform init purposes.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
@@ -88,11 +88,13 @@ UpdateFspConfig (
   GPU_CFG_DATA                    *GpuCfgData;
   UINT16                           PlatformId;
   PEG_GPIO_DATA                   *PegGpioData;
+  UINT16                          CpuId;
 
   FspmUpd       = (FSPM_UPD *)FspmUpdPtr;
   Fspmcfg       = &FspmUpd->FspmConfig;
   FspmcfgTest   = &FspmUpd->FspmTestConfig;
   PlatformId    = GetPlatformId ();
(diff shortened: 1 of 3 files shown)


Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -267,18 +269,21 @@ PlatformIdInitialize (
        (CpuDid == V_SA_DEVICE_ID_CFL_DT_3) ||
        (CpuDid == V_SA_DEVICE_ID_CFL_DT_4))) {
       SetPlatformId (PLATFORM_ID_CFL_S);  // CFL-S
+      SetPlatformBomId (CPU_ID_CFL_S);
     } else if (CpuFamilyModel == CPUID_FULL_FAMILY_MODEL_COFFEELAKE_DT_HALO &&
       ((CpuDid == V_SA_DEVICE_ID_CFL_HALO_1) ||
        (CpuDid == V_SA_DEVICE_ID_CFL_HALO_2) ||
        (CpuDid == V_SA_DEVICE_ID_CFL_HALO_3) ||
        (CpuDid == V_SA_DEVICE_ID_CFL_HALO_IOT_1))) {
       SetPlatformId (PLATFORM_ID_CFL_H);  // CFL-H
```
myguitar: Why do we need duplicated one? We have PlatformId, BomId(cpu id?), PchId(LP or H).
FYI, all platforms are using the same CFL cpu.

mauricema: I think the problem here is that we used (CPU ID + PCH ID) combination as PlatformId.  However, if different boards have same CPU and PCH, then current code will detect them as the same PlatformId.   I believe we  might want a mechanism to address this.

PR #244, opened 2019-07-31
Add initial support for Mender boot

jgutbub:
Mender is a A/B partitioning scheme
for Linux OSes. Adding initial support
for specifying the root partition label
that should be tried. Later on the root
partition label should be acquired based
on the value of the env files stored in
the EFI/FAT32 partition of the boot media.

### DIFF
--- a/PayloadPkg/OsLoader/BootParameters.c
+++ b/PayloadPkg/OsLoader/BootParameters.c
@@ -150,7 +150,7 @@ UpdateOsMemMap (
 
   //
   // Linux E820 Mmap is very similar with multiboot MMAP
-  // Here find the diffenerce to share same update logic below
+  // Here find the difference to share same update logic below
   //
   if ((LoadedImage->Flags & LOADED_IMAGE_MULTIBOOT) != 0) {
     MultiBoot = &LoadedImage->Image.MultiBoot;
@@ -203,7 +203,7 @@ DisplayInfo (
 /**
   Update OS boot parameters
 
-  This function will append reqired command line parameters,
+  This function will append required command line parameters,
   and update mem map info.
 
   @param[in]     CurrentBootOption Current boot option
--- a/Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
+++ b/Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
@@ -1,6 +1,6 @@
 ## @file
 #
-#  Copyright (c) 2018, Intel Corporation. All rights reserved.<BR>
+#  Copyright (c) 2018 - 2019, Intel Corporation. All rights reserved.<BR>
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 ##
@@ -21,7 +21,7 @@
 
     # !BSF NAME:{Boot Flags}
     # !BSF TYPE:{Combo}
-    # !BSF OPTION:{0:Normal, 1:A/B support, 2:Crash OS, 4:Trusty support, 5:Trusty and A/B support}
+    # !BSF OPTION:{0:Normal, 1:A/B support, 2:Crash OS, 4:Trusty support, 5:Trusty and A/B support, 16:Mender support}
     # !BSF HELP:{Specify boot flags (options)}
     gCfgData.BootFlags_$(1)                 |      * | 0x01 | $(3)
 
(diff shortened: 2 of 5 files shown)


PayloadPkg/OsLoader/BootParameters.c
```diff
@@ -261,6 +262,17 @@ UpdateOsParameters (
   //
   CmdFile = &LoadedImage->Image.Common.CmdFile;
 
+  if ((CurrentBootOption->BootFlags & BOOT_FLAGS_MENDER) != 0 && (CurrentBootOption->BootFlags & BOOT_FLAGS_MISC) == 0) {
+    AsciiSPrint (MenderParams, sizeof (MenderParams), " root=PARTLABEL=primary ");
+    if ((AsciiStrSize (MenderParams) + CmdFile->Size) < CMDLINE_LENGTH_MAX) {
+      AsciiStrCatS ((CHAR8 *)CmdFile->Addr, CMDLINE_LENGTH_MAX, &MenderParams[0]);
+    } else {
+      DEBUG ((DEBUG_INFO, "Unable to append Mender cmd line parameters!\n"));
```
mauricema: Do you consider this error as fatal ?   If so, should we block the boot ?

PayloadPkg/OsLoader/BootParameters.c
```diff
@@ -261,6 +262,17 @@ UpdateOsParameters (
...
+    }
+  } else if ((CurrentBootOption->BootFlags & BOOT_FLAGS_MENDER) != 0 && (CurrentBootOption->BootFlags & BOOT_FLAGS_MISC) != 0) {
+    DEBUG ((DEBUG_ERROR, "ERROR: BOOT_FLAGS_MISC & BOOT_FLAGS_MENDER cannot both be enabled!\n"));
```
mauricema: Who will set the BootFlags ?  I guess it is the config data, right ?  If so, I think this might be an un-common error.  I am not sure if it is best to use DEBUG_ERROR print or not.   Every DEBUG_ERROR print will eventually add size to the final production binary.   So I think we can only add DEBUG_ERROR print whenever it is really required.  
And again, do you consider the condition as fatal?  Should we block boot in this case ?

PayloadPkg/OsLoader/BootParameters.c
gdong1: It is not necessary to check BOOT_FLAGS_MISC.
If any of BOOT_FLAGS_MISC and BOOT_FLAGS_MENDER is set, SBL will read data to decide boot slot. the config data tool could only select one of them.

PayloadPkg/OsLoader/BootParameters.c
gdong1: Boot option config data will set these boot flag.

PayloadPkg/OsLoader/BootParameters.c
gdong1: Maybe we could put this code to SblBootParameter.c, and refer below code to make minor change.
  //
  // Add OS A/B boot info parameter
  //
  if ((BootOption->BootFlags & BOOT_FLAGS_MISC) != 0) {
    if ((BootOption->BootFlags & LOAD_IMAGE_FROM_BACKUP) == 0) {
      AsciiStrCatS (CommandLine, MaxCmdSize, " suffix=0");
    } else {
      AsciiStrCatS (CommandLine, MaxCmdSize, " suffix=1");
    }
  }

Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
```diff
@@ -21,7 +21,7 @@
 
     # !BSF NAME:{Boot Flags}
     # !BSF TYPE:{Combo}
-    # !BSF OPTION:{0:Normal, 1:A/B support, 2:Crash OS, 4:Trusty support, 5:Trusty and A/B support}
+    # !BSF OPTION:{0:Normal, 1:A/B support, 2:Crash OS, 4:Trusty support, 5:Trusty and A/B support, 16:Mender support}
     # !BSF HELP:{Specify boot flags (options)}
```
gdong1: is it a typo for "16:" , should it be "6"?

jgutbub: No, 16 is correct because the boot flags map to bit-fields based on the macros defined here:

https://github.com/slimbootloader/slimbootloader/blob/610f66c19fc420e08e0581ceb6dbb313b8bbb9ba/BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h#L22

That is why "5" is Trusty (BIT2) and A/B support (BIT0) (BIT2 + BIT0 ==> 0x4 + 0x1). Since Medner is BIT4 it should be 16 is my understanding and based on my testing.

gdong1: Got it. I had thought we have already listed 0 ~ 5 here. so it is not bit wise value.

PR #247, opened 2019-08-02
[ShellLib] Add ShellCommandRegister API

myguitar:
Currently, all shell commands are statically defined in header file.
Add shell command registration API to allow include/exclude shell
commands dynamically.
Later, some debug shell commands will be added according to build
mode or debug mask.
- TBD: Sorting shell commands by name

### DIFF
--- a/BootloaderCommonPkg/Library/ShellLib/Shell.c
+++ b/BootloaderCommonPkg/Library/ShellLib/Shell.c
@@ -1,7 +1,7 @@
 /** @file
   A minimal command-line shell.
 
-  Copyright (c) 2017 - 2018, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2019, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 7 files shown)


mauricema:
That is a good point.  For extension shell command implemented by platform,  they should use the registration as well.

BootloaderCommonPkg/Library/ShellLib/Shell.c
```diff
@@ -111,9 +111,13 @@ Shell (
   BOOLEAN Start;
   UINTN   Index, Index1;
   SHELL   Shell;
+  CONST SHELL_COMMAND **Iter;
 
+  LoadShellCommands ();
+  for (Iter = Commands; *Iter != NULL; Iter++) {
```
myguitar: Platform specific shell commands have been verified also. Currently, ShellExtensionLib is not changed. Its platform specific commands are updated via registration API before executing Shell console.
I am thinking a way to use platform shell commands with ShellLib only(without ShellExtentionLib). Let me optimize Shell related things based on this change.

mauricema: OK, let us merge this and handle the ShellExtentionLib later on.

PR #248, opened 2019-08-03
[CFL] Fix S3/S4 resume

stalamudupula:
S4 resume sets PWRBTN_EN. And since SCI_EN is set to 0,
this situation will generate spurious SMI# once GblSmi
is enabled.

Also moving RestoreS3regs before ProcessAllLocks, as SMI_LOCK
setting will prevent enabling GblSmi on S3 resume path.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
@@ -570,13 +570,18 @@ GetPlatformPowerState (
     }
   }
 
+  ///
+  /// Clear Wake Status
+  /// Also clear the PWRBTN_EN, it causes SMI# otherwise (SCI_EN is 0)
+  ///
+  IoWrite32 (ACPI_BASE_ADDRESS + R_ACPI_IO_PM1_STS, ((UINT32)~B_ACPI_IO_PM1_EN_PWRBTN_EN & R_ACPI_IO_PM1_EN_MASK) | B_ACPI_IO_PM1_STS_WAK );
+
   if ((MmioRead8 (PCH_PWRM_BASE_ADDRESS + R_PMC_PWRM_GEN_PMCON_B) & B_PMC_PWRM_GEN_PMCON_B_RTC_PWR_STS) != 0) {
     BootMode = BOOT_WITH_FULL_CONFIGURATION;
 
     ///
-    /// Clear Wake Status and Sleep Type
+    /// Clear Sleep Type
     ///
-    IoWrite16 (ACPI_BASE_ADDRESS + R_ACPI_IO_PM1_STS, B_ACPI_IO_PM1_STS_WAK);
     IoAndThenOr16 (ACPI_BASE_ADDRESS + R_ACPI_IO_PM1_CNT, (UINT16) ~B_ACPI_IO_PM1_CNT_SLP_TYP, V_ACPI_IO_PM1_CNT_S0);
   }
 
--- a/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -839,6 +839,7 @@ STATIC S3_SAVE_REG mS3SaveReg = {
   { { IO, WIDE32, { 0, 0}, (ACPI_BASE_ADDRESS + R_ACPI_IO_SMI_EN), 0x00000000 } }
 };
 
+
 /**
   Initialize Board specific things in Stage2 Phase
 
@@ -1000,7 +1001,17 @@ BoardInit (
     }
     break;
   case ReadyToBoot:
+    //
+    // Clear Smi and restore S3 regs on S3 resume
+    //
+    ClearSmi ();
+    if ((GetBootMode() == BOOT_ON_S3_RESUME) && (GetPayloadId () == UEFI_PAYLOAD_ID_SIGNATURE)) {
+      RestoreS3RegInfo (FindS3Info (S3_SAVE_REG_COMM_ID));
+    }
 
+    //
+    // Do necessary locks, and clean up before jumping tp OS
+    //
     SiliconCfgData = (SILICON_CFG_DATA *)FindConfigDataByTag (CDATA_SILICON_TAG);
     if ( (SiliconCfgData != NULL) && (SiliconCfgData->ECEnable == 1)){
       //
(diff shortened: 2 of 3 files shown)


Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -570,15 +570,12 @@ GetPlatformPowerState (
...
-    ///
-    IoWrite16 (ACPI_BASE_ADDRESS + R_ACPI_IO_PM1_STS, B_ACPI_IO_PM1_STS_WAK);
-    IoAndThenOr16 (ACPI_BASE_ADDRESS + R_ACPI_IO_PM1_CNT, (UINT16) ~B_ACPI_IO_PM1_CNT_SLP_TYP, V_ACPI_IO_PM1_CNT_S0);
-  }
+  ///
+  /// Clear Wake Status and Sleep Type
+  /// Also clear the PWRBTN_EN, it causes SMI# otherwise
+  ///
+  IoWrite32 (ACPI_BASE_ADDRESS + R_ACPI_IO_PM1_STS, ((UINT32)~B_ACPI_IO_PM1_EN_PWRBTN_EN & R_ACPI_IO_PM1_EN_MASK) | B_ACPI_IO_PM1_STS_WAK );
+  IoAndThenOr16 (ACPI_BASE_ADDRESS + R_ACPI_IO_PM1_CNT, (UINT16) ~B_ACPI_IO_PM1_CNT_SLP_TYP, V_ACPI_IO_PM1_CNT_S0);
```
mauricema: Why do we clear the sleep type unconditionally?   And why not check 
B_PMC_PWRM_GEN_PMCON_B_RTC_PWR_STS any more ?

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -839,6 +839,71 @@ STATIC S3_SAVE_REG mS3SaveReg = {
...
+  UINTN             SpiBaseAddress;
+  UINT16            TcoBase;
+  UINTN             SpiBar0;
+
+  SiliconCfgData = (SILICON_CFG_DATA *)FindConfigDataByTag (CDATA_SILICON_TAG);
+  if ( (SiliconCfgData != NULL) && (SiliconCfgData->ECEnable == 1)){
+    //
+    // Enable decoding of I/O locations 62h and 66h to LPC
+    //
+    LpcBase = MmPciBase (0, PCI_DEVICE_NUMBER_PCH_LPC, 0);
```
mauricema: these are not locks.   Maybe we should re-arrange the code.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -839,6 +839,71 @@ STATIC S3_SAVE_REG mS3SaveReg = {
...
+    MmioOr16 (SpiBar0 + R_SPI_MEM_HSFSC, (UINT16)B_SPI_MEM_HSFSC_FLOCKDN);
+  }
+
+  //
+  // set SMI LOCK (SMI_LOCK)
+  // B0:D31:F02  Offset 1024h Bit 4
+  //
+  MmioOr8 (PCH_PWRM_BASE_ADDRESS + R_PMC_PWRM_GEN_PMCON_B, (UINT8)B_PMC_PWRM_GEN_PMCON_B_SMI_LOCK);
+
+  ClearFspHob ();
```
mauricema: this is not lock.

Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
stalamudupula: SInce we captured the boot mode already, we can clear the wake status & sleep type here. RTC_PWR_STS check  was meant for only if there is an issue with RTC battery. We can add that check after this code and set to Full configuration BootMode. But clearing SleepType and WakeStatus can be done anyways, right?

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
stalamudupula: We can just re-name the function probably? Or I can move the code back to ReadyToBoot phase instead of having a function.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1000,66 +1062,13 @@ BoardInit (
...
-    TcoBase = MmioRead16(PCH_PCR_ADDRESS(PID_DMI, R_PCH_DMI_PCR_TCOBASE)) & B_PCH_DMI_PCR_TCOBASE_TCOBA;
-    IoOr16 ((TcoBase + R_TCO_IO_TCO1_CNT), B_TCO_IO_TCO1_CNT_LOCK);
-
-    break;
-  case EndOfFirmware:
     ClearSmi ();
     if ((GetBootMode() == BOOT_ON_S3_RESUME) && (GetPayloadId () == UEFI_PAYLOAD_ID_SIGNATURE)) {
       RestoreS3RegInfo (FindS3Info (S3_SAVE_REG_COMM_ID));
     }
+    ProcessAllLocks ();
```
gdong1: If flash is locked, variable could not work in Osloader.
Maybe rename it to ProgramSecuritySetting(). and  do it here or EndOfFirmware based on PlatformId.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -839,6 +839,71 @@ STATIC S3_SAVE_REG mS3SaveReg = {
...
+  //
+
+  // Set the BILD
+  MmioOr8 (SpiBaseAddress + R_SPI_CFG_BC, (UINT8)B_SPI_CFG_BC_BILD);
+
+  //
+  // set Flash Configuration Lock-Down (FLOCKDN)
+  // B0:D31:F05  Offset 04h Bit 15
+  //
+  SpiBar0 = (UINTN)PciRead32 (PCI_LIB_ADDRESS (0, PCI_DEVICE_NUMBER_PCH_SPI, PCI_FUNCTION_NUMBER_PCH_SPI, R_SPI_CFG_BAR0));
```
gdong1: SpiBaseAddress  was already got in line 869, don't need use SPI flash BDF again here.

Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
mauricema: SleepStatus can be cleared,  however, SleepType should not be cleared.  I remember OS also needs to use this information.  
The original logic seems to be reasonable for me.   If you just need to clear PWRBTN_EN, you should just do that without changing other logic.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: Either way works for me.

PR #251, opened 2019-08-06
Break down boot option CFGDATA

mauricema:
The current implementation used a huge array to represent boot options in
a single CFGDATA tag. With this approach, when a different board needs
to modify a single field, the whole array needs to be duplicated in the
CFGDATA. It is not efficient. A better approach is to break the options
into individual boot option CFGDATA tag. It will reduce the overall CFGDATA
in general. This patch implemented this mechanism. A library interface
FillBootOptionListFromCfgData() is provided to convert the CFGDATA into
the required OS_BOOT_OPTION_LIST.


Platform/ApollolakeBoardPkg/CfgData/CfgData_BootOption.dsc
```diff
@@ -6,34 +6,18 @@
...
+  # !BSF SUBT:{BOOT_OPTION_TMPL:0 :  0    :  0 :   2   :  0   :   0  :    0 :    0 :'iasimage.bin'}
   # eMMC boot P2
-  # !BSF SUBT:{OS_TMPL:1 :  0    :  0 :   2   :  0   :   0  :    0 :    3 :      ''       :       1 :      0 :     0         :     0   :  0     :     0         :     0   :   0    }
+  # !BSF SUBT:{BOOT_OPTION_TMPL:1 :  0    :  0 :   2   :  0   :   0  :    0 :    3 :      ''      }
   # SATA boot
-  # !BSF SUBT:{OS_TMPL:2 :  0    :  0 :   0   :  0   :   0  :    0 :    0 :'iasimage.bin' :       0 :      0 :     0         :     0   :  0     :     0         :     0   :   0    }
+  # !BSF SUBT:{BOOT_OPTION_TMPL:2 :  0    :  0 :   0   :  0   :   0  :    0 :    0 :'iasimage.bin'}
   # USB boot
-  # !BSF SUBT:{OS_TMPL:3 :  0    :  0 :   5   :  0   :   0  :    0 :    0 :'iasimage.bin' :       0 :      0 :     0         :     0   :  0     :     0         :     8   :   0    }
+  # !BSF SUBT:{BOOT_OPTION_TMPL:3 :  0    :  0 :   5   :  0   :   0  :    0 :    0 :'iasimage.bin'}
```
gdong1: In order to let users to use the new config data structure smoothly, how about providing an example that demos how to support Android A/B with trusty support?

Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
```diff
@@ -6,17 +6,22 @@
...
 
+    # !BSF SUBT:{CFGHDR_TMPL:BOOT_OPTION_CFG_DATA_$(1):0:0}
+    # !HDR EMBED:{BOOT_OPTION_CFG_DATA_$(1):TAG_05$(1):START}
+
     # !BSF NAME:{Image Type}
     # !BSF TYPE:{Combo}
-    # !BSF OPTION:{0:Default, 1:Android, 2:ClearLinux, 3:Acrn, 4:Fastboot}
-    # !BSF HELP:{Specify boot image type}
+    # !BSF OPTION:{0:Default, 1:Android, 2:ClearLinux, 3:Acrn, 4:Fastboot, 0xFE:Addendum, 0xFF:Not used}
+    # !BSF HELP:{Specify boot image type.}
```
gdong1: wondering what user should set if they want to create a boot option for Android with A/B support. It is Android image, and also need addendum to add boot option information for misc partition.

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -0,0 +1,87 @@
...
+      UpdateFlag = 1;
+      ImageIdx   = 0;
+      CopyMem (BootOption, BootOptionCfgData, OFFSET_OF (OS_BOOT_OPTION, Image[0]));
+    }
+
+    if (UpdateFlag > 0) {
+      StrPtr = (CHAR8 *)BootOptionCfgData->Image[0].FileName;
+      // Use either LBA or filename. '#' indicates it is LBA string.
+      if ((StrPtr[0] == '#') && (AsciiStrHexToUint64S (StrPtr + 1, NULL, &Lba) == RETURN_SUCCESS)) {
+        BootOption->Image[ImageIdx].LbaImage.LbaAddr = Lba;
```
gdong1: Looks in the "addendum" boot option also has common header (e.g. devtype, hardware partition, etc). 
1) Do we want to have it or just have the information need in Image[0] to save size further?
2) Here it only copies LBA or filename here. where we copy other information, e.g. software partition?

mauricema: Yes, it has common header,  but not everything is used, such as devtype and hardware partition.  Since it is same structure as a normal entry, it has the same size.  Plus, the swpart will be used.  

Good catch, the swpart assignment was missing.  Will add it.

Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
```diff
@@ -6,17 +6,22 @@
 ##
 
   #
-  # ========================= OS_TMPL ==========================
+  # ========================= BOOT_OPTION_TMPL ==========================
   #
   # NOTE: MUST match OS_BOOT_OPTION structure!!!
```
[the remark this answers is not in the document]
mauricema: It still needs to match OS_BOOT_OPTION for the part before BOOT_IMAGE.  Will add more comments there.

Platform/ApollolakeBoardPkg/CfgData/CfgData_BootOption.dsc
mauricema: I don't have example for now.  Maybe we can add this when a real platform needs to support this feature.

Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
mauricema: It should be marked as Android image, and then in the Flags, it should mark A/B support.  
An addendum entry will be appended to provide misc partition info.

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
mauricema: Updated a new patch.

Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
gdong1: Understood.  If one boot option (e.g. boot option 1) has addendum, will the addendum been shown as a separate boot option (e.g. boot option 2) or as an addendum together with its primary boot option (e.g. boot option 1) in ConfigEditor?

mauricema: In ConfigEditor, it will be showed as a boot option, but with Image Type set to "Addendum"

gdong1: I remember previously it will shows "boot option 1", "boot option 2" for every boot option entry. 
If we still have these information there, it would be a little confusing since "boot option 2" is really boot option 2, basically it is part of boot option 1.
Not sure if we could show one full boot option in same page as before.

mauricema: It won't be able to show everything in a single page like before.  Here the entry might not equal to boot option. since multiple entries can be combined.   We can hide some items to let user know it is not a normal option.

gdong1: Yes, better hide the things not necessary for addendum.
and still prefer not show boot option 2 for the addendum for option 1.  Need update "OS:"Boot Option $(1)""

mauricema: Added code to hide some items for addendum entry.
But regarding to not show boot option 2, I don't believe  it is feasible. 
Even if you do so,  if somebody config the image type back to normal type, it will become a normal option again and should not hide.  Let us keep it that way.   And considering people will use the normal boot option entry in most of the time, I believe it is OK.

Platform/ApollolakeBoardPkg/CfgData/CfgData_BootOption.dsc
mauricema: Tested some populated boot options with Android A/B information.  It does build boot options properly.

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
mauricema: Added logic in template to hide some items if it is configured as addendum.

mauricema:
Added back the old boot option template in case that any project still use it.   But the recommendation is to use the new boot option template BOOT_OPTION_TMPL.

Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
gdong1: Understand the title is static, not support dynamic change now. so let's use it now.

PR #255, opened 2019-08-07
[CFL] Optimize Gpio Cfg Data input

stalamudupula:
PinOffset information like PadNum, GrpIdx for a GPIO can be
adjusted in the fields available in DW1, where as ChipsetId
can be determined dynamically. This reduces the config data
size significantly. So, we can remove PinOffset info i.e. the
first dword in the GpioTableData.

Moved static definitions to top of the file in Stage2 board init.


Platform/CoffeelakeBoardPkg/Library/Stage1BBoardInitLib/GpioTables.h
```diff
@@ -10,6 +10,28 @@
 #include <GpioPinsCnlH.h>
 #include <GpioPinsCnlLp.h>
 
+//
+// GPIO_PAD Fileds
+//
+typedef struct {
```
mauricema: maybe move them into header file instead

mauricema: Never mind, it is already in H file.

PR #259, opened 2019-08-09
[CFL] MRC data save is ignored after first boot

raghavag:
Current code will check if the first 4 bytes of the existing
MRC training data, if it is 0xFFFFFFFF then MRC training data
will be saved otherwise even though there is change in the
training data, it will not be saved.

Modified the code to compare the incoming training data to the
saved data, if they do not match, save the incoming training data.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -1596,54 +1596,53 @@ SaveNvsData (
   UINT32      Address;
   UINT32      BaseAddress;
   UINT32      RegionSize;
-  UINT32      TotalSize;
+  UINT32      MrcDataRegSize;
 
-  Status = GetComponentInfo (FLASH_MAP_SIG_MRCDATA, &Address, NULL);
+  Status = GetComponentInfo (FLASH_MAP_SIG_MRCDATA, &Address, &MrcDataRegSize);
   if (EFI_ERROR(Status)) {
     return EFI_NOT_FOUND;
   }
-  if (*(UINT32 *)Address == 0xFFFFFFFF) {
-    Status = SpiConstructor ();
-    if (EFI_ERROR (Status)) {
-      return Status;
-    }
 
-    Status = SpiGetRegionAddress (FlashRegionAll, NULL,  &TotalSize);
-    if (EFI_ERROR (Status)) {
-      return Status;
-    }
+  if (Length > MrcDataRegSize) {
+    return EFI_INVALID_PARAMETER;
+  }
 
-    Status = SpiGetRegionAddress (FlashRegionBios, &BaseAddress, &RegionSize);
-    if (EFI_ERROR (Status)) {
-      return Status;
-    }
+  //
+  // Compare input data against the stored MRC training data
+  // if they match, no need to update again.
+  //
+  if (CompareMem ((VOID *)Address, Buffer, Length) == 0){
+    return EFI_ALREADY_STARTED;
+  }
 
-    BaseAddress += ((UINT32)(~TotalSize) + 1);
-    if (Address < BaseAddress) {
-      return EFI_ACCESS_DENIED;
-    }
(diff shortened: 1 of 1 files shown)


Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1602,48 +1602,53 @@ SaveNvsData (
...
+  }
 
-    Status = SpiGetRegionAddress (FlashRegionBios, &BaseAddress, &RegionSize);
-    if (EFI_ERROR (Status)) {
-      return Status;
-    }
+  Status = SpiConstructor ();
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
```
gdong1: SPI is already initialized in line 874. It is not necessary to init it again.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1602,48 +1602,53 @@ SaveNvsData (
...
-    if ((Address + ROUNDED_UP(Length, KB_(4))) > RegionSize) {
-      return EFI_OUT_OF_RESOURCES;
-    }
+  Status = SpiGetRegionAddress (FlashRegionBios, &BaseAddress, &RegionSize);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
 
+  BaseAddress += ((UINT32)(~TotalSize) + 1);
+  if (Address < BaseAddress) {
```
gdong1: Using " BaseAddress = ((UINT32)(~RegionSize) + 1);", it is not necessary to get TotalSize.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1602,48 +1602,53 @@ SaveNvsData (
...
-    if (EFI_ERROR (Status)) {
-      return Status;
-    }
+  //
+  // Compare input data against the stored MRC training data
+  // if they match, no need to update again.
+  //
+  if (CompareMem ((VOID *)Address, Buffer, Length) == 0){
+    return EFI_ALREADY_STARTED;
+  }
```
gdong1: Late need optimize this check to avoid erasing and saving all these data.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1602,48 +1602,53 @@ SaveNvsData (
   if (EFI_ERROR(Status)) {
```
gdong1: Need get MRC region size and check if (NvsSize < Length)  before really saving NVS data.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Done

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Updated new patch

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Okay

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Updated new patch

PR #263, opened 2019-08-12
CFL: Add TPMType param to Loader Platform info HOB

lsubash:
Update TpmType to LoaderPlatform info hob based on
Type updated in Platform Info.

### DIFF
--- a/BootloaderCommonPkg/Include/Guid/LoaderPlatformInfoGuid.h
+++ b/BootloaderCommonPkg/Include/Guid/LoaderPlatformInfoGuid.h
@@ -45,12 +45,6 @@ extern EFI_GUID gLoaderPlatformInfoGuid;
 //Secure debug (0 - disabled, 1 - enabled)
 #define        HWSTATE_SECURE_DEBUG              BIT3
 
-//Tpm Type dTPM 2.0 (0 - Not Availble, 1 - detected)
-#define        HWSTATE_DTPM_20                   BIT4
-
-//Tpm Type TPM 2.0 PTT (0 - Not Availble, 1 - detected)
-#define        HWSTATE_TPM_PTT                   BIT5
-
 //Definition for LOADER_PLATFORM_INFO.Flags
 
 //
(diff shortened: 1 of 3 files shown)


mauricema:
Please use [CFL]  in subject to be consistent with other commit.

BootloaderCommonPkg/Include/Guid/LoaderPlatformInfoGuid.h
```diff
@@ -45,12 +45,6 @@ extern EFI_GUID gLoaderPlatformInfoGuid;
 //Secure debug (0 - disabled, 1 - enabled)
 #define        HWSTATE_SECURE_DEBUG              BIT3
 
-//Tpm Type dTPM 2.0 (0 - Not Availble, 1 - detected)
-#define        HWSTATE_DTPM_20                   BIT4
-
-//Tpm Type TPM 2.0 PTT (0 - Not Availble, 1 - detected)
-#define        HWSTATE_TPM_PTT                   BIT5
-
```
gdong1: Better still use HWSTATE instead adding a new field? we don't need change code, e.g.
#define  HWSTATE_TPM_MASK          (BIT4 | BIT5)
#define  HWSTATE_TPM_SHIFT           4
#define  TPM_TYPE_NOT_KNOWN     (0x0 << HWSTATE_TPM_SHIFT)
#define  TPM_TYPE_NONE                 (0x1 << HWSTATE_TPM_SHIFT)
...

mauricema: I would think it makes sense to use separate new field in this case.

lsubash: HwState is maintained bit wise for each feature. As TpmType is additional info, it think it would be good to have a separate field.

gdong1: If we add new field, do need update structure version? and consumer need check the version to know if new field is there before consuming TPM information.

lsubash: I would update the structure version to "2" and check before we consume it.

gdong1: It is ok for me if update version and check it.

PR #265, opened 2019-08-13
[Tools] Add !include support for DLT files

mauricema:
To make it easier to share the most common configuration items in a
DLT file, it is better to support '!include' in DLT file. This patch
added this feature.  It also added examples in QEMU to test this.


sindhuragrandhi:
Great, thanks so much for adding this support!

PR #266, opened 2019-08-14
Add common functions on PlatformName and external config data

gdong1:
Getting platform name from config data and set platform name is
common almost for all the platforms. So move it to common library.
Loading external configuration data from SPI is also common, so
add it into common library. Platform could still use its own method
to load config data if the common one doesn't work.

### DIFF
--- a/Platform/CommonBoardPkg/Include/Library/BoardSupportLib.h
+++ b/Platform/CommonBoardPkg/Include/Library/BoardSupportLib.h
@@ -21,4 +21,38 @@ FillBootOptionListFromCfgData (
   IN OUT   OS_BOOT_OPTION_LIST   *OsBootOptionList
 );
 
+
+/**
+  Set the platform name with CFGDATA info
+
+ **/
+VOID
+EFIAPI
+PlatformNameInit (
+  VOID
+);
+
+/**
+  Load the configuration data blob from SPI flash into destination buffer.
+  It supports the sources: PDR, BIOS for external Cfgdata.
+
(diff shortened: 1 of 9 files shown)


Platform/CommonBoardPkg/Include/Library/BoardSupportLib.h
```diff
@@ -21,4 +21,38 @@ FillBootOptionListFromCfgData (
...
+  @retval  EFI_SUCCESS             Configuration data blob was loaded successfully.
+  @retval  EFI_NOT_FOUND           Configuration data blob cannot be found.
+  @retval  EFI_OUT_OF_RESOURCES    Destination buffer is too small to hold the
+                                   configuration data blob.
+  @retval  Others                  Failed to load configuration data blob.
+
+**/
+EFI_STATUS
+EFIAPI
+SpiLoadExternalConfigData (
```
mauricema: Why do we call SpiLoadExternalConfigData ?  Do we intend to load from somewhere else ?

gdong1: yes. External config data could come from other places, e.g. UFS or NVMe. The platform code could decide where to load it, and potentially platform code could have its own impl (e.g. we don't need move LoadExternalConfigData() from ConfigDataLib to BoardSupportLib, and it is a compatible change).

mauricema: ok

PR #268, opened 2019-08-15
Switch to use container for pre-OS checker/payload

jgutbub:
Since we may want to perform FW update on
pre-OS checker/payload binaries separately
from the OS Loader payload we will search
for pre-OS checker/payload in the container
entries instead of adding it into the OS
Loader FD.

### DIFF
--- a/PayloadPkg/OsLoader/OsLoader.c
+++ b/PayloadPkg/OsLoader/OsLoader.c
@@ -265,8 +265,7 @@ SetupBootImage (
   be loaded as part of the payload, if found get the entry point for
   execution later instead of jumping into the OS directly.
 
-  @param[in] PldBase   Payload Image base to search through for
-                       the pre-OS checker binary.
+  @param[in] PreOsCheckerImageBase   Base of the pre-OS checker loaded in memory.
 
   @retval  NULL        Pre-OS checker is not loaded successfully.
   @retval  Others      Pre-OS checker is loaded successfully and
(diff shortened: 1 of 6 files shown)


PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -776,11 +758,13 @@ PayloadMain (
 #endif
 
   // Check if there is a pre-OS checker that needs to be executed
-  if (PldBase != NULL) {
-    mPreOsCheckerEntry = LoadPreOsChecker (PldBase);
+  Status = LocateComponentEntry (SIGNATURE_32 ('I', 'P', 'F', 'W'), SIGNATURE_32 ('P', 'O', 'S', 'C'), &ContainerEntry, &ComponentEntry);
+  if (!EFI_ERROR (Status)) {
+    mPreOsCheckerEntry = LoadPreOsChecker (
+      ContainerEntry->Base + ((CONTAINER_HDR *)ContainerEntry->HeaderCache)->DataOffset + ComponentEntry->Offset + sizeof (LOADER_COMPRESSED_HEADER)
```
jgutbub: I am not sure this will be the best way to get the data of the 'POSC' binary data. When I try to something like was done here https://github.com/slimbootloader/slimbootloader/blob/master/BootloaderCorePkg/Stage2/Stage2.c#L188 it does not seem to be working for me in Os Loader.

mauricema: It is not the preferred way to load a component since it does not handle the verification and decompression.  It should just work with LoadCompoent() in OsLoader.   If not, we should treat it as a bug and need to be fixed.

jgutbub: Retested using the LoadComponent( )routine and it is working, not sure what the issue was when testing previously.

PR #270, opened 2019-08-15
Fix FSP and Microcode download path issue

### LINKED ISSUE #269: Build will fail when board package and silicon package is located outside of the SBL
Per design, SBL allows to set PLT_SOURCE environment variable to search supported boards and silicons from different location.  When we tried this way, the build failed because the FSP and Microcode downloading still points to the boards and silicon inside SBL rather than the new location.    

Below is the failure trace
Traceback (most recent call last):
  File "W:\SlimBoot\Test\SblOpen\BootloaderCorePkg\Tools\PrepareBuildComponentBin.py", line 334, in <module>
    sys.exit(Main())
  File "W:\SlimBoot\Test\SblOpen\BootloaderCorePkg\Tools\PrepareBuildComponentBin.py", line 328, in Main
    CopyFspBins (fsp_repo_dir, plt_dir, platform)
  File "W:\SlimBoot\Test\SblOpen\BootloaderCorePkg\Tools\PrepareBuildComponentBin.py", line 118, in CopyFspBins
    copy_list, fsp_tag = GetFspCopyList (platform)
  File "W:\SlimBoot\Test\SblOpen\BootloaderCorePkg\Tools\PrepareBuildComponentBin.py", line 74, in GetFspCopyList
    fd = open (fsp_inf, 'r')
IOError: [Errno 2] No such file or directory: 'Silicon/CoffeelakePkg/FspBin/FspBin.inf'
Traceback (most recent call last):
  File "W:\SlimBoot\Test\SblOpen\BuildLoader.py", line 1238, in <module>
    main()
  File "W:\SlimBoot\Test\SblOpen\BuildLoader.py", line 1235, in main
    args.func(args)
  File "W:\SlimBoot\Test\SblOpen\BuildLoader.py", line 1183, in cmd_build
    Build(board).build()
  File "W:\SlimBoot\Test\SblOpen\BuildLoader.py", line 1033, in build
    self.pre_build()
  File "W:\SlimBoot\Test\SblOpen\BuildLoader.py", line 917, in pre_build
    raise Exception  ('Failed to prepare build component binaries !')
Exception: Failed to prepare build component binaries !
Traceback (most recent call last):
  File "W:\SlimBoot\Test\SblBuild.py", line 42, in <module>
    sys.exit(main())
  File "W:\SlimBoot\Test\SblBuild.py", line 39, in main
    raise RuntimeError("Execution failed: %s " % (build_cmd))
RuntimeError: Execution failed: ['python', 'W:\\SlimBoot\\Test\\SblOpen\\BuildLoader.py', 'build', 'cfl']

mauricema:
When downloading FSP and Microcode into local tree, it is required
to support board/silicon packages at different locations since the
board/silicon packages might be located outside of the SBL tree pointed
by PLT_SOURCE environment variable. Current code always assumes the
board/silicon is inside SBL tree, and it needs to be fixed. This
patch fixed #269 .

### DIFF
--- a/BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
+++ b/BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
@@ -69,7 +69,7 @@ def CopyFileList (copy_list, src_dir, sbl_dir):
 
 def GetFspCopyList (platform):
     plat_pkg = GetPlatformPkgName (platform)
-    fsp_inf  = 'Silicon/%s/FspBin/FspBin.inf' % plat_pkg
+    fsp_inf  = os.path.join(os.environ['PLT_SOURCE'], 'Silicon/%s/FspBin/FspBin.inf' % plat_pkg)
 
     fd = open (fsp_inf, 'r')
     lines = fd.readlines()
@@ -246,7 +246,7 @@ def CloneUcodeRepo (ucode_dir, tag):
 
 def GetUcodeList (platform):
     plat_pkg  = GetPlatformPkgName (platform)
-    ucode_inf = 'Silicon/%s/Microcode/Microcode.inf' % plat_pkg
+    ucode_inf = os.path.join(os.environ['PLT_SOURCE'], 'Silicon/%s/Microcode/Microcode.inf' % plat_pkg)
 
     fd = open (ucode_inf, 'r')
     lines = fd.readlines()


BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -69,7 +69,7 @@ def CopyFileList (copy_list, src_dir, sbl_dir):
 
 def GetFspCopyList (platform):
     plat_pkg = GetPlatformPkgName (platform)
-    fsp_inf  = 'Silicon/%s/FspBin/FspBin.inf' % plat_pkg
+    fsp_inf  = os.path.join(os.environ['PLT_SOURCE'], 'Silicon/%s/FspBin/FspBin.inf' % plat_pkg)
```
gdong1: This could not fix the build failure issue when moving silicon package out of SblOpen since sbl_dir is hard code. 
and in this file, there is other hard code to CFL. 
It would be better to make it a generic script, so we don't need update this script when there is a new platform open sourced.

mauricema: I did test to move the CFL board/silicon to outside of SBL tree and it did work as expected.
However, if you are using different target name other than original supported 'apl', 'cfl', then it will cause issue.   Adding support for generic target name is another scope which we can address in different patch.

gdong1: OK. let's check it in firstly. Address different file name in another patch.

PR #272, opened 2019-08-15
Enhance build support for binary downloading

gdong1:
Removing hard code in PrepareBuildComponentBin.py, so it could
support other platforms. And enhance its logic to support to
run in different places.
Moving repo and commit information to driver INF so this script
could reuse code to support different drivers.

### DIFF
--- a/BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
+++ b/BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
@@ -16,34 +16,37 @@ def Fatal (msg):
     sys.stdout.flush()
     raise Exception (msg)
 
+def CloneRepo (clone_dir, driver_inf):
+    repo, commit = GetRepoAndCommit (driver_inf)
+    if repo == '' or commit == '':
+        Fatal ('Failed to find repo and commit information!')
 
-def CloneFspRepo (fsp_dir, fsp_tag):
-    if not os.path.exists(fsp_dir + '/.git'):
-        print ('Cloning Intel FSP repo ...')
-        cmd = 'git clone https://github.com/IntelFsp/FSP.git %s' % fsp_dir
+    if not os.path.exists(clone_dir + '/.git'):
+        print ('Cloning the repo ... %s' % repo)
+        cmd = 'git clone %s %s' % (repo, clone_dir)
         ret = subprocess.call(cmd.split(' '))
         if ret:
-            Fatal ('Failed to clone FSP repo to directory %s !' % fsp_dir)
+            Fatal ('Failed to clone repo to directory %s !' % clone_dir)
         print ('Done\n')
     else:
-        print ('Update Intel FSP repo ...')
+        print ('Update the repo ...')
         cmd = 'git fetch origin master'
-        ret = subprocess.call(cmd.split(' '), cwd=fsp_dir)
+        ret = subprocess.call(cmd.split(' '), cwd=clone_dir)
         if ret:
-            Fatal ('Failed to update FSP repo to directory %s !' % fsp_dir)
+            Fatal ('Failed to update repo in directory %s !' % clone_dir)
         print ('Done\n')
 
-    print ('Checking out Intel FSP commit ...')
+    print ('Checking out specified version ... %s' % commit)
 
-    cmd = 'git checkout %s' % fsp_tag
-    ret = subprocess.call(cmd.split(' '), cwd=fsp_dir)
+    cmd = 'git checkout %s' % commit
+    ret = subprocess.call(cmd.split(' '), cwd=clone_dir)
     if ret:
-        Fatal ('Failed to check out branch !')
+        Fatal ('Failed to check out specified version !')
     print ('Done\n')
 
 
 def CheckFileListExist (copy_list, sbl_dir):
-    exists = True
+    exists  = True
     for src_path, dst_path in copy_list:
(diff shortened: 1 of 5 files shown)


BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -212,122 +225,32 @@ def BuildFspBins (fsp_dir, sbl_dir, platform, flag):
...
-    CopyFileList (copy_list, ucode_dir, sbl_dir)
-
-    print ('Done\n')
+    CopyFileList (copy_list, fsp_dir)
 
 def Main():
-    curr_dir       = os.path.dirname (os.path.realpath(__file__))
-    workspace_dir  = os.path.abspath (os.path.join(curr_dir, '../../..'))
+    sbl_dir        = os.environ['PLT_SOURCE']
+    workspace_dir  = os.path.join(sbl_dir, 'download')
```
mauricema: let us not change the behavior here.  All downloaded repo should be located outside of current SBL tree.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -180,7 +193,7 @@ def BuildFspBins (fsp_dir, sbl_dir, platform, flag):
     print ('Done\n')
 
     print ('Applying QEMU FSP patch ...')
-    patch_dir = os.path.join(sbl_dir, 'Silicon/QemuSocPkg/FspBin/Patches')
+    patch_dir = os.path.join(os.environ['PLT_SOURCE'], 'Silicon/QemuSocPkg/FspBin/Patches')
```
mauricema: Does this work ?   If we set PLT_SOURCE to a valid path, you won't be able to locate QEMU patches since it is always in the SblOpen.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -52,8 +56,9 @@ def CheckFileListExist (copy_list, sbl_dir):
     return exists
 
 
-def CopyFileList (copy_list, src_dir, sbl_dir):
+def CopyFileList (copy_list, src_dir):
```
mauricema: let us keep this function more generic , and do not depends on env variables.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -16,34 +16,38 @@ def Fatal (msg):
...
+    ret = subprocess.call(cmd.split(' '), cwd=clone_dir)
     if ret:
-        Fatal ('Failed to check out branch !')
+        Fatal ('Failed to check out specified version !')
     print ('Done\n')
 
 
-def CheckFileListExist (copy_list, sbl_dir):
-    exists = True
+def CheckFileListExist (copy_list):
```
mauricema: let us keep interface more generic , do not depends on env variables.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -16,34 +16,38 @@ def Fatal (msg):
...
+    exists  = True
+    sbl_dir = os.environ['PLT_SOURCE']
```
mauricema: why do we check it in os.environ['PLT_SOURCE'] folder ?

mauricema: for low level function, it should be generic so that it can be reused to check files in other folders as well.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
gdong1: Will move to upper level for SblOpen. the idea is to have single download folder to save different downloaded repos.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
gdong1: it works if build QEMU @ SblOpen folder using buildLoader.py.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
gdong1: OK. will add it back.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
gdong1: OK. will add it back.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
gdong1: OK. will add sbl_dir back.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
mauricema: ok

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
mauricema: BTW, can we use "Download" to be consistent with the naming convention ?

gdong1: Done

PR #286, opened 2019-08-23
Add shell command to trigger firmware update

raghavag:
This patch will add shell command to trigger firmware
update on CFL/WHL platforms.


mauricema:
As far as I know the triggering FWU is platform specific.

PR #292, opened 2019-08-26
[QEMU] Support -smp option

myguitar:
QEMU supports '-smp' option to emulate up-to maximum 255 CPUs, but failed
to do MpInit with '-smp' option. Currently, there are only fixed size(4)
Processor Local APIC entries in QEMU MADT and its ApicId does not match
with detected CPU info.
- Append ProcessorLocalApic entries into MADT in runtime with MpInfo
- Make PcdCpuMaxLogicalProcessorNumber configurable in each BoardConfig
- Set PcdCpuMaxLogicalProcessorNumber to 255 in QEMU BoardConfig.py

Test>
qemu-system-x86_64 -machine q35 -nographic -serial mon:stdio
                   -pflash Outputs/qemu/SlimBootloader.bin
                   -smp 8

Signed-off-by: Aiden Park aiden.park@intel.com

### DIFF
--- a/BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
+++ b/BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
@@ -42,6 +42,11 @@ extern  UINT32          WakeUpBuffer;
 extern  CHAR8           WakeUp;
 extern  UINT32          WakeUpSize;
 
+typedef struct {
+  UINT8   Type;
+  UINT8   Length;
+} EFI_ACPI_MADT_ENTRY_COMMON_HEADER;
+
 /**
   Update Firmware Performance Data Table (FPDT).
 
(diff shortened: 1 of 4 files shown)


BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
```diff
@@ -345,52 +350,76 @@ UpdateMadt (
...
+  //
+  Length = sizeof (EFI_ACPI_5_0_MULTIPLE_APIC_DESCRIPTION_TABLE_HEADER);
+  CopyMem ((VOID *)Current, (VOID *)MadtPtr, Length);
+  MadtPtr += Length;
+  Current += Length;
+
+  //
+  // Append Processor Local APIC info
+  //
+  LocalApic = (EFI_ACPI_5_0_PROCESSOR_LOCAL_APIC_STRUCTURE *)Current;
```
mauricema: Is there any specific reason we want to put LOCAL APIC entries at the top ?
I was thinking it might be easier to copy the original MADT table and then append local APIC entries, then adjust the final length.

myguitar:
PR updated. This has been verified on QEMU, APL and CFL boards.

BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
myguitar: > Is there any specific reason we want to put LOCAL APIC entries at the top ?
> I was thinking it might be easier to copy the original MADT table and then append local APIC entries, then adjust the final length.

The original MADT has already fixed size Local APIC entries, so this hook ignores existing one. Basically, MADT entries are sorted by its TYPE, and Local APIC type is first item as Type 0. It should be okay to append it to to top or bottom since it's enumerated by Type, but this is just to keep sorted order.

mauricema: Do you think we can just remove the local APIC entries from the original MADT table to make it easier ?

mauricema: maybe save a little bit size, although not too much!  :)

myguitar: Right, if we remove the Local APIC entries from Madt.aslc, we can save the AcpiTable size and this hook would be simpler to append the entries. But, I think better to keep this code, which skips existing Local APIC entries and appends new one, to avoid an issue when Madt.aslc has any Local APIC entry.

mauricema: OK,  I am fine with it.

mauricema: OK,  I am fine with it.

PR #295, opened 2019-08-27
Resolve Klocwork issues in Payload & Platform

jgutbub:
Klocwork scanning reported several issues in the
PayloadPkg and Platform code folders, this commit
aims to resolve all of the issues currently being
reported in these folders.


jgutbub:
> > Not sure there is a way to run the Klocwork scan on the changed code locally to confirm the patch resolves all of the issues before being merged to avoid multiple commits.
> 
> Yes, if you get access you could trigger a build from branch.

I think the KW scan is running on the submodule code instead of the SblOpen branch that I provided. I will merge this patch and see if the # of KW issues decreases.

PR #296, opened 2019-08-27
To discuss the capability of adding build data in config editor

joshloo:
A DSC file is added to configure build related data. Currently, the build script consumes parameters configured in boardconfig.py.
This patch enables another mechanism to configure build parameter in config editor without removing the existing method. If there's a duplicate, the data written in config editor will overwrite the data in boardconfig.py.

The flow goes as below:
1) User runs ConfigEditor and load BuildConfigDef.dsc
2) User configures BuildData section and click "Export Build Data" option in menu
3) BuildData.py is generated in Platform\*BoardPkg\BoardData folder
4) The BoardConfig.py will check if BuildData.py exists. If it does, boardconfig.py will overwrite with the options configured there.


joshloo:
Let me resolve the conflicts first before the submitting the review

PR #298, opened 2019-08-28
Add capability to configure build data via ConfigEditor

joshloo:
A DSC file is added to configure build related data. Currently,
the build script consumes parameters configured in boardconfig.py.
This patch enables another mechanism to configure build parameter
in config editor without removing the existing method. If there's
a duplicate, the data written in config editor will overwrite the
data in boardconfig.py.

The flow goes as below:
1) User runs ConfigEditor and load BuildConfigDef.dsc
2) User configures BuildData section and click "Export Build Data"
   option in menu
3) BuildData.py is generated in Platform\*BoardPkg\BoardData folder
4) The BoardConfig.py will check if BuildData.py exists. If it does,
   boardconfig.py will overwrite with the options configured there.

Signed-off-by: joshloo <joshlootunglun@gmail.com>
Signed-off-by: Loo Tung Lun <tung.lun.loo@intel.com>


myguitar:
Agree with @gdong1 . ConfigEditor is not to give build options, but to generate config data which gives different configuration values or control different behavior without any code changes. I think considering a way like Linux Kconfig would be better option to give build options.

mauricema:
Linux Kconfig is a well known way to do build config.   If we just look from input/output perspective,  it is very similar to ConfigEditor too.  The question is :  do we want to leverage what we have, or  want to move to Kconfig based approach here ?

gdong1:
From my understanding, adding GUI configuration support to BoardConfig.py will increase developer effort to provide details help information and test it. If we could not provide accurate information, it would be useless and confuse user. Comparing it with configuration data, configuration data is high priority and I prefer we need take more time on it.

mauricema:
Yes, agree CFGDATA has higher priority.  From a long run, if we expect people to modify these options in BoardConfig.py, we will have to provide detailed help info (today it is missing, some of them have comments).

gdong1:
Currently we could add more comments if some config items are difficult to understand.  I am thinking it might be not necessary to let tool to config it. Replacing BoardConfig.py to be a TXT file (e.g. BoardConfig.dsc) is also an option if user doesn't understand python. It would help us if we could get some feedback from user/customer.

mauricema:
Python has its own flexibility.  I don't believe it can be easily replaced by a TXT or DSC file today.  
Sure, we can get more opinions from the users.

mauricema:
There are lots of discussions on how it should be handled.  At this moment, most of the developers prefer maintaining the python file directly instead of introducing new config mechanism.   As part of it, this PR will be closed.  If anything changes later on, we can still revisit it.

PR #299, opened 2019-08-28
Enable optional extra platform DSC

jgutbub:
For additional DSC file settings that are
platform specific we may want to be able
to add an extra DSC file that can use the
same $(VAR) logic that BootloaderCore DSC
file can use based on BoardConfig.py vars.


jgutbub:
Need to consider alternative solutions first.

PR #303, opened 2019-08-29
Simplify xml patching in stitchifwi

raghavag:
This patch will use a list of xml changes required and
patch them in a loop rather than patching each change
seperately.

This patch also add parameter to indicate platform for which
ifwi is getting build.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
@@ -75,9 +75,9 @@ Please follow the steps below for WHL/CFL IFWI stitching.
   4. Stitch the final image
      Run this script from Slim Bootloader root directory:
      EX:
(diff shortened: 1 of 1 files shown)


Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -293,161 +381,11 @@ def gen_xml_file(stitch_dir, cfg_var, btg_profile, spi_quad):
 
     tree = ET.parse(os.path.join(stitch_dir, cfg_var['fitinput'], 'new.xml'))
 
-    #Region Order
```
gdong1: Platform is added gen_xml_file(), but it is not used when generate xml file.
Different SKU will have different xml file. So should we pass SKU/platform info to fit command in line 379?

raghavag: True, Fixed this.

PR #305, opened 2019-08-30
Add ACM3 binary into BIOS region

sindhuragrandhi:
Signed-off-by: Grandhi <sindhura.grandhi@intel.com>

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -192,6 +192,7 @@ class BaseBoard(object):
 		self.ENABLE_LINUX_PAYLOAD  = 0
 
 		self.ACM_SIZE              = 0
+		self.ACM3_SIZE             = 0
 		self.UCODE_SIZE            = 0
 		self.CFGDATA_SIZE          = 0
 		self.MRCDATA_SIZE          = 0
@@ -339,12 +340,18 @@ class Build(object):
 			if len(u_code_images) > 0:
 				raise Exception('  Insufficient uCode entries in FIT. Need %d more.' % len(u_code_images))
 
-		if self._board.ACM_SIZE > 0:
 			# ACM
+		if self._board.ACM_SIZE > 0:
 			fit_entry = FitEntry.from_buffer(rom, fit_offset + (num_fit_entries+1)*16)
 			fit_entry.set_values(self._board.ACM_BASE, 0, 0x100, 0x2, 0)
 			print ('  Patching entry %d with 0x%08X:0x%08X - ACM' % (num_fit_entries, fit_entry.address, fit_entry.size))
 			num_fit_entries     += 1
+
+			# ACM3 Fit entry should be in sequential order with/without BTG enabled
+			# Save the next FIT entry for ACM3 here and set it later below
+			if self._board.ACM3_SIZE > 0:
+				acm3_index       = num_fit_entries
+				num_fit_entries     += 1
 
 			# BIOS Module (IBB segment 0): from FIT table end to 4GB
 			# Record it now and update later since the FIT size is unknown yet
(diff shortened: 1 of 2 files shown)


sindhuragrandhi:
This patch needs a matching patch from SblOpen (currently under review). Once the Sblopen patch is merged, this patch should be able to build successfully without issues.

sindhuragrandhi:
SblOpen ACM patch:
https://gitlab.devtools.intel.com/OWR/IoTG/BDP/Bootloader/SlimBootloader/sblplatform/merge_requests/77

BuildLoader.py
```diff
@@ -339,13 +340,24 @@ def update_fit_table (self):
 			if len(u_code_images) > 0:
 				raise Exception('  Insufficient uCode entries in FIT. Need %d more.' % len(u_code_images))
 
-		if self._board.ACM_SIZE > 0:
+			acm3_index = 0
+			if self._board.ACM_SIZE == 0:
+				acm3_index       = num_fit_entries
+				num_fit_entries     += 1
+
 			# ACM
```
gdong1: 1) ACM3 is not related with UCODE, should these code be moved out of "if self._board.UCODE_SIZE > 0:"? 
2) why num_fit_entries + 1 if self._board.ACM_SIZE is zero?

BuildLoader.py
```diff
@@ -339,13 +340,24 @@ def update_fit_table (self):
 			if len(u_code_images) > 0:
 				raise Exception('  Insufficient uCode entries in FIT. Need %d more.' % len(u_code_images))
 
-		if self._board.ACM_SIZE > 0:
+			acm3_index = 0
+			if self._board.ACM_SIZE == 0:
+				acm3_index       = num_fit_entries
+				num_fit_entries     += 1
```
myguitar: No need to have a dependency on UCODE.
Instead, update acm3_index and num_fit_entries after line#406 (else:) if ACM3_SIZE > 0.

BuildLoader.py
```diff
@@ -394,6 +406,14 @@ def update_fit_table (self):
 		else :
 			addr = fit_address.value + (num_fit_entries + 1) * 16
 
+		# Add ACM3 with the reserved fit entry saved
+		if self._board.ACM3_SIZE > 0:
+			addr = self._board.ACM3_BASE
```
myguitar: Updating addr here may break spaceleft calculation.

BuildLoader.py
```diff
@@ -394,6 +406,14 @@ def update_fit_table (self):
...
+			module_size = self._board.ACM3_SIZE
+			fit_entry = FitEntry.from_buffer(rom, fit_offset + (acm3_index+1)*16)
+			fit_entry.set_values(addr, module_size, 0x100, 0x3, 0)
```
myguitar: For spaceleft,
fit_entry.set_values(self._board.ACM3_BASE, module_size, 0x100, 0x3, 0)

BuildLoader.py
sindhuragrandhi: Agree

BuildLoader.py
sindhuragrandhi: Ok, will update it

BuildLoader.py
```diff
@@ -393,6 +400,15 @@ def update_fit_table (self):
 
 		else :
 			addr = fit_address.value + (num_fit_entries + 1) * 16
+			if self._board.ACM3_SIZE > 0:
```
myguitar: These 3lines must be located before line#402 "addr = fit_address.value + (num_fit_entries + 1) * 16".

BuildLoader.py
```diff
@@ -744,6 +761,10 @@ def create_platform_vars (self):
 			if acm_base & 0x7FFF:
 				raise Exception ('ACM base[FSP-T+CAR:0x%x] must be 32KB aligned!' % acm_base)
 
+		if getattr(self._board, 'ACM3_SIZE') > 0:
+			acm3_base = getattr(self._board, 'ACM3_BASE')
+			if acm3_base & 0xFF:
+				raise Exception ('ACM3 base[FSP-T+CAR:0x%x] must be 4KB aligned!' % acm3_base)
```
gdong1: from comment, ACM3 is 4KB aligned. should line 765 be changed to "if acm3_base & 0x0FFF:"?

sindhuragrandhi: Good catch. Its a typo. Will change it.

PR #306, opened 2019-08-30
Generate BPMGen2 parameters for CFL/WHL

raghavag:
This patch will patch parameters for BpmGen2
based on the example file provided in the BpmGen2
package.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
@@ -97,8 +97,29 @@ def get_config ():
       'ecptr'     :   'EC/ecregionpointer.bin',
       'gbe'       :   'GBE/gbe.bin',
       'bpm'       :   'bpmgen2',
+      'bpm_params':   'bpmgen2/Example.bpDef',
     }
     return cfg_var
+
+def get_bpmgen2_params_change_list ():
+    params_change_list = []
+    params_change_list.append ([
+      # variable                | value |
+      # ===================================
+      ('PlatformRules',         'CFL Client'),
+      ('BpmStrutVersion',       '0x10'),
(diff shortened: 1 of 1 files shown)


Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -97,8 +97,29 @@ def get_config ():
       'ecptr'     :   'EC/ecregionpointer.bin',
       'gbe'       :   'GBE/gbe.bin',
       'bpm'       :   'bpmgen2',
+      'bpm_params':   'bpmgen2/Example.bpDef',
```
gdong1: Where user could get bpmgen2/Example.bpDef?

raghavag: It is from BPMGen2 package. they get it when they download BpmGen2 package

gdong1: Got it.

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -97,8 +97,29 @@ def get_config ():
...
+      ('AcmRevocation',         '2'),
+      ('NEMPages',              '3'),
+      ('IbbFlags',              '0x2'),
+      ('TxtInclude',            'FALSE'),
+      ('BpmSigScheme',          '0x14:RSASSA'),
+      ('BpmSigPubKey',          'keys\pubkey.pem'),
+      ('BpmSigPrivKey',         'keys\privkey.pem'),
+      ])
+    return params_change_list
+
```
gdong1: Not sure if we could a generic StitchIfwi.py by moving this platform specific config to a config file and let StitchIfwi.py 
 consume it. any idea?

PR #311, opened 2019-09-04
BootloaderCore: Handle BIST failure in SecEntry

CunningLearner:
- add BistVal in STAGE1A_ASM_HOB structure
- use MM0 register to preserve the BIST value
- push BistVal while setup HOB in stack
- add check for CPU BIST failure and halt the system when failed

Signed-off-by: Himanshu Sahdev aka CunningLearner <sahdev.himan@gmail.com>

### DIFF
--- a/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
+++ b/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
@@ -61,13 +61,12 @@ FspApiSuccess:
         ;
         mov     esp, ecx
         add     esp, dword [ASM_PFX(PcdGet32(PcdStage1StackSize))]
-        mov     eax, esp
 
-        movd    ebx, mm0
+        movd    eax, mm0
         emms                         ; Exit MMX Instruction
 
         ; Setup HOB
-        push    ebx                  ; BistVal
+        push    eax                  ; BistVal
         push    edi                  ; TimeStamp[0] [63:32]
         push    esi                  ; TimeStamp[0] [31:0]
         push    edx                  ; CarTop
--- a/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
+++ b/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
@@ -64,6 +64,7 @@ FspApiSuccess:
         mov     eax, esp
 
         movd    ebx, mm0
+        emms                         ; Exit MMX Instruction
 
         ; Setup HOB
         push    ebx                  ; BistVal
--- a/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
+++ b/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
@@ -24,8 +24,8 @@ extern  ASM_PFX(FspTempRamInit)
 
 global  ASM_PFX(_ModuleEntryPoint)
 ASM_PFX(_ModuleEntryPoint):
-        mov     ebx, eax
-        
+        movd    mm0, eax
+
         ;
         ; Read time stamp
         ;
@@ -62,6 +62,8 @@ FspApiSuccess:
         mov     esp, ecx
         add     esp, dword [ASM_PFX(PcdGet32(PcdStage1StackSize))]
         mov     eax, esp
+
+        movd    ebx, mm0
 
         ; Setup HOB
         push    ebx                  ; BistVal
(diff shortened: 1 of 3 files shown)


BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
```diff
@@ -62,6 +64,7 @@ FspApiSuccess:
         mov     eax, esp
 
         ; Setup HOB
+        push    ebx                  ; BistVal
```
myguitar: I don't think EBX is preserved here. Only EDI and ESI are preserved. You may need to use EDI. Can you also add a debug message in Stage1A.c to see if BistVal is expected one?

mauricema: I think we can put it back into MM0 again.  
FSP 2.0 TempRamInit() API preserves the following general purpose registers EBX, EDI, ESI, EBP and the following floating point registers MM0, MM1.

mauricema: Agree.  if we see a CPU BIST error,  we might want to print this error and then Halt the CPU.  This needs to be done after the serial port initialization.

CunningLearner: Thanks for the valuable points, I used MM0 register to hold the value and as it did not need to be accessed by pushing and popping in the same way as the FPU registers, I pushed the EBX register. Also, I debug the same as @myguitar suggested and confirmed with the expected value, and put a valid check and halt the CPU when BIST failure!

BootloaderCorePkg/Stage1A/Stage1A.c
```diff
@@ -240,6 +240,9 @@ SecStartup2 (
     DEBUG ((DEBUG_INIT, "\n%a\n", mBootloaderName));
   }
 
+  if(Stage1aAsmHob->BistVal != 0)
```
myguitar: a space between if and (.

if (Stage1aAsmHob->BistVal != 0) {
  CpuHalt ("BIST failure!\n");
}

BootloaderCorePkg/Stage1A/Stage1A.c
```diff
@@ -240,6 +240,9 @@ SecStartup2 (
...
+    CpuHalt ("BIST failure!\n");
```
mauricema: We can make it more specific by saying "CPU BIST failure!"

BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
mauricema: Line 64 "mov     eax, esp", the "eax" is not used at all after the assignment.   So let us avoid another register "ebx" just for pushing it into stack.   I think we can safely remove line 64 and use "eax" to push into stack.

BootloaderCorePkg/Stage1A/Stage1A.c
CunningLearner: Ack

BootloaderCorePkg/Stage1A/Stage1A.c
CunningLearner: Done

BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
```diff
@@ -61,7 +63,11 @@ FspApiSuccess:
         add     esp, dword [ASM_PFX(PcdGet32(PcdStage1StackSize))]
         mov     eax, esp
 
+        movd    ebx, mm0
```
mauricema: I did put a comment in previous commit.  Can we remove line 64  and then do  below ?
movd  eax, mm0
push   eax

BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
CunningLearner: It was actually hit me yesterday, but I was not sure why it was actually there, maybe intentionally!?

Well, I successfully boot with eax in place, implies we can remove it!

BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
CunningLearner: Done!

BootloaderCorePkg/Stage1A/Stage1A.c
```diff
@@ -240,6 +240,9 @@ SecStartup2 (
     DEBUG ((DEBUG_INIT, "\n%a\n", mBootloaderName));
   }
 
+  if (Stage1aAsmHob->BistVal != 0)
+    CpuHalt ("CPU BIST failure!\n");
```
mauricema: Minor comments on coding styles.  :)
According to EDK2 coding style, it is required to add curly brace {} after if condition even if it has only single statement after if.  So it needs to be:
if (Stage1aAsmHob->BistVal != 0) {
    CpuHalt ("CPU BIST failure!\n");
}
Aiden also pointed it out in his previous comments.

CunningLearner: Done

PR #313, opened 2019-09-04
[ACPI] Add a feature to create MADT in runtime

myguitar:
Create MADT in runtime instead of integrating MADT to AcpiTable section
- Controlled by ENABLE_CREATE_MADT in each BoardConfig.py
- ENABLE_CREATE_MADT is disabled by default for now


BootloaderCorePkg/BootloaderCorePkg.dec
```diff
@@ -112,6 +112,18 @@
...
+  gPlatformModuleTokenSpaceGuid.PcdIntrSrcSciTrigger      |          1 | UINT16 | 0x200000E6
+  gPlatformModuleTokenSpaceGuid.PcdIntrSrcSciPolarity     |          1 | UINT16 | 0x200000E7
+  gPlatformModuleTokenSpaceGuid.PcdLocalApicNmiCount      |          1 | UINT32 | 0x200000E8
+  gPlatformModuleTokenSpaceGuid.PcdLocalApicNmiTrigger    |          1 | UINT16 | 0x200000E9
+  gPlatformModuleTokenSpaceGuid.PcdLocalApicNmiPolarity   |          0 | UINT16 | 0x200000EA
+  gPlatformModuleTokenSpaceGuid.PcdLocalApicNmiLintIdx    |          1 | UINT8  | 0x200000EB
+  gPlatformModuleTokenSpaceGuid.PcdLocalApicAddress       | 0xFEE00000 | UINT32 | 0x200000EC
+  gPlatformModuleTokenSpaceGuid.PcdIoApicAddress          | 0xFEC00000 | UINT32 | 0x200000ED
+  gPlatformModuleTokenSpaceGuid.PcdIoApicId               |          1 | UINT8  | 0x200000EE
+
```
gdong1: It would be great if adding description for these new definitions.

myguitar: There is some description in BuildLoader.py. Let me add more at the same place.

sindhuragrandhi: Will all of these PCDs have to change for each project? Seems like quite a bit of configuration. If some of them are constant, then we may want to reduce these.

gdong1: BuildLoader.py is just one of the consumer to this PCD. DEC file is the place to have full description by EDK2.

myguitar: Mostly not changed. Current Madt.h of each platforms also has fixed value, but build time configurable.

Some PCD are Silicon specific. For example,
- QEMU/APL
  PcdIntrSrcSciPolarity: Active Low
  PcdLocalApicNmiCount: 1
  PcdLocalApicNmiTrigger: Level Trigger
- WHL/CFL
  PcdIntrSrcSciPolarity: Active High
  PcdLocalApicNmiCount: 16
  PcdLocalApicNmiTrigger: Edge Trigger

Other platforms may also have these changes only.

myguitar: Okay, let me put description in DEC file.

mauricema:
Just my opinion,  PCD might not be the best way to specify MADT customization since there are too many platform specific information to define.  Do you think it is better to make platform provide MADT table entries except for CPU local APIC type ?   In this way, the common code just need to append the CPU local APIC entries to MADT.

myguitar:
> Just my opinion, PCD might not be the best way to specify MADT customization since there are too many platform specific information to define. Do you think it is better to make platform provide MADT table entries except for CPU local APIC type ? In this way, the common code just need to append the CPU local APIC entries to MADT.

It can be simply done if AcpiInit() allocates a memory from LOW to HIGH memory. But, current AcpiInit() allocates necessary memory from HIGH to LOW with ACPI Table length, copies the content and hooks PlatformUpdateAcpiTable(). It makes the hook hard to change its Table length or to create contents dynamically. What about changing AcpiInit() to allocate memory from LOW(BASE) to HIGH(TOP)?

mauricema:
> > Just my opinion, PCD might not be the best way to specify MADT customization since there are too many platform specific information to define. Do you think it is better to make platform provide MADT table entries except for CPU local APIC type ? In this way, the common code just need to append the CPU local APIC entries to MADT.
> 
> It can be simply done if AcpiInit() allocates a memory from LOW to HIGH memory. But, current AcpiInit() allocates necessary memory from HIGH to LOW with ACPI Table length, copies the content and hooks PlatformUpdateAcpiTable(). It makes the hook hard to change its Table length or to create contents dynamically. What about changing AcpiInit() to allocate memory from LOW(BASE) to HIGH(TOP)?

I think it is a good idea to change the ACPI memory allocation from low to high.

myguitar:
Thanks everyone for all your comments. This will be covered by #314 and #292 . Let me close this PR.

PR #314, opened 2019-09-06
[AcpiInitLib] Allocate ACPI tables from low to high memory

myguitar:
To make PlatformUpdateAcpiTable() hook its contents and length easily,
allocate a memory from low to high and appends each tables to higher direction

### DIFF
--- a/BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
+++ b/BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
@@ -25,8 +25,8 @@
 #include <Library/FirmwareUpdateLib.h>
 #include <Guid/BootLoaderVersionGuid.h>
 
-#define  ACPI_ALLOC(x)       (Current = (UINT8 *)(((UINT32)Current - (x)) & ~0x0F))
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Library/AcpiInitLib/AcpiInitLib.c
```diff
@@ -310,6 +316,11 @@ AcpiTableUpdate (
     DEBUG ((DEBUG_ERROR, "Acpi update failed - %r\n", Status));
   }
 
+  //
+  // Update AcpiTop for gLoaderSystemTableInfoGuid
+  //
+  S3Data->AcpiTop = (UINT32)(UINTN)Current;
```
mauricema: You might want to check if Current is still within the ACPI memory range.   It is only checked at the beginning of the loop.   Since the length might have been modified in hook,  better to check again.

myguitar: > You might want to check if Current is still within the ACPI memory range. It is only checked at the beginning of the loop. Since the length might have been modified in hook, better to check again.

Right, good point. Let me check the range again at the final point. Thanks.

PR #321, opened 2019-09-13
Add DBG2 ACPI table infrastructure

stalamudupula:
This patch provides the basic infrastructure to add a
Debug Port Table  2 (DBG2) to specify one or more ports
for debugging purposes. More info reg DBG2 @ :
https://docs.microsoft.com/en-us/previous-versions/windows/hardware/design/dn639131(v=vs.85)?redirectedfrom=MSDN

If the platform wants to report a debug port to Windows,
it should patch the DBG2 template provided with the
corresponding debug port information. And this updated
DBG2 must be referenced in RSDT.

### DIFF
--- a/Platform/CommonBoardPkg/AcpiTables/Dbg2/Dbg2.act
+++ b/Platform/CommonBoardPkg/AcpiTables/Dbg2/Dbg2.act
@@ -1,0 +1,94 @@
+/** @file
+  This file contains a structure definition for the
+  Debug Port Table 2 (DBG2). If a platform wants to report
+  a debug port, it should include this file in ACPI build, and
+  it should patch the Debug Device Info structure with the
+  corresponding debug port information.
+
+  Copyright (c) 2018 - 2019, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+//
+// Statements that include other files
+//
+
+#include <Dbg2.h>
+
+//
+// Multiple APIC Description Table
+//
+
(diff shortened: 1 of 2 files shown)


Platform/CommonBoardPkg/AcpiTables/Dbg2/Dbg2.act
```diff
@@ -0,0 +1,93 @@
+/** @file
+  This file contains a structure definition for the ACPI 2.0 Multiple APIC
+  Description Table (MADT).  Any changes to the MADT table require updating the
```
gdong1: MADT?

gdong1: and do we need add this file to CFL/APL build?

stalamudupula: Corrected the headers in new patch. Yes, platform needs to include it in AcpiTables.inf, if required.

mauricema:
EDK2 has  MdePkg\Include\IndustryStandard\DebugPort2Table.h,  do you think we align with that for structure definition ?

stalamudupula:
> EDK2 has MdePkg\Include\IndustryStandard\DebugPort2Table.h, do you think we align with that for structure definition ?

The current implementation infact follows the same header format as the Edk2Open header. Except, Debug Device Info structures are also statically assigned enough memory in SBL. Did not want to allocate memory for these structs at run time, as Edk2 code does.

mauricema:
OK

PR #322, opened 2019-09-13
Fix potential issue from external cfg data

gdong1:
Currently SpiLoadExternalConfigData will update the output buffer even
signature or other checks failure. It may cause issue to caller when
caller consume the data.
This patch update the logic to update output buffer only load external
config data success.

### DIFF
--- a/Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
+++ b/Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
@@ -141,7 +141,7 @@ SpiLoadExternalConfigData (
   UINT32       Address;
   UINT32       BlobSize;
   UINT8       *Buffer;
-  CDATA_BLOB  *CfgBlob;
+  CDATA_BLOB  CfgBlob;
   UINT32       SignedLen;
   UINT32       CfgDataLoadSrc;
   UINT32       Base;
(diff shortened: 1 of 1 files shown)


Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -164,11 +163,11 @@ SpiLoadExternalConfigData (
   //
   Status = EFI_UNSUPPORTED;
   if (CfgDataLoadSrc == FlashRegionPlatformData) {
-    Status = SpiFlashRead (FlashRegionPlatformData, Address, BlobSize, Buffer);
+    Status = SpiFlashRead (FlashRegionPlatformData, Address, BlobSize, (UINT8*)&CfgBlob);
```
mauricema: Add a space between  UINT and *  for  UINT8*

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -141,7 +141,7 @@ SpiLoadExternalConfigData (
   UINT32       Address;
   UINT32       BlobSize;
   UINT8       *Buffer;
-  CDATA_BLOB  *CfgBlob;
+  CDATA_BLOB  CfgBlob;
```
mauricema: align the variable names ?

mauricema:
What was the real issue again here?   
I think the output buffer can be updated in any case (either fail or success).   Caller needs to rely on 'Status' returned to determine if the data is good to consume.   If Status indicates failure, even if the data buffer is filled,  it should be ignored by the caller.

gdong1:
After double check the caller, looks the caller doesn't use these data if status is not success.
so this patch is not necessary.

PR #327, opened 2019-09-23
Add support to boot kernel from container

stalamudupula:
In addition to supporting IAS format for loading
kernel images, this patch adds support to load
the images present in a container. Set the token
ENABLE_PODS_BOOT to 1 to achieve this. Default is
0 i.e, IAS image is expected by default.

GenContainer.py tool can be used to create a container
with the boot files as follows:

python %SBL_ROOT%\BootloaderCorePkg\Tools\GenContainer.py create
-cl CMDL:<cmdline.txt> KRNL:<vmlinuz> INRD:<initrd>
-od <OutDir> -kd %SBL_ROOT%\BootloaderCorePkg\Tools\Keys

<cmdline.txt> = command line file
<vmlinuz>     = kernel image
<initrd>      = initrd image
<OutDir>      = dir where final PodsImage.bin is copied

### DIFF
--- a/BootloaderCommonPkg/Include/Library/ContainerLib.h
+++ b/BootloaderCommonPkg/Include/Library/ContainerLib.h
@@ -23,6 +23,12 @@
 #define AUTH_TYPE_SIG_RSA2048_SHA256  3
 #define AUTH_TYPE_SIG_RSA3072_SHA384  4
 
+// Flags for CONTAINER_HDR
+#define CONTAINER_HDR_FLAG_MONO_SIGNING     BIT0
+
+// Attributes for COMPONENT_ENTRY
+#define COMPONENT_ENTRY_ATTR_RESERVED       BIT7
+
 typedef VOID (*LOAD_COMPONENT_CALLBACK) (UINT32 ProgressId);
 
 typedef struct {
@@ -55,35 +61,12 @@ typedef struct {
   UINT32           Name;
   UINT32           Offset;
   UINT32           Size;
-  UINT8            Rserved;
+  UINT8            Attribute;
   UINT8            Alignment;
   UINT8            AuthType;
   UINT8            HashSize;
   UINT8            HashData[0];
 } COMPONENT_ENTRY;
-
-
-/**
-  Locate a component information from a container.
-
-  @param[in]     ContainerSig       Container signature.
-  @param[in]     ComponentName      Component name.
-  @param[in,out] ContainerEntryPtr  Pointer to receive container entry info.
-  @param[in,out] ComponentEntryPtr  Pointer to receive component entry info.
-
-  @retval EFI_UNSUPPORTED          Unsupported AuthType.
-  @retval EFI_NOT_FOUND            Cannot locate component.
-  @retval EFI_SECURITY_VIOLATION   Authentication failed.
-  @retval EFI_SUCCESS              component region is located successfully.
-
-**/
-EFI_STATUS
-LocateComponentEntry (
-  IN      UINT32                  ContainerSig,
-  IN      UINT32                  ComponentName,
-  IN OUT  CONTAINER_ENTRY       **ContainerEntryPtr,
-  IN OUT  COMPONENT_ENTRY       **ComponentEntryPtr
-  );
 
 
 /**
@@ -159,5 +142,77 @@ LoadComponent (
   IN OUT VOID    **Buffer,
   IN OUT UINT32   *Length
   );
+
+/**
+  Locate a component information from a container.
+
+  @param[in]     ContainerSig       Container signature.
+  @param[in]     ComponentName      Component name.
+  @param[in,out] ContainerEntryPtr  Pointer to receive container entry info.
+  @param[in,out] ComponentEntryPtr  Pointer to receive component entry info.
+
+  @retval EFI_UNSUPPORTED          Unsupported AuthType.
+  @retval EFI_NOT_FOUND            Cannot locate component.
+  @retval EFI_SECURITY_VIOLATION   Authentication failed.
+  @retval EFI_SUCCESS              component region is located successfully.
+
+**/
+EFI_STATUS
+LocateComponentEntry (
+  IN      UINT32                  ContainerSig,
+  IN      UINT32                  ComponentName,
+  IN OUT  CONTAINER_ENTRY       **ContainerEntryPtr,
+  IN OUT  COMPONENT_ENTRY       **ComponentEntryPtr
+  );
+
+/**
+  Get Next Component in a container.
+
+  @param[in]       ContainerSig    Container signature or component type.
+  @param[in, out]  ComponentName   Pointer to receive next component name.
+
+  @retval EFI_INVALID_PARAMETER    ComponentName is NULL.
+  @retval EFI_NOT_FOUND            Cannot find next component.
+  @retval EFI_SUCCESS              Next component was found successfully.
+
+**/
+EFI_STATUS
+EFIAPI
+GetNextAvailableComponent (
+  IN     UINT32     ContainerSig,
+  IN     UINT32    *ComponentName
+);
+
+/**
+  This function registers a container.
+
+  @param[in]  ContainerBase      Container base address to register.
+
+  @retval EFI_NOT_READY          Not ready for register yet.
+  @retval EFI_BUFFER_TOO_SMALL   Insufficant max container entry number.
+  @retval EFI_OUT_OF_RESOURCES   No space to add new container.
+  @retval EFI_SUCCESS            The container has been registered successfully.
+
+**/
+EFI_STATUS
+RegisterContainer (
+  IN  UINT32   ContainerBase
+  );
+
+/**
+  This function unregisters a container with given signature.
+
+  @param[in]  Signature      Container signature.
+
+  @retval EFI_NOT_READY          Not ready for unregister yet.
+  @retval EFI_NOT_FOUND          Not container available for unregisteration.
+  @retval EFI_SUCCESS            The container has been unregistered successfully.
+
+**/
+EFI_STATUS
+UnregisterContainer (
+  IN  UINT32   Signature
+  );
+
 
 #endif
--- a/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
+++ b/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
@@ -24,89 +24,6 @@
 #define  IS_FLASH_ADDRESS(x)   (((UINT32)(UINTN)(x)) >= 0xF0000000)
 
 /**
-  This function registers a container.
-
-  @param[in]  ContainerBase      Container base address to register.
-
-  @retval EFI_NOT_READY          Not ready for register yet.
-  @retval EFI_BUFFER_TOO_SMALL   Insufficant max container entry number.
-  @retval EFI_OUT_OF_RESOURCES   No space to add new container.
-  @retval EFI_SUCCESS            The container has been registered successfully.
-
-**/
-STATIC
-EFI_STATUS
-RegisterContainer (
-  IN UINT32    ContainerBase
-  )
-{
-  CONTAINER_LIST       *ContainerList;
-  CONTAINER_HDR        *ContainerHdr;
-  UINT32                Index;
-  VOID                 *Buffer;
-
-  ContainerList = (CONTAINER_LIST *)GetContainerListPtr ();
-  if (ContainerList == NULL) {
-    return EFI_NOT_READY;
-  }
-
-  Index = ContainerList->Count;
-  if (Index >= PcdGet32 (PcdContainerMaxNumber)) {
-    return EFI_BUFFER_TOO_SMALL;
-  }
-
-  ContainerHdr = (CONTAINER_HDR *)ContainerBase;
-  Buffer = AllocatePool (ContainerHdr->DataOffset);
-  if (Buffer == NULL) {
-    return  EFI_OUT_OF_RESOURCES;
-  }
-
-  ContainerList->Entry[Index].Signature   = ContainerHdr->Signature;
-  ContainerList->Entry[Index].HeaderCache = (UINT32)Buffer;
-  ContainerList->Entry[Index].Base        = ContainerBase;
-  CopyMem (Buffer, (VOID *)ContainerBase, ContainerHdr->DataOffset);
-  ContainerList->Count++;
-
-  return EFI_SUCCESS;
-}
-
-/**
-  This function unregisters a container in the last entry.
-
-  @retval EFI_NOT_READY          Not ready for unregister yet.
-  @retval EFI_NOT_FOUND          Not container available for unregisteration.
-  @retval EFI_SUCCESS            The container has been unregistered successfully.
-
-**/
-STATIC
-EFI_STATUS
-UnregisterLastContainer (
-  VOID
-  )
-{
-  EFI_STATUS            Status;
-  CONTAINER_LIST       *ContainerList;
-  UINT32                Index;
-
-  ContainerList = (CONTAINER_LIST *)GetContainerListPtr ();
-  if (ContainerList == NULL) {
-    return EFI_NOT_READY;
-  }
-
-  Index = ContainerList->Count;
-  if (Index > 0) {
-    Index--;
-    FreePool ((VOID *)ContainerList->Entry[Index].HeaderCache);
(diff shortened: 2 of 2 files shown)


BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -540,7 +625,8 @@ def main():
 
 	# Command for create
 	cmd_display = sub_parser.add_parser('create', help='create a container image')
-	cmd_display.add_argument('-l',  dest='layout',   type=str, required=True, help='Container layout intput file')
+	cmd_display.add_argument('-l',  dest='layout',   type=str, help='Container layout intput file')
```
mauricema: can we add some check for the exclusion for '-l' and '-cl' ?

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -318,12 +424,15 @@ LoadAndSetupImage (
   }
   AddMeasurePoint (0x4070);
 
-  if ((LoadedImage->Flags & LOADED_IMAGE_IAS) != 0) {
+  if (FeaturePcdGet (PcdPodsBootEnabled)) {
+    Status = ParsePodsImage (BootOption, LoadedImage);
+  }
+  else if ((LoadedImage->Flags & LOADED_IMAGE_IAS) != 0) {
```
mauricema: put into the same line as "}"

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -318,12 +424,15 @@ LoadAndSetupImage (
   }
   AddMeasurePoint (0x4070);
 
-  if ((LoadedImage->Flags & LOADED_IMAGE_IAS) != 0) {
+  if (FeaturePcdGet (PcdPodsBootEnabled)) {
```
mauricema: can we do similar check as IAS like :
if ((LoadedImage->Flags & LOADED_IMAGE_PODS) != 0)   ?

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -11,80 +11,68 @@ UINT32  *mPreOsCheckerEntry;
...
-      DEBUG ((DEBUG_INFO, "One file in IAS file, take it as bzImage .... \n"));
+      DEBUG ((DEBUG_INFO, "One file in IAS/PODS file, take it as bzImage .... \n"));
     }
     return EFI_SUCCESS;
   }
 
-  ImageType = IAS_IMAGE_TYPE (IasImage->ImageType);
-  DEBUG ((DEBUG_INFO, "IAS Image Type = 0x%x\n", ImageType));
-  if (ImageType == IAS_TYPE_CLASSIC) {
+  if ((ImageType & 0xF) == IAS_TYPE_CLASSIC) {
```
mauricema: can we pass in the processed ImageType to this function ?
It means the caller do the ImageType conversion before calling it.

stalamudupula:
> In your examples, did you miss kernel file and initrd file ? KRNL: , INIT:, etc ?

Some formatting issue while showing that message here, otherwise my commit msg elsewhere looks fine.

BootloaderCommonPkg/Include/Library/ContainerLib.h
```diff
@@ -23,6 +23,12 @@
 #define AUTH_TYPE_SIG_RSA2048_SHA256  3
 #define AUTH_TYPE_SIG_RSA3072_SHA384  4
 
+// Flags for CONTAINER_HDR
+#define CONTAINER_HDR_FLAG_MONO_SIGNING     BIT0
+
+// Attributes for COMPONENT_ENTRY
+#define COMPONENT_ENTRY_ATTR_RESERVED       BIT7
+
```
gdong1: Could we add more comments? not sure why COMPONENT_ENTRY_ATTR_RESERVED is required.
Container has subregion, pod has't subregion, only has files, both using container header, how we know it is container or pod image, by CONTAINER_HDR_FLAG_MONO_SIGNING?

gdong1:
> > In your examples, did you miss kernel file and initrd file ? KRNL: , INIT:, etc ?
> 
> Some formatting issue while showing that message here, otherwise my commit msg elsewhere looks fine.
If there are multiple files, 1) do we need support to specify alignment? 2) do we support compress multiple files together instead of compress them separately?

BootloaderCommonPkg/Include/Library/ContainerLib.h
stalamudupula: CONTAINER and POD are interchangeable. CONTAINER or POD has got the region/components inside them.

gdong1: has sub-region in the container need verify header firstly, and has files in the container doesn't need verify the header separately since it will be verified with its payload. so I guess you are using CONTAINER_HDR_FLAG_MONO_SIGNING to diff them.

gdong1: what COMPONENT_ENTRY_ATTR_RESERVED is used for?

stalamudupula: COMPONENT_ENTRY_ATTR_RESERVED is used to set the _SG_ (last entry in the container) component as reserved, so that cant be updated or touched.

gdong1: What the last entry used for (to get authtype for container header?). could we define a entry signature, e.g. "_HDR", instead of adding a new attribute field.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -337,6 +429,69 @@ LocateComponentEntry (
...
+  ContainerHdr = (CONTAINER_HDR *)ContainerEntry->HeaderCache;
+  if (ContainerHdr->Count == 0) {
+    return Status;
+  }
+
+  CurrEntry = (COMPONENT_ENTRY *)&ContainerHdr[1];
+  for (Index = 0; Index < (UINT32)(ContainerHdr->Count - 1); Index++) {
+    NextEntry = (COMPONENT_ENTRY *)((UINT8 *)(CurrEntry + 1) + CurrEntry->HashSize);
+    if ((CurrEntry->Attribute & COMPONENT_ENTRY_ATTR_RESERVED) == 0) {
+      if ((NextEntry->Attribute & COMPONENT_ENTRY_ATTR_RESERVED) == 0) {
```
myguitar: Why current and next need to be checked?

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -337,6 +429,69 @@ LocateComponentEntry (
...
+        if (*ComponentName == 0) {
```
myguitar: if ComponentName is NULL?

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -246,6 +271,98 @@ AuthenticateComponent (
...
+  )
+{
+  EFI_STATUS                Status;
+  CONTAINER_HDR            *ContainerHdr;
+  CONTAINER_ENTRY          *ContainerEntry;
+  COMPONENT_ENTRY          *CompEntry;
+  UINT8                    *AuthData;
+  UINT32                    ContainerHdrSize;
+  UINT8                     AuthType;
+  UINT64                    SignatureBuffer;
```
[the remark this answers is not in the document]
stalamudupula: This is only for printing purpose.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -246,6 +271,98 @@ AuthenticateComponent (
...
+  UINT32                    Index;
+  LOADER_COMPRESSED_HEADER *CompressHdr;
+
+  ContainerHdr     = (CONTAINER_HDR *)ContainerBase;
+  SignatureBuffer  = ContainerHdr->Signature;
+  DEBUG ((EFI_D_INFO, "Registering container %4a\n", (CHAR8 *)&SignatureBuffer));
+
+  // Register container temporarily
+  Status = RegisterContainerInternal (ContainerBase);
+  if (!EFI_ERROR (Status)) {
```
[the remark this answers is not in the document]
stalamudupula: RegisterContainerInternal() is a private function, and others need not call this. Only RegisterContainer() is exposed outside. Based on the internal register return value, we decide to authenticate the component or not. But internal needs to deal with only registering, not authenticating.

WIll change the comment , temporarily means, if theres an error, we might unregister it soon.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -246,6 +271,98 @@ AuthenticateComponent (
...
+          DataLen  = CompEntry->Offset;
+          Status   = AuthenticateComponent (DataBuf, DataLen, CompEntry->AuthType,
+                                            AuthData, CompEntry->HashData, COMP_TYPE_INVALID);
+        }
+      }
+    }
+
+    if (EFI_ERROR (Status)) {
+      // Unregister the container since authentication failed
+      UnregisterContainer (0xFFFFFFFF);
```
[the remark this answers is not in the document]
stalamudupula: Will move the code back to have the UnregisterLastContainer (). This way we can have unregister by index.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
stalamudupula: Good catch. Need to redo the code here. Basic intent is to support if there is only 1 component and second component is _SG_ (last component)

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
stalamudupula: Already handled at the beginning.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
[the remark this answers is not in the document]
myguitar: I think reg() and de-reg() is not good approach in failure case. What about authenticating first, then reg it if there is no error? In that case, de-reg() is not required.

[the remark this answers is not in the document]
stalamudupula: Unregister() is still required. Imagine there are 2 PODS with same container signature. We can unregister first one after we load the compinents. In that way, registration and loading other container doesnt throw error.

BootloaderCommonPkg/Include/Library/ContainerLib.h
mauricema: The last entry holds the AUTH data for the whole POD image.  Instead of adding the auth data at the end, we reuse a normal component entry to provide the auth info. In this way, the flow is the same and the auth algo can be determined use the same way as normal component.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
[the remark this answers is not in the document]
mauricema: I would agree reg/de-reg might not be the best way to handle it in this case.  But by doing so, it makes the code simpler to handle since all functions rely on the GetContainerPtr() pointer to get container info.  Before registration, the info cannot be retrieved.   We can for sure do the other way, but need more code to deal with it.  Since only RegisterContianer() is external API.  I think as long as we handle it carefully inside this lib,  it should be fine.  We can put some notes there if required.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -337,6 +429,71 @@ LocateComponentEntry (
...
+        *ComponentName = CurrEntry->Name;
+        Status = EFI_SUCCESS;
+        break;
+      } else if (*ComponentName == CurrEntry->Name) {
+        if ((NextEntry->Attribute & COMPONENT_ENTRY_ATTR_RESERVED) == 0) {
+          *ComponentName = NextEntry->Name;
+          Status = EFI_SUCCESS;
+          break;
+        } else {
+          Status = EFI_NOT_FOUND;
```
mauricema: if next entry is reserved,  shouldn't we continue to find next normal entry and return it ?

stalamudupula: I am assuming mono sig component or all reserved components to be at the end of the container. If not, I can check for count and continue to search for all other unreserved components.

mauricema: it should be that way.  but for code flow, let us do not assume that.  Let us clean the logic:
1. Find first non-reserved entry.
2. If  *Component=0, return this entry.
3. If  *Component = Entry->Name,  return next non-reserved entry.

BootloaderCommonPkg/Include/Library/ContainerLib.h
gdong1: there are AuthType in container header, and AuthType in entry header.  it is a little confusing to use the AuthType in the entry header for whole POD image.  do we use AuthType in container header in POD image case?

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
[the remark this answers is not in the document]
myguitar: Understood. For code readability, what about having separate AuthenticateInternal()?
RegisterContainer() {
  RegisterContainerInternal()
  AuthenticateInternal()
  UnregisterContainer()
}
One more thing, if PcdVerifiedBootEnabled is disabled, AuthenticateComponent() returns EFI_NOT_FOUND. Need to handle disabled VerifiedBoot properly.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
myguitar: The API name - GetNextComponent() is a little confusing. It looks like just an iteration. So, GetNextAvailableComponent() would be better?

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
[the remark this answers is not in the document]
gdong1: for POD image (just like IAS image), do we need call register? I have thought only container that have sub-region need register.

BootloaderCommonPkg/Include/Library/ContainerLib.h
mauricema: Yes, we still use the container header AuthType to verify the header only.   Header authentication only control header, not the POID image contents.   I think using a reserved entry to do the POD signing is a cleaner solution.  In this way for POD image, it follows the same flow as normal container verification.  There is no difference at all.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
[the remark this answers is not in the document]
mauricema: @myguitar   Good point.  We can rename the function to make it easier to understand.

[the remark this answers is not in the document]
mauricema: @gdong1   We treat POD in the same way as container.  There is no difference to the library.   I think it is better to do things in a consistent way.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -246,6 +271,119 @@ AuthenticateComponent (
...
+  ContainerEntry   = GetContainerBySignature (ContainerHeader->Signature);
+  ContainerHdr     = (CONTAINER_HDR *)ContainerEntry->HeaderCache;
+  ContainerHdrSize = GetContainerHeaderSize (ContainerHdr);
+  if (ContainerHdrSize > 0) {
+    AuthType = ContainerHdr->AuthType;
+    AuthData = (UINT8 *)ContainerHdr + ALIGN_UP(ContainerHdrSize, AUTH_DATA_ALIGN);
+    if ((AuthType == AUTH_TYPE_NONE) && FeaturePcdGet (PcdVerifiedBootEnabled)) {
+      Status = EFI_SECURITY_VIOLATION;
+    } else {
+      Status = AuthenticateComponent ((UINT8 *)ContainerHdr, ContainerHdrSize,
```
myguitar: AuthenticateComponent() returns EFI_NOT_FOUND if PcdVerifiedBootEnabled is 0.

PR #329, opened 2019-09-26
Enable NOOPT target for debugging purpose

myguitar:
Compile optimization sometimes needs to be disabled for debugging.
EDKII BaseTools provide NOOPT target, so leverage it.
The default GCC '-O0' and VS '/Od' option results in huge size image,
so the optimization level is adjusted with approximately level.

Add a new build option '-no' or '--noopt' for NOOPT target
- Release build option '-r' will ignore '--noopt' option
ex) python BuildLoader.py build qemu --noopt

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dsc
+++ b/BootloaderCorePkg/BootloaderCorePkg.dsc
@@ -18,7 +18,7 @@
   DSC_SPECIFICATION                   = 0x00010005
   OUTPUT_DIRECTORY                    = Build/$(PLATFORM_NAME)
   SUPPORTED_ARCHITECTURES             = IA32|X64|ARC
-  BUILD_TARGETS                       = DEBUG|RELEASE
+  BUILD_TARGETS                       = DEBUG|RELEASE|NOOPT
   SKUID_IDENTIFIER                    = DEFAULT
   FLASH_DEFINITION                    = $(PLATFORM_NAME)/$(PLATFORM_NAME).fdf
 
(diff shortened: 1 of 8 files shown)


BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -396,4 +396,9 @@
   *_*_*_CC_FLAGS = -DLITE_PRINT
 !endif
 
-
+  # -O0 results in too big size. Override it to -O1 with lto
+!if $(TARGET) == NOOPT
+  *_GCC5_IA32_CC_FLAGS = -flto -O1
```
mauricema: can we make it more generic ?

BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -396,4 +396,9 @@
   *_*_*_CC_FLAGS = -DLITE_PRINT
 !endif
 
-
+  # -O0 results in too big size. Override it to -O1 with lto
+!if $(TARGET) == NOOPT
+  *_GCC5_IA32_CC_FLAGS = -flto -O1
+  *_GCC5_IA32_DLINK_FLAGS = -flto -O1
+  *_VS2015x86_IA32_CC_FLAGS = /Ox
```
mauricema: why only 2015 ?

BootloaderCorePkg/BootloaderCorePkg.dsc
myguitar: I will do as below.
*_GCC49_*_CC_FLAGS = -O1
*_GCC49_*_DLINK_FLAGS = -O1
*_GCC5_*_CC_FLAGS = -flto -O1
*_GCC5_*_DLINK_FLAGS = -flto -O1

BootloaderCorePkg/BootloaderCorePkg.dsc
myguitar: For now, let me override GCC only and remove VS override. For VS, default /Od (disable optimization) will be used.

myguitar:
Let me merger this later after double checking with various gcc version.

myguitar:
Verified with gcc-4.8, gcc-4.9, gcc-5, gcc-6, gcc-7, gcc-8 on ubuntu 18.04

PR #330, opened 2019-09-27
Fix cfgdata dword parsing issue

bteo1:
This patch to fix the GenCfgData tool that wrongly parsing the dword data from an array.
For exmaple: 
CfgData1 |  *| 0x8 {0x5fff, 0x7fff, ...}, but generate result for CfgData1[0] = 0 and so on.


bteo1:
I was using "git merge upstream/master" to update my fork in first submit.
Hope this round does fit the issue after using rebase.

PR #333, opened 2019-09-27
Update GenContainer.py to prepare PODS image

stalamudupula:
GenContainer.py tool can be used to create a container
with the boot files as follows:

python %SBL_ROOT%\BootloaderCorePkg\Tools\GenContainer.py create
-cl CMDL:<cmdline.txt> KRNL:<vmlinuz> INRD:<initrd>
-od <OutDir> -kd %SBL_ROOT%\BootloaderCorePkg\Tools\Keys

<cmdline.txt> = command line file
<vmlinuz> = kernel image
<initrd> = initrd image
<OutDir> = dir where final PodsImage.bin is copied

### DIFF
--- a/BootloaderCorePkg/Tools/GenContainer.py
+++ b/BootloaderCorePkg/Tools/GenContainer.py
@@ -20,11 +20,15 @@ class COMPONENT_ENTRY (Structure):
 		('name',        ARRAY(c_char, 4)),   # SBL pod entry name
 		('offset',      c_uint32),   # Component offset in byte from the payload (data)
 		('size',        c_uint32),   # Region/Component size in byte
-		('reserved',    c_uint8),    # reserved
+		('attribute',   c_uint8),    # Attribute:  BIT7 Reserved component entry
 		('alignment',   c_uint8),    # This image need to be loaded to memory  in (1 << Alignment) address
 		('auth_type',   c_uint8),    # Refer AUTH_TYPE_VALUE: 0 - "NONE"; 1- "SHA2_256";  2- "RSA2048SHA256"; 3- "SHA2_384"; 4 - RSA3072SHA384
 		('hash_size',   c_uint8)     # Hash data size, it could be image hash or public key hash
 		]
+
+	_attr = {
+	  'RESERVED' : 0x80
+	}
 
 	def __new__(cls, buf = None):
 		if buf is None:
@@ -50,10 +54,20 @@ class CONTAINER_HDR (Structure):
 		('data_offset',  c_uint16),         # Offset of payload (data) from header in byte
 		('data_size',    c_uint32),         # Size of payload (data) in byte
 		('auth_type',    c_uint8),          # Refer AUTH_TYPE_VALUE: 0 - "NONE"; 2- "RSA2048SHA256"; 4 - RSA3072SHA384
-		('image_type',   c_uint8),          # Not used for now
-		('flags',        c_uint8),          # Not used for now
+		('image_type',   c_uint8),          # 0: Normal
+		('flags',        c_uint8),          # BIT0: monolithic signing
 		('entry_count',  c_uint8),          # Number of entry in the header
 		]
+
+	_flags = {
+	  'MONO_SIGNING' : 0x01
+	}
+
+	_image_type = {
+	  'NORMAL'     :  0x00,
+	  'CLASSIC'    :  0xF3,
+	  'MULTIBOOT'  :  0xF4,
+	}
 
 	def __new__(cls, buf = None):
 		if buf is None:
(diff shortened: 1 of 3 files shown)


BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -328,76 +362,133 @@ def dump (self):
...
 
+		if image_type == '':
+			image_type = 'NORMAL'
+
 		if container_file == '':
 			container_file = container_sig + '.bin'
 		key_path = os.path.join(self.key_dir, key_file)
 
-		self.init (container_sig.encode(), alignment)
+    # build header
```
mauricema: pls adjust alignment

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -483,12 +585,34 @@ def gen_container_bin (container_list, out_dir, inp_dir, key_dir = '.', tool_dir
 		out_file = container.create (each)
 		print ("Container '%s' was created successfully at:  \n  %s" % (container.header.signature.decode(), out_file))
 
+def gen_layout (comp_list):
+	# prepare the layout from individual components from '-cl'
+	layout 		= "('PODS', 'PodsImage.bin', 'CLASSIC', 'RSA2048' , 'TestSigningPrivateKey.pem', 0x10, 0),\n"
```
mauricema: shouldn't the key be passed in through command line in the similar way as IasTool ?

mauricema: add the same for the final  output file name.  It should be from command line.

mauricema: I think when '-cl' is provided,  you also need to provide  key path and output file path.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -328,76 +362,133 @@ def dump (self):
...
+				offset = next_offset
 			if base_offset is None:
 				base_offset = offset
 			component.offset = offset - base_offset
 			offset += component.size
 
-		self.adjust_header ()
+		if is_mono_signing:
+			# for monolithic signing, set proper flags and update header
+			self.set_header_flags (CONTAINER_HDR._flags['MONO_SIGNING'])
```
myguitar: A little confused in set_header_flags (overwrite). So, this line does OR operation, right? Not set value.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -328,76 +362,133 @@ def dump (self):
...
-			alignment = int (alignment, 0)
+
+		# for monolithic signing, need to add a reserved _SG_ entry to hold the auth info
+		mono_sig = '_SG_'
+		is_mono_signing = True if layout[-1][0] == mono_sig else False
+
+		# get the first entry in layout as CONTAINER_HDR
+		container_sig, container_file, image_type, auth_type, key_file, alignment, region_size = layout[0]
+
+		if alignment == 0:
```
myguitar: Just in case, if alignment is not necessary, what's the value in alignment? 0 goes to 0x1000 all the time.

BootloaderCorePkg/Tools/GenContainer.py
stalamudupula: Addressed these in new patch.

BootloaderCorePkg/Tools/GenContainer.py
stalamudupula: If overwrite is set to 1, it is actually setting the flags again. hence the name.

BootloaderCorePkg/Tools/GenContainer.py
mauricema: alignment should be set to 1 if the alignment is not required.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -534,15 +679,18 @@ def main():
...
 	# Command for create
 	cmd_display = sub_parser.add_parser('create', help='create a container image')
-	cmd_display.add_argument('-l',  dest='layout',   type=str, required=True, help='Container layout intput file')
-	cmd_display.add_argument('-od', dest='out_dir',  type=str, default='.', help='Container output directory')
-	cmd_display.add_argument('-kd', dest='key_dir',  type=str, default='', help='Input key directory')
+	group = cmd_display.add_mutually_exclusive_group (required=True)
+	# '-l' or '-cl', one of them is mandatory
+	group.add_argument('-l',  dest='layout',   type=str, help='Container layout intput file if no -cl')
+	group.add_argument('-cl', dest='comp_list',nargs='+', help='List of Component files if no -l')
+	cmd_display.add_argument('-o', dest='out_dir',  type=str, default='.', help='Container output directory/file')
```
mauricema: change "out_dir " to "out_path"
change "key_dir " to "out_path"

PR #335, opened 2019-09-28
Add support for variable cell width in ConfigEditor

mauricema:
Current ConfigEditor only supports UINT8 format cell in table.
This patch added support for variable cell width including UINT8,
UINT16, UINT32 in table widget. Test configuration items were
also added in QEMU to test these format.

### DIFF
--- a/Platform/QemuBoardPkg/CfgData/CfgDataDef.dsc
+++ b/Platform/QemuBoardPkg/CfgData/CfgDataDef.dsc
@@ -141,10 +141,25 @@
   # !BSF HELP:{Silicon Test 1}
   gCfgData.SiliconTest1           |      * | 0x04 | 0x11223347
 
-  # !BSF NAME:{Silicon Test 9}
-  # !BSF TYPE:{EditNum, HEX, (0x00000000,0xFFFFFFFF)}
-  # !BSF HELP:{Silicon Test 9}
-  gCfgData.SiliconTest2           |      * | 0x04 | 0x11223348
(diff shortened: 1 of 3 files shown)


Platform/QemuBoardPkg/CfgData/CfgDataDef.dsc
```diff
@@ -141,10 +141,25 @@
...
 
-  # !BSF NAME:{Silicon Test 9}
-  # !BSF TYPE:{EditNum, HEX, (0x00000000,0xFFFFFFFF)}
-  # !BSF HELP:{Silicon Test 9}
-  gCfgData.SiliconTest2           |      * | 0x04 | 0x11223348
+  # !BSF NAME:{Silicon Test 2}
+  # !BSF TYPE:{Table}
+  # !BSF OPTION:{ 0:1:HEX, 1:1:HEX, 2:1:HEX, 3:1:HEX}
+  # !BSF HELP:{Silicon Test 2 to show BYTE table configuration}
+  gCfgData.SiliconTest2           |      * | 0x04 | 0x04030201
```
myguitar: 0x04030201 is test purpose? It should be { 0x01, 0x02, 0x03, 0x04 }. Support both way?

mauricema: Yes, it is for test.  It supports both array format {0x01, 0x02, 0x03, 0x04}, and  DWORD format 0x04030201

PR #342, opened 2019-10-02
Add image type param for GenContainer.py

stalamudupula:
Container type can be input from command line
from a list of [NORMAL, CLASSIC, MULTIBOOT] while
generating a container using GenContainer.py.
Setting default as NORMAL.

### DIFF
--- a/BootloaderCorePkg/Tools/GenContainer.py
+++ b/BootloaderCorePkg/Tools/GenContainer.py
@@ -590,11 +590,11 @@ def gen_container_bin (container_list, out_dir, inp_dir, key_dir = '.', tool_dir
 		out_file = container.create (each)
 		print ("Container '%s' was created successfully at:  \n  %s" % (container.header.signature.decode(), out_file))
 
-def gen_layout (comp_list, out_file, key_file):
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -590,11 +590,13 @@ def gen_container_bin (container_list, out_dir, inp_dir, key_dir = '.', tool_dir
...
 
-def gen_layout (comp_list, out_file, key_file):
+def gen_layout (comp_list, img_type, out_file, key_file):
 	# prepare the layout from individual components from '-cl'
-	if key_file == '':
-		key_file = 'TestSigningPrivateKey.pem'
-	layout 		= "('BOOT', '%s', 'CLASSIC', 'RSA2048' , '%s', 0x10, 0),\n" % (out_file, key_file)
+	if img_type not in CONTAINER_HDR._image_type.keys():
+		raise Exception ("Invalid Container Type '%s' !" % img_type)
+	if img_type == '':
```
mauricema: I don't think this condition will ever meet.  Line 595 will make sure img_type is not ''

PR #352, opened 2019-10-05
Implement USB de-initialization flow

### LINKED ISSUE #351: SBL will assert while trying to boot next USB option after enabling USB KB support
To reproduce it on APL,  just enable USB KB by changing following in BoardConfig.py
self.CONSOLE_IN_DEVICE_MASK  = 0x00000003

After it is done,  if USB is not the 1st boot option,  an assertion will be seen.

mauricema:
USB sub-system will have host controller scheduling frames on its
own once it is initialized and enabled. Leaving it running while
payload restarting or OS booting could potentially cause memory
corruption since the DMA might still be running on the background
targeting to previously allocated memory. The safer approach is
to stop the USB controller.

It also fixed #351.


BootloaderCommonPkg/Library/UsbInitLib/UsbInitLib.c
```diff
@@ -37,12 +26,42 @@ RegisterUsbDevice (
   IN PEI_USB_IO_PPI   *UsbIoPpi
   )
 {
-  if (mUsbIoCount < ARRAY_SIZE (mUsbIoArray)) {
-    mUsbIoArray[mUsbIoCount++] = UsbIoPpi;
+  if (mUsbInit.UsbIoCount < ARRAY_SIZE (mUsbInit.UsbIoArray)) {
+    mUsbInit.UsbIoArray[mUsbInit.UsbIoCount++] = UsbIoPpi;
   }
```
myguitar: Better to handle error condition as well.

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -552,6 +552,9 @@ BeforeOSJump (
   }
```
myguitar: pls put Payload changes in a separate commit.

BootloaderCommonPkg/Library/UsbInitLib/UsbInitLib.c
mauricema: sure, will add it.

PayloadPkg/OsLoader/OsLoader.c
mauricema: will do it.

mauricema:
pushed a new commit per comments.

PR #353, opened 2019-10-05
Skip IAS image verification if VERIFIED_BOOT is disabled

mauricema:
Current SBL code will assert if HAVE_VERIFIED_BOOT is 0. This patch
added check for PcdVerifiedBootEnabled to decide if IAS verification
is required.


jgutbub:
Looks good to me.

PR #361, opened 2019-10-08
Support 64bit EXT FS & non-512 block sizes

jgutbub:
EXT2/3 library has some limiatation to support
hardware block sizes larger than 512 (e.g. 4KB)
and also does not currently support the flag
INCOMPAT_64BIT which indicates larger group
descriptor sizes. This patch adds flexibility
to support 512 and 4KB block sizes as well as
64bit EXT file systems.

### DIFF
--- a/BootloaderCommonPkg/Library/Ext23Lib/Ext2Fs.c
+++ b/BootloaderCommonPkg/Library/Ext23Lib/Ext2Fs.c
@@ -224,7 +224,7 @@ BDevStrategy (
   if (ReadWrite != F_READ && ReadWrite != F_WRITE) {
     return EFI_INVALID_PARAMETER;
   }
-  if ((Size % BDEV_BLOCKSIZE) != 0) {
+  if ((Size % PrivateData->BlockSize) != 0) {
     return EFI_INVALID_PARAMETER;
   }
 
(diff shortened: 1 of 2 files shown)


BootloaderCommonPkg/Library/Ext23Lib/Ext2Fs.c
```diff
@@ -277,12 +277,23 @@ ReadInode (
...
   EXTFS_DINODE *DInodePtr;
 
   Fp = (FILE *)File->FileSystemSpecificData;
   FileSystem = Fp->SuperBlockPtr;
 
-  InodeSector = FSBTODB (FileSystem, INODETOFSBA (FileSystem, INumber));
+  Ext2FsGrpDes = FileSystem->Ext2FsGrpDes;
+  Ext2FsGrpDes = (EXT2GD*)((UINT32)Ext2FsGrpDes + (INOTOCG(FileSystem, INumber) * FileSystem->Ext2FsGDSize));
+
+  InodeSector = Ext2FsGrpDes->Ext2BGDInodeTables + (((INumber - 1) % FileSystem->Ext2Fs.Ext2FsINodesPerGroup) / FileSystem->Ext2FsInodesPerBlock);
```
mauricema: better to use ModU64x32 and DivU64x32 instead.  It will also help check  divided by 0 issue.

jgutbub: Updated it as advised. I might need to do another patch to clean up the types/sizes being used to truly make it 64bit and follow edk2 styles.

PR #364, opened 2019-10-09
Clean up EXT library

jgutbub:
The EXT library has some unused code that
we can remove to help reduce the size and
to clean things up some more. Also add a
routine for dumping the group descriptor
table which can be helpful for debugging
issues.


jgutbub:
This is a first patch to clean up some things but more work is needed, did not want to do too many things in one review, mostly it removes some routines we probably will not ever use, uses AsciiStrLen routine instead of local StrLen routine, and removes the endianness things since x86 is always little endian.

PR #374, opened 2019-10-10
[WHL] Fix firmware update failure

raghavag:
Current code set payload id depending on the gpio settings
and user selection from configuration data. When UEFI payload is
selected using GPIO or config data, payload id is being set to
UEFI irrespective of boot mode, which cause notification function
to get called, this locks the spi which inturn fails firmware update

Modified code to set payload id only in non-firmware update boot mode.

### DIFF
--- a/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -629,6 +629,10 @@ UpdatePayloadId (
   UINT32              PayloadId;
   UINT32              GpioLevel;
 
+  if (GetBootMode() == BOOT_ON_FLASH_UPDATE) {
+    return;
+  }
+
   PayloadId = 0;
   GenericCfgData = (GEN_CFG_DATA *)FindConfigDataByTag (CDATA_GEN_TAG);
   if (GenericCfgData != NULL) {
--- a/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -91,7 +91,6 @@
   ((UINT32) (((A).PinNum) | (((A).Group) << 16)))
 
 FVID_TABLE                  *mFvidPointer    = NULL;
-SILICON_CFG_DATA            *mSiliconCfgData = NULL;
 ///
 /// Overcurrent pins, the values match the setting of EDS, please refer to EDS for more details
 ///
@@ -617,7 +616,6 @@ PatchCpuSsdtTable (
   }
 }
 
-
 /**
   Update bootloader reserved region.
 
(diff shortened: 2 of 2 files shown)


Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -629,6 +629,10 @@ UpdatePayloadId (
   UINT32              PayloadId;
   UINT32              GpioLevel;
 
+  if (GetLoaderGlobalDataPointer ()->BootMode == BOOT_ON_FLASH_UPDATE) {
```
myguitar: Please use GetBootMode ()

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1002,6 +1002,62 @@ GpioInit (
...
+  )
+{
+  EFI_STATUS      Status;
+  GEN_CFG_DATA    *GenericCfgData;
+  SILICON_CFG_DATA  *SiliconCfgData;
+  UINT32          PayloadSelGpioData;
+  UINT32          PayloadSelGpioPad;
+  UINT32          PayloadId;
+
+  if (GetLoaderGlobalDataPointer ()->BootMode == BOOT_ON_FLASH_UPDATE) {
```
myguitar: Same here - GetBootMode

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1036,42 +1089,9 @@ BoardInit (
...
-            PayloadId = UEFI_PAYLOAD_ID_SIGNATURE;
-          }
-        }
-        DEBUG ((DEBUG_INFO, "Set PayloadId to 0x%08X based on GPIO config\n", PayloadId));
-      }
-    }
-    mSiliconCfgData = SiliconCfgData;
-
-    SetPayloadId (PayloadId);
+    mSiliconCfgData = (SILICON_CFG_DATA *)FindConfigDataByTag (CDATA_SILICON_TAG);
```
myguitar: mSiliconCfgData is required? Even if current stage is Stage2, but better to avoid global variable.  UpdatePayloadId () also does not use mSiliconCfgData.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1036,42 +1089,9 @@ BoardInit (
...
 
     if (GetLoaderGlobalDataPointer ()->BootMode != BOOT_ON_FLASH_UPDATE) {
```
myguitar: Can you change it as well to use GetBootMode()?

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1036,42 +1089,9 @@ BoardInit (
...
       UpdateBlRsvdRegion ();
```
myguitar: UpdatePayloadId() can be put here before UpdateBlEsvdRegion().

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: agree.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Done

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Done

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Done, it is used by platformupdateacpitable, not sure why it is used that way, so left untouched. but removed now

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: Done

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
raghavag: we thought it is cleaner and consistent across platforms to check boot mode inside UpdatePayloadId function, but will move it here

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -2012,8 +2025,10 @@ PlatformUpdateAcpiTable (
   Ptr  = (UINT8 *)Table;
   End  = (UINT8 *)Table + Table->Length;
 
+  SiliconCfgData = (SILICON_CFG_DATA *)FindConfigDataByTag (CDATA_SILICON_TAG);
```
mauricema: Better to move it to after line #2030.  So that it will only be called if ECBR table exists.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
myguitar: It looks not changed.

raghavag: Done now

PR #380, opened 2019-10-14
Allow platform to degrade eMMC from HS400 to HS200

mauricema:
This patch allows platform to degrade eMMC HS400 to HS200 using
static configuration. To do this, please add the following into
BoardConfig.py:
  self.ENABLE_EMMC_HS400 = 0
This is useful when platform has hardware issue to run at eMMC
HS400 mode.

### DIFF
--- a/BootloaderCommonPkg/BootloaderCommonPkg.dec
+++ b/BootloaderCommonPkg/BootloaderCommonPkg.dec
@@ -200,8 +200,9 @@
 
 
 [PcdsFeatureFlag]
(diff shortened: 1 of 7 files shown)


mauricema:
Please do not merge yet,  need to collect more feedbacks.

mschuetterle:
The implement is fine with me. I just thought it's a good idea to explicitly mention the reason for this change: APLI-07 errata. But this is common code and not only for Apollo Lake. So, leave it up to you.

mauricema:
I'll keep it generic for now.

PR #382, opened 2019-10-14
[CFL] Change default payload sel based on GPIO

raghavag:
This patch changes the default behavior of gpio pin for payload
selection, currently user configured GPIO pin being high boots
to uefi payload, modified code will boot to OS loader when GPIO pin
is high.


mauricema:
looks good

PR #385, opened 2019-10-16
Add change current boot option to boot cmd

jgutbub:
The 'boot' command has been updated to add
an option to update the current boot option
to be booted next, e.g. mCurrentBoot. To
change the current boot option just need to
run 'boot' and then select 'c' and the
index to change the current boot option to.

### DIFF
--- a/BootloaderCommonPkg/Library/ShellLib/CmdBoot.c
+++ b/BootloaderCommonPkg/Library/ShellLib/CmdBoot.c
@@ -17,6 +17,7 @@
 #include <Guid/OsBootOptionGuid.h>
 #include "Shell.h"
 
+extern UINT8    mCurrentBoot;
 
 /**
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/ShellLib/CmdBoot.c
```diff
@@ -17,6 +17,7 @@
 #include <Guid/OsBootOptionGuid.h>
 #include "Shell.h"
 
+extern UINT8    mCurrentBoot;
```
jgutbub: OS Loader uses this value to track which boot option to try next. We can modify it here in the 'boot' command if we want to change the next option.

https://github.com/slimbootloader/slimbootloader/blob/master/PayloadPkg/OsLoader/OsLoader.c#L911

PR #388, opened 2019-10-16
[OsLoader] Clean-up PreOsChecker

myguitar:
Split PreOsChecker in a separate file and add minimum APIs
- IsPreOsCheckerLoaded ()
- LoadPreOsChecker ()
- StartPreOsChecker ()

### DIFF
--- a/PayloadPkg/OsLoader/PreOsChecker.c
+++ b/PayloadPkg/OsLoader/PreOsChecker.c
@@ -1,0 +1,152 @@
+/** @file
+
+  Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Library/DebugLib.h>
+#include <Library/ContainerLib.h>
+#include <Library/ElfLib.h>
+#include <Library/LinuxLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/BootloaderCommonLib.h>
+
+#define SIGNATURE_IPFW    SIGNATURE_32 ('I', 'P', 'F', 'W')
+#define SIGNATURE_POSC    SIGNATURE_32 ('P', 'O', 'S', 'C')
+
+typedef struct {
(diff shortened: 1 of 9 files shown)


PayloadPkg/OsLoader/PreOsChecker.c
```diff
@@ -0,0 +1,149 @@
...
+
+  PreOsParams.OsBootState.Esi     = (UINT32) OsBootParam;
+  PreOsParams.OsBootState.Eip     = OsBootParam->Hdr.Code32Start;
+  PreOsParams.OsBootState.Eflags  = 0;
+
+  EntryPoint = (PRE_OS_CHECKER_ENTRY)mPreOsCheckerEntry;
+  EntryPoint (&PreOsParams);
+
+  //
+  // Not reach here
```
jgutbub: Should we add a CPU dead loop or ASSERT here for debug?

myguitar: Updated in coming push. Thanks.

PayloadPkg/OsLoader/PreOsChecker.c
```diff
@@ -0,0 +1,152 @@
...
+  //  DEBUG ((DEBUG_INFO, "Pre-OS checker is already loaded at entry @ 0x%p\n", mPreOsCheckerEntry));
+  //  return EFI_STATUS;
+  //}
+
+  //
+  // Check if there is a pre-OS checker that needs to be executed
+  //
+  Buffer = NULL;
+  Length = 0;
+  Status = LoadComponent (SIGNATURE_IPFW, SIGNATURE_POSC, &Buffer, &Length);
```
mauricema: Do we have a way to indicate if PreOsChecker support is required or not ?   
If a platform does not need PreOsChecker,  maybe we should not try to load the image at all.

myguitar: Updated in coming push. Token space has been updated. Will follow up on-going boards which will be impacted by this change once this is merged.

PR #396, opened 2019-10-17
[OsLoader] OsLoader lifecycle change

myguitar:
This patch allows OsLoader to handle all Boot Options without restarting.

Currently, OsLoader restarts to handle next Boot Option from the beginning.
This behavior does not preserve allocated memories from Payload heap,
so it makes hardware/software state hard to maintain.

### DIFF
--- a/PayloadPkg/OsLoader/LoadImage.c
+++ b/PayloadPkg/OsLoader/LoadImage.c
@@ -7,80 +7,24 @@
 
 #include "OsLoader.h"
 
-CONST CHAR16  *mConfigFileName[2] = {
+#define LOADED_IMAGES_INFO_SIGNATURE   SIGNATURE_32 ('L', 'I', 'I', 'S')
+
+typedef struct {
+  UINTN                   Signature;
+  LOADED_IMAGE           *LoadedImageList[LoadImageTypeMax];
+} LOADED_IMAGES_INFO;
+
+STATIC CONST CHAR16  *mConfigFileName[2] = {
   L"config.cfg",
   L"boot/grub/grub.cfg"
 };
 
 /**
-  Get hardware partition handle from boot option info
+  Get Boot image from raw partition
 
-  This function will initialize boot device, and get hardware partition
-  handle based on boot option.
-
-  @param[in]  BootOption      Current boot option
-  @param[out] HwPartHandle    Hardware partition handle for boot image
-
-  @retval  RETURN_SUCCESS     If partition was found successfully
-  @retval  Others             If partition was not found
-**/
-EFI_STATUS
-FindBootPartition (
-  IN  OS_BOOT_OPTION         *BootOption,
-  OUT EFI_HANDLE             *HwPartHandle
-  )
-{
-  RETURN_STATUS              Status;
-  UINTN                      BootMediumPciBase;
-
-  AddMeasurePoint (0x4040);
-  //
-  // Get OS boot device address
-  //
-  BootMediumPciBase = GetDeviceAddr (BootOption->DevType, BootOption->DevInstance);
-  DEBUG ((DEBUG_INFO, "BootMediumPciBase(0x%x)\n", BootMediumPciBase));
-  BootMediumPciBase = TO_MM_PCI_ADDRESS (BootMediumPciBase);
-
-  //
-  // Init Boot device functions
-  //
-  Status = MediaSetInterfaceType (BootOption->DevType);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "Failed to set media interface - %r\n", Status));
-    return RETURN_UNSUPPORTED;
-  }
-
-  DEBUG ((DEBUG_INFO, "Getting boot image from %a\n", GetBootDeviceNameString(BootOption->DevType)));
-  Status = MediaInitialize (BootMediumPciBase, DevInitAll);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "Failed to init media - %r\n", Status));
-    return Status;
-  }
-
-  AddMeasurePoint (0x4050);
-  MediaTuning (BootMediumPciBase);
-  AddMeasurePoint (0x4055);
-
-  DEBUG ((DEBUG_INFO, "Try to find boot partition\n"));
-  Status = FindPartitions (BootOption->HwPart, HwPartHandle);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "Failed to find partition - %r\n", Status));
-    ClosePartitions (*HwPartHandle);
-    return Status;
-  }
-  AddMeasurePoint (0x4060);
-
-  DEBUG ((DEBUG_INFO, "Find partition success\n"));
-  return RETURN_SUCCESS;
-}
-
-
-/**
-  Get IAS image from raw partition
-
-  Using boot option info, this function will read IAS image from raw
+  Using boot option info, this function will read Boot image from raw
   partition based on hardware partition info saved in LoadedImage.
-  After IAS image is loaded into memory, its information will be saved
+  After Boot image is loaded into memory, its information will be saved
   to LoadedImage.
 
   @param[in]      BootOption      Current boot option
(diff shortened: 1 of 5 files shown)


BootloaderCommonPkg/Library/ShellLib/CmdBoot.c
```diff
@@ -369,9 +367,6 @@ PrintBootOption (
                  );
     }
 
-    if (Index == mCurrentBoot) {
-      ShellPrint (L" *Current");
-    }
```
mauricema: I think it is good to print out a marker for current boot option. Why do we remove this ?

BootloaderCommonPkg/Library/ShellLib/CmdBoot.c
```diff
@@ -503,7 +498,6 @@ ShellCommandBootFunc (
       ShellPrint (L"Enter index to change to (0 to %u)\n",
                   BootOptionList->OsBootOptionCount - 1
                   );
-      ShellPrint (L"(current index %u) ", mCurrentBoot);
```
mauricema: this print is useful. why remove ?

jgutbub: This print is kind of in the spirit of how fdisk/gdisk will print the default values that you can just hit enter on to leave it as the default value. Maybe the text was not clear as "(current index)" but I agree it would be nice to have something display the current value in case the user did not really want to do 'c' option after realizing the current value.

jgutbub: I was even thinking maybe we could add this to the OS Loader boot option printing, it made it easier to know which boot option would be tried next.

BootloaderCommonPkg/Library/ShellLib/CmdBoot.c
jgutbub: I was even thinking maybe we could add this to the OS Loader boot option printing, it made it easier to know which boot option would be tried next.

myguitar: I agree this is good to have. Next PRs would handle this. Let me keep the mCurrentBoot variable in this commit as a initial patch. Eventually, this global variable needs to be removed and we need to improve GetCurrentBootOption() and GetNextBootOption() more intuitive (maybe adding SetCurrentBootOption as well).

BootloaderCommonPkg/Library/ShellLib/CmdBoot.c
myguitar: Same as above. Will be kept in coming push.

myguitar:
Main changes are in BootOsImage ().
1. InitBootDevice
2. FindBootPartitions
3. InitBootFileSystem
4. LoadBootImage
5. ParseBootImage
6. SetupBootImage
7. UpdateBootParameters
8. StartBooting
and error handling.

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -628,13 +575,233 @@ StartBooting (
...
+    *FsHandle = NULL;
+
+    //
+    // Get boot image A/B info - Supported in RAW partition only.
+    //
+    BootSlot = GetBootSlot (OsBootOption, HwPartHandle);
+    DEBUG ((DEBUG_INFO, "BootSlot = 0x%x\n", BootSlot));
+    if (BootSlot == 1) {
+      OsBootOption->BootFlags |= LOAD_IMAGE_FROM_BACKUP;
+    }
```
gdong1: A/B support could be used for file system. e.g. Current TGL and EHL could boot Yocto (using Mender) from A or B partition.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -527,63 +463,105 @@ GetTraditionalLinux (
...
+      LoadedImage[Index].LoadImageType = LOAD_IMAGE_EXTRA0 + Index;
+    }
+  } else {
+    LoadedImage = (LOADED_IMAGE *)AllocateZeroPool (sizeof (LOADED_IMAGE));
+    LoadedImage->HwPartHandle = HwPartHandle;
+    if (BootFlags & BOOT_FLAGS_TRUSTY) {
+      LoadedImage->LoadImageType = LOAD_IMAGE_TRUSTY;
+    } else {
+      LoadedImage->LoadImageType = LOAD_IMAGE_NORMAL;
+    }
```
gdong1: Looks the logic is not same with original one.
By original logic, in one boot option there could be 7 images (misc image, normal image, trusty image, 4 extra images) loaded at same boot. could we still support it?

PayloadPkg/OsLoader/OsLoader.c
myguitar: Got it. Let me update it in next commit.

PayloadPkg/OsLoader/LoadImage.c
myguitar: It's not yet. Let me update logic to load all Boot Images in next commit.

myguitar:
Updated according to review comment. Verified USB/eMMC boot on APL with Linux image and ACRN image.
@mschuetterle Can you also review this by any chance?

PayloadPkg/OsLoader/LoadImage.c
myguitar: The same logic has been updated in the latest push.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -527,63 +474,198 @@ GetTraditionalLinux (
...
-  // Enable tradition linux boot only in debug build
-  GuidHob = GetNextGuidHob (&gBootLoaderVersionGuid, GetHobListPtr ());
-  if (GuidHob != NULL) {
-    VersionHob = (EXT_BOOT_LOADER_VERSION *)GET_GUID_HOB_DATA (GuidHob);
-    if ((VersionHob->Version.ImageVersion.BldDebug != 0) ||
-         !FeaturePcdGet (PcdVerifiedBootEnabled)) {
+  //
+  // Launch Traditional Linux for debugging purpose only
+  //
+  if ((DebugCodeEnabled () == TRUE) || !FeaturePcdGet (PcdVerifiedBootEnabled)) {
```
gdong1: we don't need compare 2 BOOLEAN by removing "== TRUE"

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -628,13 +575,303 @@ StartBooting (
...
+  LOADED_IMAGE        *LoadedImage;
+  LOADED_IMAGE        *LoadedTrustyImage;
+  EFI_STATUS           Status;
+
+  GetLoadedImageByType (LoadedImageHandle, LoadImageTypeNormal, &LoadedImage);
+  GetLoadedImageByType (LoadedImageHandle, LoadImageTypeTrusty, &LoadedTrustyImage);
+
+  if (LoadedImage == NULL && LoadedTrustyImage == NULL) {
+    return EFI_LOAD_ERROR;
+  }
```
gdong1: Normal OS image is always required, and trusty image is optional. so just check LoadedImage should be fine if we want to check.

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -644,78 +881,104 @@ BootOsImage (
...
-    Status = StartBooting (&LoadedTrustyImage);
-  } else {
-    Status = StartBooting (&LoadedImage);
+  //
+  // Update Boot Parameters
+  //
+  Status = UpdateBootParameters (OsBootOption, LoadedImageHandle);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_INFO, "Failed to Update Boot Parameters - Status = %r\n", Status));
+    goto ERROR;
```
gdong1: Looks we are trying to free all the memory if this boot option boots failure. Inside UpdateBootParameters () I believe there are several places to allocate big memory. we could free them later.

PayloadPkg/OsLoader/LoadImage.c
myguitar: Let me remove it.

PayloadPkg/OsLoader/OsLoader.c
myguitar: Right, let me update this.

PayloadPkg/OsLoader/OsLoader.c
myguitar: Right. This PR is to make general flow in top interface layer. After this PR, we need to improve all memory management parts in sub layers.

PayloadPkg/OsLoader/OsLoader.c
myguitar: Updated as commented.

myguitar:
Updated according to review comments. Verified booting Linux from USB and ACRN from eMMC.

PR #405, opened 2019-10-17
[APL] Add config option for XDCI mode

### LINKED ISSUE #397: [APL] Add USB OTG enable/disable to config templates?
I think the configuration templates have everything useful exposed for enabling / disabling SoC peripherals, which is great. The only thing which seems to be missing for APL is something to set `UsbOtg` on the FSP settings. I noticed that the USB xDCI was enabled by default with SBL, which is something that I wasn't expecting.

myguitar:
This patch allows a board to configure XDCI mode.
By default, keep XDCI mode as PCI mode which is default value in FSP.
The mode can be controlled by Config Data.
- Silicon Setting > Device Enable/Disable > Control 2 > Enable XDCI

If XDCI mode needs to be overrided, put new value in .dlt(delta) file.
- DEV_EN_CFG_DATA.DevEnControl2.XdciEnable | 0x0 (Disabled)
- DEV_EN_CFG_DATA.DevEnControl2.XdciEnable | 0x2 (ACPI mode)

It will fix #397 with the override value in a delta file.


umaplehurst:
Looks great. Tested on this side and all is good.

PR #407, opened 2019-10-18
Silicon init with eMMC HS400 mode config

### LINKED ISSUE #406: ENABLE_EMMC_HS400 needs to also cascade into FSP settings
If `BoardConfig.py` has `self.ENABLE_EMMC_HS400 = 0` it looks like this will change SBL behavior, and SBL will not use HS400, but once the OS (or UEFI payload) boots, then the OS will still go into HS400. For the OS to also be limited to HS200, the setting needs to be cascaded into FSP (`FspsConfig->eMMCHostMaxSpeed`).

myguitar:
This patch is a follow-up patch of #380. If a platform disables HS400 mode,
silicon init code will configure eMMC in HS200. Other modes not in scope.

This will also fix #406.


myguitar:
> Thank you for adding FSP UPD update for it. I missed this part originally.

No problem. Let me wait for issue submitter's feedback.

umaplehurst:
Looks good to me, many thanks!

myguitar:
> Looks good to me, many thanks!

You are welcome. I hope you enjoy Slim Bootloader and look forward to your more comment/feedback/contribution.

Issue #393, opened 2019-10-16
[APL] Can't stitch image when OEM public key is not present


umaplehurst:
I get the following error when giving `StitchLoader.py` an input image which does not have an OEM public key:

```
Patching Slim Bootloader Flash Map table ...
Traceback (most recent call last):
  File "Platform/ApollolakeBoardPkg/Script/StitchLoader.py", line 648, in <module>
    ret = create_ifwi_image (args.ifwi_in, args.ifwi_out, args.bios_out, args.plat_data, args.non_redundant, stitch_dir)
  File "Platform/ApollolakeBoardPkg/Script/StitchLoader.py", line 551, in create_ifwi_image
    patch_flash_map (ifwi_data, platform_data)
  File "Platform/ApollolakeBoardPkg/Script/StitchLoader.py", line 429, in patch_flash_map
    raise Exception("Cannot locate component '%s' in BPDT !" % path)
Exception: Cannot locate component 'IFWI/BIOS/BP0/BPDT/BpdtCseBup/oem.key' in BPDT !
```

My workaround is to comment out the `OEMKEY.bin` entry from `BoardConfig.py`. Given that at the moment things only work anyway with Boot Guard disabled, according to the documentation, this should be a non-fatal issue? Or is the OEM key used for any other important things? FIT seems to be happy enough creating an image that contains no OEM key.

sagraw2:
OEMKey manifest is required if OEM wants any of these things:
- enable BootGuard
- use OEM Debug unlock tokens
- replace any of the Intel provided binaries (such as iUni, aDSP, ISH)
If none of the above is used, there is no need to bundle OEMKey manifest within stitched firmware image.

Let me investigate a little bit more on why StitchLoader.py is mandating the presence of it and will update here later.

rprangar:
Fixed.

PR #408, opened 2019-10-18
OemKm and Bpm are not a MUST

sagraw2:
This patch fixes the issued identified in #393

### DIFF
--- a/Platform/ApollolakeBoardPkg/Script/StitchLoader.py
+++ b/Platform/ApollolakeBoardPkg/Script/StitchLoader.py
@@ -371,7 +371,6 @@ def patch_flash_map (image_data, platform_data = 0xffffffff):
       b'SG1B' : "IFWI/BIOS/BP0/BPDT/BpdtIbb/IBB",
       b'SG1A' : "IFWI/BIOS/BP0/BPDT/BpdtIbb/IBBL",
       b'_BPM' : "IFWI/BIOS/BP0/BPDT/BpdtIbb/BPM.met",
-      b'OEMK' : "IFWI/BIOS/BP0/BPDT/BpdtCseBup/oem.key",
     }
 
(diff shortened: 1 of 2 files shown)


Platform/ApollolakeBoardPkg/Script/StitchLoader.py
```diff
@@ -426,7 +426,9 @@ def patch_flash_map (image_data, platform_data = 0xffffffff):
 
             comp  = IFWI_PARSER.locate_component (ifwi, path)
             if not comp:
-                raise Exception("Cannot locate component '%s' in BPDT !" % path)
+                if desc.sig not in [b'_BPM', b'OEMK']:
```
mauricema: Since current SBL code still use _BPM,  should we leave it ?
Also it there a real case that no BPM exists base IFWI ?

myguitar:
Can you rebase your branch instead of merge? It's a single commit actually, but looks like 3 commits.

Platform/ApollolakeBoardPkg/Script/StitchLoader.py
sagraw2: BPM is used only for BtG purposes.
And, I believe the existing build infrastructure always creates BPM and therefore we see it in all IFWI's.

In the BtG build scenario, this part of the script will be executed after KM and BPM is already built in IFWI and therefore 'BPM' and 'KM' should be there in IFWI image.

mauricema: If BPM will always be there,   "if desc.sig not in [b'OEMK']  should be good enough, is it right ?

myguitar: Agree. I think checking OEMK only would be enough.

PR #410, opened 2019-10-21
[CFL] TPM selection type support to StitchIfwi

lsubash:
Added a option in CFL StitchIfwi script to pass the TPM type to
be selected for Intel FIT tool stitching.

[TEST] = Run StitchIfwi.py with -tpm=dtpm for WHL
         and verify use of dTPM.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
@@ -77,7 +77,7 @@ Please follow the steps below for WHL/CFL IFWI stitching.
      EX:
        Assuming stitching workspace is at D:\Stitch and building ifwi for WHL platform
        To stitch IFWI with SPI QUAD mode and Boot Guard profile VM:
-         python  Platform\CoffeelakeBoardPkg\Script\StitchIfwi.py -b vm -q -a whl -w D:\Stitch -s Stitch_Components.zip
+         python  Platform\CoffeelakeBoardPkg\Script\StitchIfwi.py -b vm -q -a whl -w D:\Stitch -s Stitch_Components.zip -t ptt
        To clean all generated files:
          python  Platform\CoffeelakeBoardPkg\Script\StitchIfwi.py -c -w D:\Stitch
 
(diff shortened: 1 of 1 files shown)


Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -368,6 +368,52 @@ def sign_binary(infile, stitch_dir, cfg_var):
     hash = bytearray(open('output/input/pubkey'+'.hash','rb').read())
     print(''.join('%02X ' % b for b in hash))
 
+def update_tpm_type(tpm, tree):
```
mauricema: I think we can use python dictionary to optimize the code below to make it easier for maintenance.

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -368,6 +368,52 @@ def sign_binary(infile, stitch_dir, cfg_var):
...
+
+  # Enable ptt/dtpm
+  if tpm == "dtpm":
+      #PttConfiguration
+      node = tree.find('./PlatformProtection/IntelPttConfiguration/PttSupported')
```
myguitar: Also, file path can be generalized with os.path.join().

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
lsubash: Will do

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
lsubash: This is an xml entry. I would update with dict methods.

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -368,6 +368,42 @@ def sign_binary(infile, stitch_dir, cfg_var):
     hash = bytearray(open('output/input/pubkey'+'.hash','rb').read())
     print(''.join('%02X ' % b for b in hash))
 
+def update_tpm_type(tpm_type, tree):
```
sagraw2: Since we are now using this new method, plese remove the code in Line 408 and around which is already updating xml for TPM Type.

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -616,8 +660,12 @@ def main():
         print ("Cleaning completed successfully !\n")
         return 0
 
+    if args.btg_profile in ["vm","fvme"] and args.tpm == "none":
+        print "ERROR: Choose appropriate Tpm type for BootGuard profile 3 and 5"
+        return 0
+
     print ("Executing stitch.......")
-    if stitch (stitch_dir, args.stitch_zip, args.btg_profile, args.quad_mode, args.plat_data, args.platform):
+    if stitch (stitch_dir, args.stitch_zip, args.btg_profile, args.quad_mode, args.plat_data, args.platform, args.tpm):
```
sagraw2: There is a example line in the beginning of the script. Can we update that too since we are adding this new argument for Tpm Type?

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -605,6 +643,12 @@ def main():
                     default=None,
                     help='Specify a platform specific data (HEX, DWORD) for customization')
 
+    ap.add_argument('-tpm',
```
myguitar: As other arguments, better to use '-t', '--tpm' '-t' if no one is using it.

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
lsubash: Yes. will clean up the code.

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
lsubash: Yeah, Will use "-t" option.

Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
lsubash: will add.

PR #413, opened 2019-10-24
Set default Container type to CLASSIC

stalamudupula:
Container parse logic cannot handle NORMAL (default) case
currently. So, if -t option is not used while generating a
Container, unsupported image type is reported during runtime.

Change the default type to CLASSIC, and end-user can use
-t option to switch to MULTIBOOT if needed.


myguitar:
So, are you going to make a following up patch for normal image? What about implementing missing part for normal image in UpdateLoadedImage ()?

PR #414, opened 2019-10-24
Remove global variable in boot cmd

jgutbub:
The 'boot' command is making use of a global
variable declared in the OS Loader but it might
not always be available if the Shell lib is
used with other code. This patch removes the
dependency on the global variable.

Also fixes a small issue with not providing a
value at the prompt; instead prompt the user
with a error message when no valid value has
been provided.


jgutbub:
@myguitar I think your previous proposed changes to boot command included the removal of the global variable usage. Let me know if I missed anything.

PR #416, opened 2019-10-25
Update mem command and remove io command

jgutbub:
The 'mem' command is being updated to handle
PCI device formatted addresses and also to
support IO device interfacing. Removing the
'io' command to reduce code size since it
has been intergrated into the 'mem' command.


myguitar:
In my opinion, better to have separate cmds for mem, io and pci. There is still pci command, and using b.d.f would be better instead of pcie config space memory address. In terms of code size, we can include mandatory shell commands only for Release binary in LoadShellCommands by using DEBUG_CODE_START/END.

BootloaderCommonPkg/Library/ShellLib/CmdMem.c
```diff
@@ -50,115 +52,187 @@ ShellCommandMemFunc (
...
-    }
-
-    if ((Index % ItemsPerRow) != 0) {
-      ShellPrint (L"\n");
+      break;
     }
   }
 
-  return EFI_SUCCESS;
+  if (EFI_ERROR(Status)) {
```
jgutbub: I've updated to add some usage examples here.

mauricema:
> In my opinion, better to have separate cmds for mem, io and pci. There is still pci command, and using b.d.f would be better instead of pcie config space memory address. In terms of code size, we can include mandatory shell commands only for Release binary in LoadShellCommands by using DEBUG_CODE_START/END.

I think it makes sense to share the similar code flow for mem, io, pci cfg since all of them are very similar to each other.  Further, it is more close the UEFI Shell "mm" command which supports mmio, io, pci, etc.   I also agree we should keep PCI command. It can be used to list all PCI devices and also dump PCI cfg space.   When use "mem" to dump PCI space, the address is b.d.f, format, not pcie mmio address any more.

BootloaderCommonPkg/Library/ShellLib/CmdMem.c
```diff
@@ -50,115 +52,187 @@ ShellCommandMemFunc (
...
-      * (volatile UINT16 *)Addr = (UINT16)Value;
-      break;
-    case 4:
-      * (volatile UINT32 *)Addr = (UINT32)Value;
-      break;
-    case 8:
-      * (volatile UINT64 *)Addr = (UINT64)Value;
-      break;
-    default:
+    if (IsIoAddr && IsPciAddr) {
```
mauricema: Why do we check it inside the while loop?   Shouldn't we do it outside ?

BootloaderCommonPkg/Library/ShellLib/CmdMem.c
```diff
@@ -50,115 +52,187 @@ ShellCommandMemFunc (
...
   }
 
-  // Parse access width
-  if (Argc > 2) {
-    Width = StrHexToUintn (Argv[2]);
+  while (!EFI_ERROR (Status)) {
     if ((Width != 1) && (Width != 2) && (Width != 4) && (Width != 8)) {
-      ShellPrint (L"Error: Width must be 1, 2, 4, or 8\n");
-      return EFI_INVALID_PARAMETER;
+      ShellPrint (L"Error, width %u not supported\n", Width);
```
mauricema: Why do we check width inside the while loop?   Shouldn't we do it outside ?

BootloaderCommonPkg/Library/ShellLib/CmdMem.c
```diff
@@ -50,115 +52,187 @@ ShellCommandMemFunc (
...
   } else {
-    // Invalid
-    goto usage;
+    Status = EFI_INVALID_PARAMETER;
   }
 
-  // Parse access width
-  if (Argc > 2) {
-    Width = StrHexToUintn (Argv[2]);
+  while (!EFI_ERROR (Status)) {
```
mauricema: Do we really need a while loop here ?

BootloaderCommonPkg/Library/ShellLib/CmdMem.c
```diff
@@ -50,115 +52,187 @@ ShellCommandMemFunc (
...
 
-    for (Index = Start; Index < Count; Index++) {
-      if ((Index % ItemsPerRow) == 0) {
-        ShellPrint (L"%08x: ", Addr);
+    if (Write) {
+      for (Index = 0; Index < Count; Index++, Addr+=Width) {
+        switch (Width) {
+        case 1:
+          if (IsIoAddr == FALSE) {
+            * (volatile UINT8 *)Addr = (UINT8)Value;
```
mauricema: You can use  MmioWrite8/16/32 instead()

BootloaderCommonPkg/Library/ShellLib/CmdMem.c
```diff
@@ -50,115 +52,187 @@ ShellCommandMemFunc (
...
+    if ((Addr & (Width - 1)) != 0) {
+      ShellPrint (L"Error, unaligned access\n");
+      Status = EFI_INVALID_PARAMETER;
+      break;
     }
 
-    for (Index = Start; Index < Count; Index++) {
-      if ((Index % ItemsPerRow) == 0) {
-        ShellPrint (L"%08x: ", Addr);
+    if (Write) {
```
mauricema: Maybe better to break the code below into Read and Write sub-functions to make it easier to read.

BootloaderCommonPkg/Library/ShellLib/CmdMem.c
```diff
@@ -50,115 +52,187 @@ ShellCommandMemFunc (
...
+        switch (Width) {
+        case 1:
+          if (IsIoAddr == FALSE) {
+            * (volatile UINT8 *)Addr = (UINT8)Value;
+          } else {
+            IoWrite8 (Addr, (UINT8)Value);
+          }
+          break;
+        case 2:
+          if (IsIoAddr == FALSE) {
```
mauricema: BOOLEAN does not need to compare with TRUE or FALSE.  You can use if (IsIoAddr) instead.

myguitar:
> > In my opinion, better to have separate cmds for mem, io and pci. There is still pci command, and using b.d.f would be better instead of pcie config space memory address. In terms of code size, we can include mandatory shell commands only for Release binary in LoadShellCommands by using DEBUG_CODE_START/END.
> 
> I think it makes sense to share the similar code flow for mem, io, pci cfg since all of them are very similar to each other. Further, it is more close the UEFI Shell "mm" command which supports mmio, io, pci, etc. I also agree we should keep PCI command. It can be used to list all PCI devices and also dump PCI cfg space. When use "mem" to dump PCI space, the address is b.d.f, format, not pcie mmio address any more.

I think using "mem", "io", "pci" is more intuitive and "mm" needs one more parameter to specific "mem", "io", "pci". I don't have strong opinion about shell commands if you are okay.
What about changing command "mm" instead of "mem" with same format?
    MM Address [Width] [-[MEM | MMIO | IO | PCI | PCIE]] [:Value]
    1|2|4|8 supported byte widths - Default is 1 byte. 
    -MEM = Memory, 
    -MMIO = Memory Mapped IO, 
    -IO = in/out, 
    -PCI = PCI Config space (format: 000000ssbbddffrr)
    -PCIE = PCIE Config space (format: 000000ssbbddffrrr)
    Default access type is memory (MEM)

mauricema:
> > > In my opinion, better to have separate cmds for mem, io and pci. There is still pci command, and using b.d.f would be better instead of pcie config space memory address. In terms of code size, we can include mandatory shell commands only for Release binary in LoadShellCommands by using DEBUG_CODE_START/END.
> > 
> > 
> > I think it makes sense to share the similar code flow for mem, io, pci cfg since all of them are very similar to each other. Further, it is more close the UEFI Shell "mm" command which supports mmio, io, pci, etc. I also agree we should keep PCI command. It can be used to list all PCI devices and also dump PCI cfg space. When use "mem" to dump PCI space, the address is b.d.f, format, not pcie mmio address any more.
> 
> I think using "mem", "io", "pci" is more intuitive and "mm" needs one more parameter to specific "mem", "io", "pci". I don't have strong opinion about shell commands if you are okay.
> What about changing command "mm" instead of "mem" with same format?
> MM Address [Width] [-[MEM | MMIO | IO | PCI | PCIE]] [:Value]
> 1|2|4|8 supported byte widths - Default is 1 byte.
> -MEM = Memory,
> -MMIO = Memory Mapped IO,
> -IO = in/out,
> -PCI = PCI Config space (format: 000000ssbbddffrr)
> -PCIE = PCIE Config space (format: 000000ssbbddffrrr)
> Default access type is memory (MEM)

Sure,  let use rename "mem" to "mm" to align with UEFI Shell command  UEFI Shell command is like this:
mm address [value] [-w 1|2|4|8] [-MEM | -MMIO | -IO | -PCI | -PCIE] [-n] 
I think we still need "count" to indicate how many bytes to display for read operation.  "-c count" might be required.

jgutbub:
Someone else can port the mm command from EDK2.

PR #426, opened 2019-11-01
Fix comments to SBL core MemAllocLib to reflect code behavior

rprangar:


BootloaderCorePkg/Library/MemoryAllocationLib/MemoryAllocationLib.c
```diff
@@ -54,8 +54,9 @@ InternalUpdateMemPoolBottom (
   Allocates a buffer of type EfiBootServicesData.
 
   Allocates the number bytes specified by AllocationSize of type EfiBootServicesData and returns a
-  pointer to the allocated buffer.  If AllocationSize is 0, then a valid buffer of 0 size is
-  returned.  If there is not enough memory remaining to satisfy the request, then NULL is returned.
+  pointer to the allocated buffer.  If AllocationSize is 0, LdrGlobal->MemPoolCurrTop is returned.
+  If there is not enough memory remaining to satisfy the request, InternalUpdateMemPoolTop ASSERTS 
+  and the function does not return.
```
gdong1: yes. this is the current code behavior to ASSERT even in release build. 
Do we have plan to support to build ASSERT out in release mode? If yes, the function will return.

mauricema: We have no plan to remove the ASSERT in release build for SBL.

rprangar: I think this is a good feedback. I can change the assert to deadloop.

gdong1: OK. then it is fine.

PR #428, opened 2019-11-02
CmdFs: Support display console print in 'fs ls'

myguitar:
This allows 'fs ls' command to print directory or file lists to both serial
and display console according to CONSOLE_OUT_DEVICE_MASK.

### DIFF
--- a/BootloaderCommonPkg/Library/Ext23Lib/Ext2FsLs.c
+++ b/BootloaderCommonPkg/Library/Ext23Lib/Ext2FsLs.c
@@ -105,6 +105,7 @@ STATIC CONST CHAR8 *CONST mTypeStr[] = { "unknown", "REG",  "DIR",  "CHR",  "BLK
 
   @param[in] File           pointer to an file private data
   @param[in] Pattern        not used for now
+  @param[in] ConsoleOutFunc redirect output message to a console
 
   @retval EFI_SUCCESS       list directories or files successfully
   @retval EFI_NOT_FOUND     not found specified dir or file
@@ -113,8 +114,9 @@ STATIC CONST CHAR8 *CONST mTypeStr[] = { "unknown", "REG",  "DIR",  "CHR",  "BLK
 EFI_STATUS
 EFIAPI
 Ext2fsLs (
-  IN  OPEN_FILE     *File,
-  IN  CONST CHAR8   *Pattern
+  IN  OPEN_FILE         *File,
+  IN  CONST CHAR8       *Pattern,
+  IN  CONSOLE_OUT_FUNC   ConsoleOutFunc
   )
 {
   EFI_STATUS      Status;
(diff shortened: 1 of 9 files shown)


myguitar:
> Thank you for making this change.
> This solution has lots of pool allocation/de-allocation, do you think it might be easier to just pass in a print callback function pointer ?
> EX: ListDir (mFsHandle, DirFilePath, ShellPrint);

That's my initial implementation exactly. Thanks for the comment. Let me simply revert this to previous commit.

BootloaderCommonPkg/Library/FatLib/FatLib.c
```diff
@@ -38,9 +40,9 @@ PrintFileName (
   }
 
   if (File->Attributes & FAT_ATTR_DIRECTORY) {
-    DEBUG ((DEBUG_INFO, "  %s/\n", FileName));
+    ConsoleOutFunc (L"  %s/\n", FileName);
```
mauricema: Better to check if  ConoleOutFunc is NULL since it is passed in from outside.

BootloaderCommonPkg/Library/FatLib/FatLib.c
```diff
@@ -38,9 +40,9 @@ PrintFileName (
...
   } else {
-    DEBUG ((DEBUG_INFO, "  %-16s %d\n", FileName, File->FileSize));
+    ConsoleOutFunc (L"  %-16s %d\n", FileName, File->FileSize);
```
mauricema: Better to check if  ConoleOutFunc is NULL since it is passed in from outside.

BootloaderCommonPkg/Library/Ext23Lib/Ext2FsLs.c
```diff
@@ -214,15 +216,15 @@ Ext2fsLs (
...
+        ConsoleOutFunc (L"  %a/\n", New->EntryName);
       } else if (New->EntryType == 1) {
         FileSize = 0;
         Rc = ReadInode (New->EntryInode, File);
         if (!Rc) {
           Fp = (FILE *)File->FileSystemSpecificData;
           FileSize = Fp->DiskInode.Ext2DInodeSize;
         }
-        DEBUG ((DEBUG_INFO, "  %-16a %d\n", New->EntryName, FileSize));
+        ConsoleOutFunc (L"  %-16a %d\n", New->EntryName, FileSize);
```
mauricema: Better to check if  ConoleOutFunc is NULL since it is passed in from outside.  Similar for line 141 and 219.

mauricema: Maybe have a single check at the function entry.  If ConsoleOutFunc is NULL, return error.

myguitar: Thanks for the comment. You are right. Null checking at the function entry is better way. Let me update it in both Fat and Ext libs.

PR #430, opened 2019-11-02
Klocwork scanning fixes for SBL common code

mauricema:
This patches fixed Klocwork scanning reports in SBL common code.

### DIFF
--- a/BootloaderCommonPkg/Library/IppCryptoLib/auth/pcpngrsaencodec.c
+++ b/BootloaderCommonPkg/Library/IppCryptoLib/auth/pcpngrsaencodec.c
@@ -169,16 +169,14 @@ void gsRSApub_cipher(IppsBigNumState* pY,
 
    BNU_CHUNK_T* dataY = BN_NUMBER(pY);
 
-   if (m != NULL) {
-      cpSize nsY = m->publicExpFun(dataY,
-                                  BN_NUMBER(pX), BN_SIZE(pX),
-                                  RSA_PUB_KEY_E(pKey), BITS_BNU_CHUNK(RSA_PUB_KEY_BITSIZE_E(pKey)),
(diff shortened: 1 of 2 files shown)


BootloaderCommonPkg/Library/IppCryptoLib/auth/pcpngrsaencodec.c
```diff
@@ -169,16 +169,14 @@ void gsRSApub_cipher(IppsBigNumState* pY,
...
-      cpSize nsY = m->publicExpFun(dataY,
-                                  BN_NUMBER(pX), BN_SIZE(pX),
-                                  RSA_PUB_KEY_E(pKey), BITS_BNU_CHUNK(RSA_PUB_KEY_BITSIZE_E(pKey)),
-                                  RSA_PUB_KEY_NMONT(pKey),
-                                  pBuffer);
-      FIX_BNU(dataY, nsY);
-      BN_SIZE(pY) = nsY;
-      BN_SIGN(pY) = ippBigNumPOS;
-   }
+   cpSize nsY = m->publicExpFun(dataY,
```
myguitar: KlocWork reports an error because of unnecessary NULL check?

mauricema: No.  It reported m->praviteExpFun could be NULL.  
But the previous fix added  check to verify if m is NULL or not for many other cases.  That fix does not really address the issue reported by KW and it is not necessary.  m itself is not a problem and it won't be NULL in the code flow.    So remove the "m" check here to match the original code.

myguitar: Got it. Thanks for your explanation.

PR #431, opened 2019-11-02
uart buffer flush during reboot

koktong-ong:
@mauricema this is the patch fot uart buffer flush as per discussion


myguitar:
This has been already merged, but a comment about commit message which has only subject line.
'Signed-off-by' is mandatory and short summary line needs to be in commit message.

PR #433, opened 2019-11-04
Remove hard coding from SecureBoot and IppCrypto libs

lsubash:


myguitar:
Can you run 'python BaseTools/Scripts/PatchCheck.py'?

BootloaderCommonPkg/Include/Library/SecureBootLib.h
```diff
@@ -61,6 +58,9 @@ DoHashVerify (
   @param[in]  Signature       Signature for the data buffer.
   @param[in]  PubKey          Public key data pointer.
   @param[in]  PubKeyHash      Public key hash value when ComponentType is not used.
+  @param[in]  PubKeyHashAlg   Pubkey Hash Alg .
+  @param[in]  SignKeyType     Type of Public key Sign .
+  @param[in]  SignHashAlg     Hash Alg used for signing.
   @param[out] OutHash         Calculated data hash value.
```
gdong1: Do we really support any combination of hash alg and sign type?
Currently we already defined SIG_TYPE_RSA2048_SHA256 and SIG_TYPE_RSA3072_SHA384. is it enough to only passing this kind of combined type to make API simple?

lsubash: In order to support all combinations of hash and sign types, we should keep hashalg and signtype as seperate parameters. We could remove SIG_TYPE_RSA2048_SHA256  and SIG_TYPE_RSA3072_SHA384 definitions.

BootloaderCommonPkg/Include/Library/IasImageLib.h
```diff
@@ -65,7 +65,7 @@ typedef struct {        /* a file (sub-image) inside a boot image */
 //
 typedef struct {
   UINT32  Signature;
-  UINT8 PubKeyMod[RSA_MOD_SIZE];
+  UINT8 PubKeyMod[RSA_MOD_SIZE_2048];
```
myguitar: RSA_MOD_SIZE_MAX? How to handle RSA_MOD_SIZE_3072?

BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
```diff
@@ -39,60 +42,83 @@ DoRsaVerify (
...
+  }
+
+  DEBUG ((DEBUG_INFO, "SignHashAlg 0x%x!\n", SignHashAlg));
+
   InpPubKey = (RSA_PUB_KEY *)PubKey;
   TmpPubKey = PubKeyBuf;
-  for (Index = 0; Index < RSA_MOD_SIZE; Index++) {
-    TmpPubKey[Index] = InpPubKey->PubKeyData[RSA_MOD_SIZE - 1 - Index];
+  for (Index = 0; Index < RsaModSz; Index++) {
+    TmpPubKey[Index] = InpPubKey->PubKeyData[RsaModSz - 1 - Index];
```
myguitar: Buffer size is RSA_MOD_SIZE_2048. if SignKeyType == SIG_TYPE_RSA3072, out of bound array index.

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -166,7 +166,7 @@ GetCapsuleFromRawPartition (
     return EFI_NOT_FOUND;
   }
 
-  if (FwUpdHeader->PubKeySize != RSA_MOD_SIZE + RSA_E_SIZE + sizeof (UINT32)) {
+  if (FwUpdHeader->PubKeySize != RSA_MOD_SIZE_2048 + RSA_E_SIZE + sizeof (UINT32)) {
```
myguitar: RSA_MOD_SIZE_2048 is only supported? What about RSA_MOD_SIZE_3072?

BootloaderCommonPkg/Include/Library/CryptoLib.h
```diff
@@ -15,15 +15,31 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+#define  HASH_TYPE_SHA256              0
+#define  HASH_TYPE_SHA384              1
+#define  HASH_TYPE_SHA512              2
+#define  HASH_TYPE_SM3                 3
 
 #define SHA256_DIGEST_SIZE       32
+#define SHA384_DIGEST_SIZE       48
+#define SM3_DIGEST_SIZE          32
+#define SHA512_DIGEST_SIZE       64
+#define HASH_DIGEST_MAX          SHA384_DIGEST_SIZE
```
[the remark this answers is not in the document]
lsubash: we support only SHA384 currently. SHA512 is not in plan for now.

BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
lsubash: RsaModSz would be set by calling modules based on key info.
Keytype/Auth type to be added to required headers as in config data struct and others,

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
lsubash: I have restricted this patch for only modifying interfaces and hard coding removals in secure libs.
These should be taken care while adding support for RSA3072.

BootloaderCommonPkg/Include/Library/IasImageLib.h
lsubash: It would be modified while adding RSA3072 support

BootloaderCommonPkg/Include/Library/CryptoLib.h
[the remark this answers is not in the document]
myguitar: If so, what about adding SHA512 later when you enable SHA512? It makes confusion for now.

BootloaderCommonPkg/Include/Library/IasImageLib.h
myguitar: Then, can you put the changes for hard-code fix only in this PR and make a separate PR for RSA3072?

BootloaderCommonPkg/Include/Library/SecureBootLib.h
mauricema: Since we are redefining the interfaces anyway, let us make it right at the beginning. 
To me, it makes sense to have separate argument for hash and signing.

BootloaderCommonPkg/Include/Library/IasImageLib.h
mauricema: I think this is for IAS image format only, which will be deprecated.  So we only support RSA2048 for IAS.  To support new RSA3072, it is required to move to container.

BootloaderCommonPkg/Include/Library/CryptoLib.h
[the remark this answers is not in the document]
lsubash: I would change the HASH_DIGEST_MAX  to SHA512.

BootloaderCommonPkg/Include/Library/IasImageLib.h
lsubash: agreed with Maurice.

BootloaderCommonPkg/Include/Library/SecureBootLib.h
lsubash: Guo, I am trying to keep secure lib generic for all components to use.
I understand we might not support all combinations in signing process.
Anyways, in secure lib and IPP crypto wrappers we need to support these parameters as separate.

BootloaderCommonPkg/Include/Library/CryptoLib.h
```diff
@@ -77,6 +93,7 @@ RsaVerify (
   CONST UINT8              *Signature,
   CONST UINT32              SignatureLen,
   CONST UINT8               SignatureType,
+  CONST UINT8               HashType,
```
sagraw2: Two comments:
1) Do we need 'SignatureType' argument? It does NOT convey any information.

2) I believe this method is to verify RSA signature using PKCS1v15 scheme. Should we reflect as such in the API name itself (RsaVerify_PKCS1v15)? In future, we may need to support PSS scheme to and then we can have a separate method (i.e. RsaVerify_PSS)?

BootloaderCommonPkg/Include/Library/SecureBootLib.h
```diff
@@ -8,9 +8,6 @@
 #ifndef __VERIFIED_BOOT_LIB_H__
 #define __VERIFIED_BOOT_LIB_H__
 
-#define  HASH_TYPE_SHA256              0
-#define  HASH_TYPE_SHA384              1
-#define  HASH_TYPE_SHA512              2
 
 #define  SIG_TYPE_RSA2048_SHA256       0
```
sagraw2: These should also be removed from here.

BootloaderCommonPkg/Include/Library/SecureBootLib.h
```diff
@@ -61,6 +58,9 @@ DoHashVerify (
   @param[in]  Signature       Signature for the data buffer.
   @param[in]  PubKey          Public key data pointer.
   @param[in]  PubKeyHash      Public key hash value when ComponentType is not used.
+  @param[in]  PubKeyHashAlg   Pubkey Hash Alg .
```
sagraw2: This method is doing too many things and there are too many arguments.
Should we divide this method into multiple simpler methods?
DoRsaVerifyComponent()
DoRsaVerifyData()

Are there any use-cases where current api is being used in non-component scenarios?

BootloaderCommonPkg/Include/Library/CryptoLib.h
lsubash: "SignatureType" is changed to sign_type. Comments were to be updated.

Yes, we could change the API name as RsaVerify_PKCS1v15 to reflect PKCS1v15.

BootloaderCommonPkg/Include/Library/SecureBootLib.h
lsubash: Instead of splitting this API, could we define a new param struct for combining args together?

BootloaderCommonPkg/Include/Library/CryptoLib.h
sagraw2: The first argument 'PubKey' should completely define what kind of PubKey it is : i.e. 2K or  3K.
HashType will define what kind of hashing algo was used during signature generation.

What purpose does 'sign_type' argument serve here?

BootloaderCommonPkg/Include/Library/SecureBootLib.h
sagraw2: I still feel we need separate API here to differentiate Comp and non-comp use case for a simpler API interface to consumer.

lsubash: Will separate API's for 
DoVerifyComponent() ==> Hash Verification
DoRsaVerifyData() ==> Rsa Verification

mauricema: We have DoHashVerify() already.  What is the difference between DoHashVerify/DoVerifyComponent ?

BootloaderCommonPkg/Include/Library/CryptoLib.h
mauricema: At higher API layer, we should use generic function name such as SignatureVerify () so that it will work for whatever new algorithm we want to use in future .   Algorithm should be a parameter.  Internally inside library, we can have more specific function implementation.

BootloaderCommonPkg/Include/Library/SecureBootLib.h
lsubash: Discussed with Sachin, I will come up design for adding data structures for Signature & Public key and updated API interfaces.

BootloaderCommonPkg/Include/Library/CryptoLib.h
lsubash: "sign_type"/"sign_len" -  this info should be already available in calling interfaces.
Reusing the info, instead of calculations again.

BootloaderCommonPkg/Include/Library/IasImageLib.h
myguitar: Okay. Understood.

lsubash:
I would update a pull request once  PR#462 gets merged.
These changes need to be reworked on top of PR#462

PR #435, opened 2019-11-05
Add hash type param to bootloader build and signing tools

lsubash:
hash_type parameter added to build tool API's as required and
current supported hash in tools is for SHA2_256.

Added functionality for retriving RSA private key type.

### DIFF
--- a/BootloaderCorePkg/Tools/BuildUtility.py
+++ b/BootloaderCorePkg/Tools/BuildUtility.py
@@ -314,7 +314,7 @@ def gen_flash_map_bin (flash_map_file, comp_list):
 def copy_expanded_file (src, dst):
 	gen_cfg_data ("GENDLT", src, dst)
 
-def gen_config_file (fv_dir, brd_name, platform_id, pri_key, cfg_db_size, cfg_size, cfg_int, cfg_ext):
+def gen_config_file (fv_dir, brd_name, platform_id, pri_key, cfg_db_size, cfg_size, cfg_int, cfg_ext, hash_type):
 	# Remove previous generated files
 	for file in glob.glob(os.path.join(fv_dir, "CfgData*.*")):
 			os.remove(file)
@@ -400,7 +400,7 @@ def gen_config_file (fv_dir, brd_name, platform_id, pri_key, cfg_db_size, cfg_si
 
 	cfg_final_file = os.path.join(fv_dir, "CFGDATA.bin")
 	if pri_key:
-		cfg_data_tool ('sign', ['-k', pri_key, cfg_merged_bin_file], cfg_final_file)
+		cfg_data_tool ('sign', ['-k', pri_key, '-auth', hash_type, cfg_merged_bin_file], cfg_final_file)
 	else:
 		shutil.copy(cfg_merged_bin_file, cfg_final_file)
 
@@ -470,7 +470,7 @@ def gen_payload_bin (fv_dir, pld_list, pld_bin, priv_key, brd_name = None):
 	gen_container_bin ([pld_list], fv_dir, fv_dir, key_dir, '')
 
 
-def gen_hash_file (src_path, hash_path = '', is_key = False):
+def gen_hash_file (src_path, hash_type, hash_path = '', is_key = False):
 	if not hash_path:
 		hash_path = os.path.splitext(src_path)[0] + '.hash'
 	with open(src_path,'rb') as fi:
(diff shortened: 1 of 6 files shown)


BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -480,7 +480,8 @@ def gen_hash_file (src_path, hash_path = '', is_key = False):
 			raise Exception ("Invalid public key binary!")
 		di = di[4:]
 		di = di[:0x100][::-1] + di[0x100:][::-1]
-	ho = hashlib.sha256(di)
+	if hash_type == 'SHA2_256':
```
mauricema: what if hash_type is not SHA2_256 ?  Need an exception here.

BootloaderCorePkg/Tools/CfgDataTool.py
```diff
@@ -116,6 +117,23 @@ def RsaSignFile(self, InFile, OutFile, IncDat=False, IncKey=False):
         open(OutFile, 'wb').write(Bins)
 
 
+    def GetPrivateKeyInfo (self):
```
mauricema: Shoudn't this be a generic function ?  Should we move it to common  ?

BootloaderCorePkg/Tools/CfgDataTool.py
```diff
@@ -785,6 +803,7 @@ def Main():
                             help='Configuration binary file')
     SignParser.add_argument('-o', dest='cfg_out_file', type=str, help='Signed configuration output binary file name to be generated', required=True)
     SignParser.add_argument('-k', dest='cfg_pri_key', type=str, help='Private key file (PEM format) used to sign configuration data', required=True)
+    SignParser.add_argument('-hash', dest='hash_type', type=str, help='Hash Type with RSA signing', required=True)
```
mauricema: I think it is better to use auth_type instead hash_type.   So it can be potentially used to support new signing algo (non RSA) later on.

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -228,3 +234,43 @@ def compress (in_file, alg, out_path = '', tool_dir = ''):
...
+	temp = re.findall(r'\d+', Match.group(1))
+	key_type = list(map(int, temp))
+
+	print ("Key Type: RSA%s" % key_type[0])
+
+	return key_type
+
+def get_sign_val (sign_type):
+	sign_val = 0
+	if sign_type == "RSA2048":
```
mauricema: use a global dictionary to make it easier

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -228,3 +234,43 @@ def compress (in_file, alg, out_path = '', tool_dir = ''):
...
+		sign_val = 0
+	elif sign_type == "RSA3072":
+		sign_val = 1
+
+	return sign_val
+
+def get_hash_val (hash_type):
+	hash_val = 0
```
mauricema: use a global dictionary to make it easier

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -427,7 +427,7 @@ def SignImage(RawData, OutFile, PrivKey):
     fwupdate_bin_file = 'fwupdate_unsigned.bin'
     open(fwupdate_bin_file, 'wb').write(unsigned_image + RawData)
 
-    rsa_sign_file(PrivKey, pubkey_file, fwupdate_bin_file, OutFile, True, True)
+    rsa_sign_file(PrivKey, pubkey_file, fwupdate_bin_file, OutFile, True, True, self.SIGN_HASH_TYPE)
```
mauricema: since we are  changing interface anyway, let us put important argument at the beginning.
In this case I think the type should be first parameter.

BootloaderCorePkg/Tools/BuildUtility.py
lsubash: will add exception cases where ever its applicable.

lsubash: will do

BootloaderCorePkg/Tools/CfgDataTool.py
lsubash: will move it commonutility

BootloaderCorePkg/Tools/CfgDataTool.py
lsubash: Ok. I would change to Authtype.

BootloaderCorePkg/Tools/CommonUtility.py
lsubash: will do.

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
lsubash: I would add start params for all changed interfaces applicable.

PR #440, opened 2019-11-06
Enable PatchCheck in travis

myguitar:
This will run PatchCheck.py in travis before building a target.
If the format is invalid, travis will exit immediately.


myguitar:
Test done. This will run PatchCheck.py at the beginning of travis and exit if there is any format error.

PR #441, opened 2019-11-06
Klocwork cleanup for PayloadPkg

rprangar:
Signed-off by Ravi Rangarajan.


myguitar:
Missing 'Signed-off-by'

mauricema:
Out format commit msg format usually need at least some detailed description in commit msg beyond the title.  And it is better to add signed off using "git comit --amend -s", it will add both name and email.

PR #446, opened 2019-11-09
Re-arrange CPU info entry order so MpInitLib

mauricema:
This patch sorts the CPU info entry according to their thread distances.
It is required to list CPU thread with furthest distance first in MADT
so as to better utilize CPU resources.

### DIFF
--- a/BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
+++ b/BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
@@ -16,6 +16,59 @@ volatile MP_DATA_EXCHANGE_STRUCT   mMpDataStruct;
 UINT8                             *mBackupBuffer;
 UINT32                             mMpInitPhase = EnumMpInitNull;
 
+/**
+  Sort the CPU entry according to their thread distances
+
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Library/MpInitLib/Ia32/MpInitLib.c
```diff
@@ -16,6 +16,59 @@ volatile MP_DATA_EXCHANGE_STRUCT   mMpDataStruct;
...
+
+  It is required to list CPU thread with further distance first so as to
+  fully utilize unshared CPU resources.
+
+  @param[in]  SysCpuInfo     Pointer to the ALL_CPU_INFO structure.
+
+**/
+STATIC
+VOID
+SortSysCpu (
```
myguitar: What about adding basic sort lib? We are using sort in pci enumeration and shell commands. I think insertion sort can be used as common.

mauricema: can we do this in a separate commit ?

myguitar: Sure. I think a separate PR would be better to cover all cases.

PR #451, opened 2019-11-12
Hash store definition update

lsubash:
Hash Store definitions updated - HASH_STORE_TABLE is updated and HASH_STORE_DATA is added to provide info with variable length sizes. Usage bits would notify for multiple component/key using the same hash.  These data structure would optimize the hash store size  in storage.

TpmLib: TpmExtendStageHash updated to pass DigestHash.
Component Hash is passed with TpmExtendStageHash if available.


BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
```diff
@@ -14,14 +14,15 @@
...
+#ifndef    HASH_DIGEST_MAX
+#define    HASH_DIGEST_MAX   64
 #endif
 #define    MAX_KEY_DIGEST_COUNT 3
 
 typedef struct {
   UINT32               ComponentType;
-  UINT8                Digest[SHA256_DIGEST_SIZE];
+  UINT8                Digest[HASH_DIGEST_MAX];
+  UINT8                HashType;
```
mauricema: I think it is better to make each HASH_ITEM DWORD aligned. 

Also might be good to consider KeyUsage to support using multiple keys to sign a component.
 UINT8                ComponentType;
 UINT8                HashAlgo;
 UINT8                Reserved[2];
 UINT32              KeyUsage;
 UINT8                Digest[SHA256_DIGEST_SIZE];	  UINT8                Digest[HASH_DIGEST_MAX];

lsubash: Yes, Hash item size would be for SHA512 (64 bits).
As this is defined specific to hash store, I believe  we could add supported key usage support should be global PCD.

Also, I am considering to add support for multiple hashes(as sha256, sha384) for components inconsideration of measured boot to extend multiple PCR's.

UINT8 ComponentType;
UINT8 HashAlgoBitMap;
UINT8 Reserved[2];
UINT8 Digest_256[SHA256_DIGEST_SIZE]; 
UINT8 Digest_384[HASH_DIGEST_MAX];

BootloaderCorePkg/Include/HashStore.h
```diff
@@ -22,6 +22,7 @@ typedef struct {
   UINT32             Signature;
   UINT32             Valid;
   HASH_STORE_DATA    Data[HASH_INDEX_MAX_NUM];
+  UINT8              HashType;
```
sagraw2: I would recommend to store HashType in HASH_STORE_DATA to be able to accomodate future use-cases.

BootloaderCorePkg/Include/HashStore.h
```diff
@@ -22,6 +22,7 @@ typedef struct {
...
 } HASH_STORE_TABLE;
```
sagraw2: Do we need versioning for this data structure? Would we be able to update it in future to accomodate any new requirements?

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
sagraw2: @Maurice : Your comment w.r.t KeyUsage applies to the hashes stored in HashStore too. Isn't it?

mauricema: yes, it does. 
BTW, we should consider size as well.  It is in Stage1A, so size is critical.  To better optimize it, maybe we can use a length field in the structure to indicate hash size.  It can poentially support various hash length then.  How about change Reserved to DigestLength ?

BootloaderCorePkg/Include/HashStore.h
mauricema: yes, I agree.  Also should we add a revision field for extension in future ?

BootloaderCorePkg/Include/HashStore.h
mauricema: agree.  captured in previous comment.

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
lsubash: @mauricema Could you let know how could to interpret the KeyUsage?

mauricema: In curtain use, it is required to accept multiple keys to sign a blob.  KeyUsage can help indicate that.  It is very similar to the boot guard key manifest.  Please take that as a reference.

lsubash: Thanks Maurice.
Updated data structures in payloadkeyhash,

typedef struct {
  UINT32 ComponentType;
  UINT8 HashAlgo;
  UINT8 Reserved[2];
  UINT8 KeyUsage;
  UINT8 Digest[HASH_INDEX_MAX_NUM];
} KEY_HASH_ITEM;

typedef struct {
  UINT8                Revision;
  UINT8                Reserved0[3];
  UINT32               DigestCount;
  KEY_HASH_ITEM        KeyHash[0];
} PAYLOAD_KEY_HASH;

BootloaderCorePkg/Include/HashStore.h
lsubash: Updated HashStore data structs,,

typedef struct {
  UINT8  Data[HASH_DIGEST_MAX];
  UINT8  KeyUsage;
} HASH_STORE_DATA;

typedef struct {
  UINT32             Signature;
  UINT8 			 Revison;
  UINT8              Valid;
  UINT8              HashAlgoType;
  UINT8              Reserved;
  HASH_STORE_DATA    Data[HASH_INDEX_MAX_NUM];
} HASH_STORE_TABLE;

BootloaderCommonPkg/Include/Library/BootloaderCommonLib.h
```diff
@@ -131,7 +133,8 @@ GetComponentHash (
 RETURN_STATUS
 SetComponentHash (
   IN       UINT8             ComponentType,
-  IN CONST UINT8            *HashData
+  IN CONST UINT8            *HashData,
+  IN       UINT8             HashType
```
gdong1: How about defining a enum variable for HashType instead of using UINT8?

lsubash: Yeah, i would add enum variables as suggested.
Also, for other supported ones in cryptolib.

lsubash: Guo, I would add the change hash types, signing types and other definitions in Crypto as a separate patch.

lsubash:
Updated Hash definitions. Earlier SHA256 is defined to '0'.
#define  HASH_TYPE_SHA256              1
#define  HASH_TYPE_SHA384              2
#define  HASH_TYPE_SHA512              3
#define  HASH_TYPE_SM3                 4

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
mauricema: After further discussion,  the following is perfered
typedef struct {
UINT32  Usage;
UINT8    HashAlg;
UINT8    Reserved;
UINT16  DigestLen;
UINT8    Digest[0];
} KEY_HASH_ITEM;

BootloaderCorePkg/Include/HashStore.h
mauricema: pls sync with previous comments on the structure.

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
lsubash: We could keep fixed  digest hash to PayloadKeyHashHob where as the HashStore in storage (hashstore.h) can be maintained variable length.

BootloaderCorePkg/Include/HashStore.h
lsubash: This would be structured for variable digest length to optimize the storage.

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
```diff
@@ -14,14 +14,17 @@
...
 #define    MAX_KEY_DIGEST_COUNT 3
 
 typedef struct {
-  UINT32               ComponentType;
-  UINT8                Digest[SHA256_DIGEST_SIZE];
+  UINT32               Usage;
+  UINT8                HashAlg;
+  UINT8                Reserved;
+  UINT16               DigestLen;
+  UINT8                Digest[HASH_DIGEST_MAX];
```
mauricema: Why do we need HASH_DIGEST_MAX ?  Shouldn't be  Digest[0] ?

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
```diff
@@ -600,25 +600,37 @@ TpmIndicateReadyToBoot (
...
+  @param[in] HashExtend       Component hash to be extended
 
   @retval RETURN_SUCCESS      Operation completed successfully.
   @retval Others              Unable to extend stage hash.
 **/
 RETURN_STATUS
 TpmExtendStageHash (
-  IN       UINT8            ComponentType
+  IN       UINT8            ComponentType,
+  IN       UINT32           Signature,
```
mauricema: how Signature get used here ?

BootloaderCorePkg/Library/BootloaderLib/BootloaderLib.c
```diff
@@ -53,48 +92,30 @@ GetComponentHash (
...
 
-  HashIndex = ComponentType;
-  LdrGlobal = GetLoaderGlobalDataPointer ();
-  HashStorePtr = (HASH_STORE_TABLE *)LdrGlobal->HashStorePtr;
-  if ((HashStorePtr == NULL) || (HashIndex != COMP_TYPE_PAYLOAD_DYNAMIC)) {
-    return RETURN_UNSUPPORTED;
+  if (Index == HashStorePtr->DigestCount){
+    return RETURN_NOT_FOUND;
+  } else {
+    *HashData = (UINT8 *)HashEntryData->Digest;
```
mauricema: Check if HashData/HashAlg is NULL pointer before assignment.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: Signature is added for future implementations.
Ex. Signing is done with SHA256 hash and hashstore would be updated for same.
In measured boot, if we need to support SM3 alg, this API TpmExtendStageHash  would calculate SM3 hash and extend. Signature is used with GetComponetInfo() 

GetComponentInfo (**Signature**, &Src, &Length);

mauricema: If you have ComponentType, why need Component signature ?

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
mauricema: If possible, let us use same structure for both storage and HOB so that code can be reused.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: @mauricema I understand GetComponentInfo would use the Signature.
Does it supports with component type?

sagraw2: Also, what is the purpose for 'HashExtend' argument?
If we need a API for directly extending a hash, we should probably add a new API rather then re-using this API.

BootloaderCorePkg/Library/BootloaderLib/BootloaderLib.c
lsubash: will do.

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
lsubash: Yeah, i would maintain same structure for both Storage and HOB

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: This is intended to extend hash provided instead of  retrieving from hashstore. 
As recommended, we could have an separate API to extend Hash directly.

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
```diff
@@ -14,21 +14,26 @@
 ///
 extern EFI_GUID gPayloadKeyHashGuid;
 
-#ifndef    SHA256_DIGEST_SIZE
-#define    SHA256_DIGEST_SIZE   32
+#ifndef    HASH_DIGEST_MAX
+#define    HASH_DIGEST_MAX   64
```
mauricema: why do we do #ifndef ?   Anybody defined this previously?   If so, we should fix it to use a single definition.

lsubash: This is no longer required. I will remove it.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
```diff
@@ -611,14 +611,46 @@ TpmExtendStageHash (
...
+
+  @retval RETURN_SUCCESS               Operation completed successfully.
+  @retval RETURN_INVALID_PARAMETER     Imvalid Digest  buf
+
+  @retval Others                       Unable to extend stage hash.
+**/
+RETURN_STATUS
+TpmExtendHash (
+  IN       UINT8            *Digest,
+  IN       UINT16            HashAlg
```
mauricema: do we need UINT16 ?  or UINT8 ?

BootloaderCorePkg/BootloaderCorePkg.dec
```diff
@@ -111,6 +111,9 @@
 
   gPlatformModuleTokenSpaceGuid.PcdSmbiosTablesSize       |     0x1000 | UINT16 | 0x200000E3
 
+  # Size of the Hash store allocated in bootloader
+  gPlatformModuleTokenSpaceGuid.PcdHashStoreSize       | 0x00000000 | UINT32 | 0x200000F1
```
mauricema: Change the default to be  0x200  , maybe ?

BootloaderCorePkg/Include/HashStore.h
```diff
@@ -8,21 +8,12 @@
 #ifndef __HASH_STORE_H__
 #define __HASH_STORE_H__
 
+#include <Guid/PayloadKeyHashGuid.h>
+
 #define  HASH_STORE_SIGNATURE                SIGNATURE_32('_', 'H', 'S', '_')
-#define  HASH_STORE_DIGEST_LENGTH            32
+#define  HASH_STORE_DIGEST_LENGTH            64
```
mauricema: if this is not required, please remove.

BootloaderCorePkg/Library/BootloaderLib/BootloaderLib.c
mauricema: where did you check  HashAlg is NULL or not ?

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: We expect hash values defined in Cryptolib.h which are NOT bitmasks. UINT8 would be okay.
#define  HASH_TYPE_SHA256              1
#define  HASH_TYPE_SHA384              2
#define  HASH_TYPE_SHA512              3
#define  HASH_TYPE_SM3                 4

BootloaderCorePkg/BootloaderCorePkg.dec
lsubash: Will do

BootloaderCorePkg/Include/HashStore.h
lsubash: Will do

BootloaderCorePkg/Library/BootloaderLib/BootloaderLib.c
lsubash: Done

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
mauricema: let us use UINT8  then.

lsubash:
Validations Done with this patch,
Build tests for TGL, EHL, CFL
Boot tests - TGL, CFL
Tool tests: Container.Sign, Container.Replace()

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
sagraw2: For TPM library interface, we are using variables from IndustryStandard\Tpm20.h file.
Hash is referred by TPMI_ALG_HASH variable in the existing methods in TpmLib.h
We should use that,

BootloaderCommonPkg/Include/Library/TpmLib.h
```diff
@@ -116,6 +116,24 @@ TpmExtendStageHash (
...
+  @param[in] Digest                    Hash Digest to extend
+  @param[in] HashAlg                   Tcg Alg to extend
+
+  @retval RETURN_SUCCESS               Operation completed successfully.
+  @retval RETURN_INVALID_PARAMETER     Imvalid Digest  buf
+
+  @retval Others                       Unable to extend stage hash.
+**/
+RETURN_STATUS
+TpmExtendHash (
```
sagraw2: How is this method different from TpmExtendStageHash?
Can we use a better name to imply the usage?

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
```diff
@@ -14,21 +14,23 @@
...
-  KEY_HASH_ITEM        KeyHash[0];
-} PAYLOAD_KEY_HASH;
-
+  UINT32             Signature;
+  UINT8              Revision;
+  UINT8              Reserved[3];
+  UINT32             UsedLength;
+  UINT32             TotalLength;
+  UINT8              Data[0];
+} HASH_STORE_TABLE;
```
sagraw2: Can we add few lines about this data structure ..i.e. when/how it is used by SBL?

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
```diff
@@ -14,21 +14,23 @@
...
-#define    SHA256_DIGEST_SIZE   32
-#endif
-#define    MAX_KEY_DIGEST_COUNT 3
 
+#pragma pack(1)
 typedef struct {
-  UINT32               ComponentType;
-  UINT8                Digest[SHA256_DIGEST_SIZE];
-} KEY_HASH_ITEM;
+  UINT32           Usage;
```
sagraw2: What are the valid values for Usage?
Let's add a note in the code so that person reading the code can easily understand it.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: SBL crypto hash would be available in stage components through Hash store/container headers. My initial thought was to add a hash convert from SBL crypto hash to TPMI_ALG_HASH inside TpmExtendHash. 
 I would a convert to TPMI_ALG_HASH and pass it.

BootloaderCommonPkg/Include/Library/TpmLib.h
lsubash: In TPMExtendStageHash -  Hash would be retrieved from Hash Store based on component ID.
In TPMExtendHash - it would extend hash provided by calling function.

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
lsubash: will do

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
lsubash: Currently **Usage** bits corresponds to below components. There would be more than one hash for a component as Public Keys.

#define  COMP_TYPE_STAGE_1B            0
#define  COMP_TYPE_STAGE_2             1
#define  COMP_TYPE_PAYLOAD             2
#define  COMP_TYPE_FIRMWARE_UPDATE     3
#define  COMP_TYPE_PUBKEY_CFG_DATA     4
#define  COMP_TYPE_PUBKEY_FWU          5
#define  COMP_TYPE_PUBKEY_OS           6

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -37,28 +37,36 @@ class LZ_HEADER(Structure):
...
     }
 
 # Sign values defined should match with cryptolib.h
 SIGN_TYPE_VALUE = {
 # {   Sign_string:        Sign_Value}
-            "RSA2048"     : 0,
-            "RSA3072 "    : 1,
-            "SM2     "    : 2,
+            "RSA2048"     : 1,
+            "RSA3072 "    : 2,
```
sagraw2: I think we discussed this earlier and decided that in the new data structures, a signature will be identified with 3 properites: 
sig_size (2K, 3K, 4K), key_type(RSA, ECC, etc) and sig_alg(PKCSv1.5, PSS etc).

What is the purpose of SIGN_TYPE_VALUE?

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -37,28 +37,36 @@ class LZ_HEADER(Structure):
...
-            "SHA2_384"       : 1,
-            "SHA2_512"       : 2,
-            "SM3_256"        : 3,
-            "RSA2048SHA256"  : 0,
-            "RSA3072SHA384"  : 1,
+            "SHA2_256"       : 1,
+            "SHA2_384"       : 2,
+            "SHA2_512"       : 3,
+            "SM3_256"        : 4,
+            "RSA2048SHA256"  : 1,
```
sagraw2: How is this being used in build scripts?

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
gdong1: A more clear method is to use typedef enum {}  for such kind of list. But the enum type size depends on compiler although there are some options to make size more specific (e.g. add a big unused number when define enum to use UINT32, or use -fshort-enums for gcc to be one byte if possible) . Later we could explore it further.
For this patch,  how about we rename it clearly as CompType and use UINT8 instead of UINT32 to make it more clear.
and if possible, move these COMP TYPE definition to this file and let SecureBootLib.h include this file, make sense?

gdong1: Maybe we could use following definition. It uses UINT8 and leverage compiler to have a data type check.
//
// Hash Table Definition for Component and Publick Key 
//
typedef UINT8   HASH_TABLE_COMP_TYPE;
#define COMP_TYPE_STAGE_1B 0
#define COMP_TYPE_STAGE_2   1
... ...
typedef struct {
  HASH_TABLE_COMP_TYPE  HashCompType;
  UINT8                                 HashAlg;
  UINT16                               DigestLen;
  UINT8                                 Digest[0];
} HASH_STORE_DATA;

BootloaderCorePkg/Tools/CommonUtility.py
lsubash: Yes, This won't get used. I would clean up this.

BootloaderCorePkg/Tools/CommonUtility.py
lsubash: In GenContainer, we have Authtypes supported as {SHA2_256, SHA2_384, RSA2048SHA256,..}
For extracting the Hash Type from AuthType in container header , dictionary  AUTH_TYPE_HASH_VALUE could be used. This is added for upcoming patches.

BootloaderCommonPkg/Include/Guid/PayloadKeyHashGuid.h
gdong1: After sync with Maurice and Subash, got to know the usage is the bitmask usage for hashes. So the final change would be similar as below:
//
// Hash Table Definition for Component and Publick Key Usage
//
typedef UINT32  HASH_COMP_USAGE;
#define HASH_USAGE_STAGE_1B  BIT0
#define HASH_USAGE_STAGE_2   BIT1
#define HASH_USAGE_PAYLOAD  BIT2
#define HASH_USAGE_FIRMWARE_UPDATE  BIT3
#define HASH_USAGE_ROOT_KEY  BIT8
#define HASH_USAGE_PUBKEY_CFG_DATA   BIT9
#define HASH_USAGE_PUBKEY_FWU      BIT10
#define HASH_USAGE_PUBKEY_OS        BIT11
typedef struct {
HASH_COMP_USAGE  HashCompUsage;
UINT8 HashAlg;
UINT8    Reserved;
UINT16 DigestLen;
UINT8 Digest[0];
} HASH_STORE_DATA;

lsubash: Thanks Guo. I have moved hash component definitions to keyHashGuid.h.
Regarding bitwise, added reference definitions for usage. These definitions wouldn't consumed now as we need to change many API definitions.

PR #455, opened 2019-11-14
Add hash type param to GenContainer tool

lsubash:
sigining hash type parameter added with command line
args for create, replace and sign component image.

### DIFF
--- a/BootloaderCorePkg/Tools/GenContainer.py
+++ b/BootloaderCorePkg/Tools/GenContainer.py
@@ -483,7 +483,7 @@ class CONTAINER ():
 
 		return out_file
 
-	def replace (self, comp_name, comp_file, comp_alg, key_file, new_name):
+	def replace (self, comp_name, comp_file, comp_alg, key_file, hash_type, new_name):
 		if self.header.flags & CONTAINER_HDR._flags['MONO_SIGNING']:
 			raise Exception ("Counld not replace component for monolithically signed container!")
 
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -734,6 +736,7 @@ def main():
 	cmd_display.add_argument('-c',  dest='compress', choices=['lz4', 'lzma', 'dummy'],  default='dummy', help='compression algorithm')
 	cmd_display.add_argument('-a',  dest='auth', choices=['rsa2048', 'sha256', 'none'], default='none',  help='authentication algorithm')
 	cmd_display.add_argument('-k',  dest='key_file',  type=str, default='', help='Private key file path to sign component')
+	cmd_display.add_argument('-ht', dest='hash_type', type=str, required=True, help='Hash Alg for signing')
```
mauricema: -ht should be optional.   you should keep a default value if not specified.

lsubash: Sure, i would update to "SHA2_256" as default one.

myguitar: Can you give choices as well?

lsubash: Currently we have only SHA256 support. While adding SHA384 support in signing, I could add it.

mauricema: I am fine with it.

PR #459, opened 2019-11-18
Fix no SBL serial output after restart Windows

myguitar:
SerialPortLib sets UART Scratch Register to a known value at the first
initialization to avoid unnecessary re-initialization afterward. But,
one of Windows driver(SIO) re-configures UART Registers while the Scratch
Register is preserved. As a result, SBL skips serial port initialization
after restarting from Windows.
Therefore, force to initialize Serial Port at Stage1A all the time simply.

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dsc
+++ b/BootloaderCorePkg/BootloaderCorePkg.dsc
@@ -336,6 +336,7 @@
   $(PLATFORM_PACKAGE)/Stage1A/Stage1A.inf {
     <PcdsFeatureFlag>
       gPlatformCommonLibTokenSpaceGuid.PcdMinDecompression | TRUE
+      gPlatformCommonLibTokenSpaceGuid.PcdForceToInitSerialPort | TRUE
     <LibraryClasses>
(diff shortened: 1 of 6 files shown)


BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -303,6 +303,7 @@
   gPayloadTokenSpaceGuid.PcdCsmeUpdateEnabled             | $(ENABLE_CSME_UPDATE)
   gPlatformCommonLibTokenSpaceGuid.PcdEmmcHs400SupportEnabled | $(ENABLE_EMMC_HS400)
   gPlatformCommonLibTokenSpaceGuid.PcdPreOsCheckerEnabled | $(ENABLE_PRE_OS_CHECKER)
+  gPlatformCommonLibTokenSpaceGuid.PcdForceToInitSerialPort | FALSE
```
mauricema: this is not actually required since it is the same as defined in  DEC.

myguitar: Removed in next push.

myguitar:
> Looks good. Thank you for enhancing the library.

No problem. Thanks for giving me advice to handle this.

PR #471, opened 2019-11-25
Add FSP API reset request handling

mauricema:
Current SBL does not check the FSP Notify API return status. As a
result, if reset is requested, it is ignored by SBL. This patch
added FSP reset request check for FspMemoryInit, FspSiliconInit and
FspNotifyPhase.


BootloaderCorePkg/Include/Library/FspApiLib.h
```diff
@@ -126,4 +126,16 @@ RebaseFspComponent (
...
 
+
+/**
+  This function will handle FSP reset request.
+
+  @param[in] Status     FSP API return status.
+
+**/
+VOID
+FspResetHandler (
```
gdong1: + EFIAPI for the library APIs

mauricema: Thanks,  let me add it.

PR #473, opened 2019-11-26
[APL] Provide WMI service for FWU trigger

mauricema:
This patch added WMI service in ACPI to provide a unified mechnism
for user to trigger a firmware update flow. New ASL methods are
defined. \DWMI.WQ00 is for read and \DWMI.WS00 is for write.

To use it from Linux (Ubuntu):
  - Install acpi-call module
    sudo apt install acpi-call-dkms
  - Load module
    sudo modprobe acpi_call
  - Read trigger register value
    echo '\DWMI.WQ00 0' | sudo tee /proc/acpi/call
    sudo cat /proc/acpi/call
  - Write trigger register value
    echo '\DWMI.WS00 0 0x12' | sudo tee /proc/acpi/call

To use it in Windows:
  - Open a command line window with admin right.
  - Run VB script to read / write the trigger register
    set Service = GetObject("winmgmts:root/wmi")
    set EnumSet = Service.InstancesOf ("AcpiFirmwareCommunication")
    for each Instance in EnumSet
      Wscript.Echo "Current  Val: " & Hex(instance.Command)
      instance.Command = 1
      instance.Put_()
      Wscript.Echo "Set New Val: " & Hex(instance.Command)
    next 'instance


Platform/CommonBoardPkg/AcpiTables/Dsdt/FwuWmi.asl
```diff
@@ -0,0 +1,115 @@
...
+    0x01,         // Instance Count
+    0x00          // Flags
+  })
+
+  Method(WQ00, 1)
+  {
+    Return (FWUC(0, 0))
+  }
+
+  Method(WS00, 2)
```
rprangar: Arg0 is not used. Still necessary to take 2 args?

mauricema: It is defined by WMI calling convention.  Arg0 is instance.

PR #476, opened 2019-11-28
Update debug device support

gdong1:
PcdDebugOutputDeviceMask is defined in BootloaderCommonPkg
They are a little duplicated. so move PcdDebugInterfaceFlags to
BootloaderCommonPkg and removed PcdDebugOutputDeviceMask.

Add a new PCD PcdDebugDeviceNumber to indicate the debug instance
based on debug interface media.

For the value for SetDebugPort/GetDebugPort, Bit7 set means ISA
serial port, else means PCH serial port. Update other code for
above changes.


BootloaderCommonPkg/Include/Guid/LoaderPlatformDataGuid.h
```diff
@@ -16,9 +17,14 @@ extern EFI_GUID gLoaderPlatformDataGuid;
...
 
-#define  DEBUG_OUTPUT_DEVICE_LOG_BUFFER     BIT0
-#define  DEBUG_OUTPUT_DEVICE_SERIAL_PORT    BIT1
-#define  DEBUG_OUTPUT_DEVICE_CONSOLE        BIT2
+//
+// Debug Interfaces
+//
+#define  DEBUG_DEVICE_RAM                   BIT0
+#define  DEBUG_DEVICE_ISA_SERIAL            BIT1
+#define  DEBUG_DEVICE_USB                   BIT3
```
mauricema: I don't think we should have ISA_SERIAL and SERIAL I/O.
And I don't think we should remove console device.

BootloaderCommonPkg/BootloaderCommonPkg.dec
```diff
@@ -115,7 +115,13 @@
   gPlatformCommonLibTokenSpaceGuid.PcdSpiIasImage2RegionSize   |0x00000000|UINT32|0x2000019C
 
   ## This PCD controls enabled debug output devcie
-  gPlatformCommonLibTokenSpaceGuid.PcdDebugOutputDeviceMask    |0x00000003|UINT32|0x20000400
+  # Debug Interfaces. BIT0-RAM, BIT1-ISA UART, BIT2 -Not used, BIT3-USB3, BIT4-Serial IO, BIT5-TraceHub
+  gPlatformCommonLibTokenSpaceGuid.PcdDebugInterfaceFlags      |0x00000011|UINT32|0x20000400
```
mauricema: Why do we rename PCD and make changes to the bit definitions ?

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
```diff
@@ -38,7 +38,8 @@ CpuHalt (
   DEBUG ((DEBUG_ERROR, "\nSTAGE_%a: System halted!\n", mStage[GetLoaderStage()]));
 
   // Flush all console buffer if serial console is not active
-  if ((PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_SERIAL_PORT) == 0) {
+  if ( ((PcdGet32 (PcdDebugInterfaceFlags) & (DEBUG_DEVICE_ISA_SERIAL | DEBUG_DEVICE_SERIALIO)) == 0) &&
+       ((PcdGet32 (PcdDebugInterfaceFlags) & DEBUG_DEVICE_RAM) != 0)) {
```
mauricema: Why do we have dependency on DEBUG_DEVICE_RAM ?

BootloaderCommonPkg/Library/BootloaderDebugLib/DebugLib.c
```diff
@@ -75,24 +74,20 @@ DebugPrint (
...
   //
-  if (PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_LOG_BUFFER) {
+  if ((PcdGet32 (PcdDebugInterfaceFlags) & DEBUG_DEVICE_RAM) != 0) {
     DebugLogBufferWrite  ((UINT8 *)Buffer, Length);
   }
 
-  OutputToSerial = (PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_SERIAL_PORT) ? TRUE : FALSE;
-  if (PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_CONSOLE) {
-    ConsoleWrite ((UINT8 *)Buffer, Length);
+  ConsoleWrite ((UINT8 *)Buffer, Length);
```
mauricema: Why do we always do console write ?

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -508,7 +508,8 @@ BeforeOSJump (
   DEBUG ((DEBUG_INIT, "\n%a\n\n", Message));
 
   // Print debug log buffer if serial port is not an active debug output device
-  if ((PcdGet32 (PcdDebugOutputDeviceMask) & DEBUG_OUTPUT_DEVICE_SERIAL_PORT) == 0) {
+  if ( ((PcdGet32 (PcdDebugInterfaceFlags) & (DEBUG_DEVICE_ISA_SERIAL | DEBUG_DEVICE_SERIALIO)) == 0) &&
+       ((PcdGet32 (PcdDebugInterfaceFlags) & DEBUG_DEVICE_RAM) != 0)) {
```
mauricema: same question as previous one.

Platform/ApollolakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
```diff
@@ -46,22 +46,25 @@ EarlyPlatformDataCheck (
...
-  StitchData = (STITCH_DATA *)&Data;
+  StitchData = (STITCH_DATA *)(0xFFFFFFF4);
   if (StitchData->Marker != 0xAA) {
-    // No data, set default debug port to 2
     // PlatformID will be deferred to be detected
-    SetDebugPort (2);
+    DebugDevice = PcdGet8 (PcdDebugDeviceNumber);
+    if ((PcdGet32 (PcdDebugInterfaceFlags) & DEBUG_DEVICE_ISA_SERIAL) != 0) {
+      // Indicate ISA serial debug device is used
+      DebugDevice |= BIT7;
```
mauricema: why do we do it for APL ?    APL has no legacy UART at all.

Silicon/CoffeelakePkg/Library/PlatformHookLib/PlatformHookLib.c
```diff
@@ -66,7 +66,7 @@ GetSerialPortBase (
 
   DebugPort = GetDebugPort ();
   if (DebugPort >=  PCH_MAX_SERIALIO_UART_CONTROLLERS) {
-    if (DebugPort == 0xFE) {
+    if (DebugPort == 0x80) {
```
mauricema: why do we change this ?

BootloaderCommonPkg/BootloaderCommonPkg.dec
mauricema: Mask seems to be more reasonable to me.  And it matches with other PCD such as ConsoleOuputDeviceMask, ConsoleInputDeviceMask, etc.

BootloaderCommonPkg/BootloaderCommonPkg.dec
```diff
@@ -115,7 +115,13 @@
...
+
+  ## This PCD controls deug device number
+  # Serial IO device: 0 - Serial UART0, 1 - Serial UART1, 2 - Serial UART2,
+  # ISA UART:         0 - ISA UART device (base 0x2F8),   1 - ISA UART device (base 0x3F8)
+  gPlatformCommonLibTokenSpaceGuid.PcdDebugDeviceNumber        |0x02|UINT8|0x20000410
```
mauricema: I think it is better to let platform decide the mapping between Debug port number and the actual debug device.  It is hard to cover it in common.   I also think it is better to call PcdDebugPortNumber to align with GetDebugPort/GetDebugPort.

BootloaderCommonPkg/BootloaderCommonPkg.dec
gdong1: PcdDebugInterfaceFlags is used in FSP UPD in 5 new platforms.  Moving PcdDebugInterfaceFlags from BootloaderCorePkg to BootloaderCommonPkg so that it could be used for all the drivers. PcdDebugInterfaceFlags could be used to set FSP UPD directly.  PcdDebugOutputDeviceMask is just removed to avoid duplicated.

BootloaderCommonPkg/BootloaderCommonPkg.dec
gdong1: Will rename PcdDebugDeviceNumber to PcdDebugPortNumber. For the given debug interface ( RAM, ISA UART, serial IO port, USB, ITH, etc), PcdDebugPortNumber is the debug port number starting from 0. It is very clear to read the code.
The platform could override it if really want to do so.

BootloaderCommonPkg/Include/Guid/LoaderPlatformDataGuid.h
gdong1: There are several UPDs related with ISA serial and PCH serial. Had better have a common way for user to set which one is expected to use.  here we listed debug device interface, console doesn't belong to debug device.

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
gdong1: DEBUG_DEVICE_RAM means the debug info is saved in memory. If this one is not enabled, the debug info should be empty. no point to do output.

BootloaderCommonPkg/Library/BootloaderDebugLib/DebugLib.c
gdong1: In general the new code has same logic.   ConsoleWrite will check if need output to console device.

PayloadPkg/OsLoader/OsLoader.c
gdong1: see above reply.

Platform/ApollolakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
gdong1: will remove ISA code for APL.

Silicon/CoffeelakePkg/Library/PlatformHookLib/PlatformHookLib.c
gdong1: using DebugPort (a byte) to know which serial device  (ISA serial or PCH serial) and port number. If BIT7 is set, it is ISA serial device, else it is PCH serial device. other bits are used for port number, for ISA serial, 0 means 0x2F8 device, 1 means 0x3F8 device.

BootloaderCommonPkg/Include/Guid/LoaderPlatformDataGuid.h
mauricema: In SBL, ISA UART and PCH UART are handled in the same way.  We don't need a special type for each of them.  FSP might have its own definitions, we don't have to change it to match FSP.  Instead, platform code can do necessary remapping to set the FSP UPD properly.
Console debug device is very special for SBL.  FSP will not do it, SBL does support this feature and should not be removed.   It is to redirect debug output to console device.   EX:  Frame buffer is console,  we can redirect debug to frame buffer even though we don't have DebugLib for frame buffer device.

BootloaderCommonPkg/BootloaderCommonPkg.dec
mauricema: These FSP UPD is not architecture or common, so it might change at any time for different FSP.  While SBL common needs to be stable.  I think we should decouple from FSP UPDs, define what required from SBL design.  If required,  platform needs to do the remapping.

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
mauricema: This condition is handled already below.  If it is not enabled, the UsedLength will be the same as Header Length.   So nothing will be output from the code.

BootloaderCommonPkg/Library/BootloaderDebugLib/DebugLib.c
mauricema: Debug output should not call ConsoleWrite() by default.  It needs to be controlled through PCD.   ConsoleLib itself is relative big.  We should not by default call this console interface in DebugLib in early stages when the size is limited.   Also user might not want to output all debug message into console device such as graphics console.

Silicon/CoffeelakePkg/Library/PlatformHookLib/PlatformHookLib.c
mauricema: 0xFE,0xFF has been documented on the github wiki pages and the stitching scripts for CFL.   Unless it is absolutely required,  I would like to leave it and let the platform to decide the meaning for the debug port number.

BootloaderCommonPkg/BootloaderCommonPkg.dec
mauricema: If debug interface has multiple devices set (), what does 0 mean?   is it UART 0 or USB 0 or ITH 0 ?  How do you decide which one you use.   I think debug port number should be a combination of  the debug device and its instance number.

BootloaderCommonPkg/BootloaderCommonPkg.dec
gdong1: Agree it is not necessary for SBL to change its definition to align with FSP.  SBL design should be able to provide required information for FSP. Currently PcdDebugOutputDeviceMask didn't provide if the serial port is ISA type or PCH type, the platform code need this info to set FSP UPDs.

BootloaderCommonPkg/BootloaderCommonPkg.dec
gdong1: Exactly. debug interface should combine with debug port number.

Platform/CoffeelakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
```diff
@@ -153,11 +153,10 @@ EarlyPlatformDataCheck (
   STITCH_DATA          *StitchData;
 
   // Stitching process might pass some plafform specific data.
-  StitchData = (STITCH_DATA *)(0xFFFFFFF4);
+  StitchData = (STITCH_DATA *) 0xFFFFFFF4;
   if (StitchData->Marker != 0xAA) {
-    // No data, set default debug port to 2
     // PlatformID will be deferred to be detected
-    SetDebugPort (0xFF);
+    SetDebugPort ( PcdGet8 (PcdDebugPortNumber));
```
mauricema: The default is 0xFF.  So it needs to add it into BoardConfig.py for CFL platform.

gdong1: Done

PR #488, opened 2019-12-05
Preserve ebx register in AsmEnableAvx()

myguitar:
System reset occurs after returning from AsmEnableAvx() in Stage1A.c
because ebx register is used for cpuid, but not restored.
- Save/Restore ebx register
- Move AsmEnableAvx () after init idt and serialport

### DIFF
--- a/BootloaderCommonPkg/Library/ExtraBaseLib/Ia32/CpuSupport.nasm
+++ b/BootloaderCommonPkg/Library/ExtraBaseLib/Ia32/CpuSupport.nasm
@@ -57,6 +57,7 @@ FlushExit:
 ;------------------------------------------------------------------------------
 global ASM_PFX(AsmEnableAvx)
 ASM_PFX(AsmEnableAvx):
+    push    ebx
     ;
     ; Check if AVX is supported
     ;
@@ -79,4 +80,5 @@ ASM_PFX(AsmEnableAvx):
     xsetbv
 
 NoAvxSupport:
+    pop     ebx
     ret
--- a/BootloaderCorePkg/Stage1A/Stage1A.c
+++ b/BootloaderCorePkg/Stage1A/Stage1A.c
@@ -243,6 +243,9 @@ SecStartup2 (
   BoardInit (PostTempRamInit);
   AddMeasurePoint (0x1040);
 
+  // Enable more CPU featurs
+  AsmEnableAvx ();
+
   if (DebugCodeEnabled()) {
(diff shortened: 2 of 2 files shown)


BootloaderCommonPkg/Library/ExtraBaseLib/Ia32/CpuSupport.nasm
```diff
@@ -57,6 +57,10 @@ FlushExit:
 ;------------------------------------------------------------------------------
 global ASM_PFX(AsmEnableAvx)
 ASM_PFX(AsmEnableAvx):
+    push    eax
```
mauricema: According to calling convention,  EAX, ECX, EDX is not required to be preserved.
And EBX is not used here.    Want to understand why it is necessary.

mauricema:
BTW, on which platform did you see this issue ?

BootloaderCommonPkg/Library/ExtraBaseLib/Ia32/CpuSupport.nasm
mauricema: Sorry,  you are right ,  EBX will be used by CPUID.  
But the code will just need to push/pop EBX.   No need to do it for EAX, ECX and EDX.

myguitar: This is gcc build. I need to go back and check gcc calling convention, but this is to save/restore used registers regardless of calling convention.

myguitar:
> BTW, on which platform did you see this issue ?

In gcc build, I can see the issue on qemu and apl. (haven't checked cfl yet).

BootloaderCommonPkg/Library/ExtraBaseLib/Ia32/CpuSupport.nasm
mauricema: cdecl calling convention should be the same on gcc.  
So only EBX needs to be saved.

mauricema:
Got it.  We will need to save EBX.

BootloaderCommonPkg/Library/ExtraBaseLib/Ia32/CpuSupport.nasm
myguitar: Yes, you are right. Out tool chain is using cdecl calling convention. The eax, ecx and edx are scratch registers, so we don't have to preserve them. Let me save/restore ebx only.

myguitar:
PR has been updated for review.

PR #489, opened 2019-12-06
[CFL] Trigger firmware update from OS

raghavag:
This patch does the following

1) Enable triggering firmware update from OS
2) When firmware update mode is given control, state machine
   is set to capsule processing and firmware update platform
   specific trigger is cleared. State machine will be used
   hereafter to track firmware update
3) Created CheckStateMachine method in BoardSupportLib.c to
   check state machine to see if firmware update is in progress
   and set boot mode to firmware update.
4) Removed CMOS way of triggering firmware update and wrote code
   to use Over-Clocking WDT Scratchpad (OC_WDT_SCRATCH) bits for
   triggering firmware update
5) Update shell fwupdate command to use OC_WDT_SCRATCH bits.
6) Removed extra reset during sbl firmware update
7) Removed reset after updating configuration data update

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -480,78 +480,16 @@ SetStateMachineFlag (
 )
 {
   EFI_STATUS              Status;
-  UINT32                  UpdateSize;
   UINT32                  FwUpdStatusOffset;
-  FW_UPDATE_STATUS        FwUpdStatus;
-  FW_UPDATE_COMP_STATUS   FwUpdCompStatus[MAX_FW_COMPONENTS];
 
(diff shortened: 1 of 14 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -711,6 +700,15 @@ AfterUpdateEnforceFwUpdatePolicy (
     }
   }
 
+  //
+  // When we are setting SM as FW_UPDATE_SM_PART_AB, this indicates
+  // SBL update is done, One more reset is not required, end the
+  // firmware update and reset the system.
+  //
+  if ((FwPolicy.Fields.StateMachine & 0xFF) == FW_UPDATE_SM_PART_AB) {
```
mauricema: why do you AND with 0xFF ?  It is UINT8 already.

mauricema:
This patch does not work on APL.
FWU goes into infinite loop.

PR #490, opened 2019-12-06
Build stuck on unicode locale Windows

myguitar:
This issue happens under two conditions
  1. Unicode language environment in Windows
  2. A python calls 'BaseTools/toolsetup.bat'
     (In EDKII, edksetup.bat directly in Windows command shell)

- 'BuildLoader.py' calls 'BaseTools/toolsetup.bat' in a subprocess
- 'BaseTools/toolsetup.bat' calls 'nmake cleanall'
- 'cleanall' target runs 'python NmakeSubdirs.py' directly
- 'NmakeSubdirs.py' creates multi-threads
-  The threads create another subprocesses

But, one of multi-threads is on deadlock when python handles stdout and
stderr in a subprocess pipe only if the output includes unicode chars.
Therefore, only stderr will be handled in the pipe same as a single
thread call.

### DIFF
--- a/BaseTools/Scripts/PatchCheck.py
+++ b/BaseTools/Scripts/PatchCheck.py
@@ -369,7 +369,7 @@ class GitDiffCheck:
 
         stripped = line.rstrip()
 
-        if self.force_crlf and eol != '\r\n':
+        if self.force_crlf and eol != '\r\n' and not self.filename.endswith('.patch'):
             self.added_line_error('Line ending (%s) is not CRLF' % repr(eol),
                                   line)
         if '\t' in line:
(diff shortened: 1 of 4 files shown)


gdong1:
Are the changes for Basetool coming from EDK2 base tool? If not, should we report issue to EDK2 base tool?

myguitar:
> Are the changes for Basetool coming from EDK2 base tool? If not, should we report issue to EDK2 base tool?

Right, the change has been made in EDK2 BaseTools. I will follow up next step.

gdong1:
> > Are the changes for Basetool coming from EDK2 base tool? If not, should we report issue to EDK2 base tool?
> 
> Right, the change has been made in EDK2 BaseTools. I will follow up next step.
Got it. we need sync EDK2 base tool later sometime.

PR #501, opened 2019-12-12
Add SmbusLib

gdong1:
Add a common SmbusLib support for Intel PCH
Define PLT_MEDIUM_MISC_TYPE in device table for
devices that are not for OS boot.

### DIFF
--- a/BootloaderCommonPkg/Include/Guid/DeviceTableHobGuid.h
+++ b/BootloaderCommonPkg/Include/Guid/DeviceTableHobGuid.h
@@ -14,6 +14,10 @@
 ///
 extern EFI_GUID gDeviceTableHobGuid;
 
+typedef enum {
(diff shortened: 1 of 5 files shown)


mauricema:
There are two many  ASSERT() in code, can we try to use error flow instead ?  
Also for common used parameter checks, might be good to have a single function to do it to save size.

gdong1:
> There are two many ASSERT() in code, can we try to use error flow instead ?
> Also for common used parameter checks, might be good to have a single function to do it to save size.

This patch just did minimal changes based on EDK2 open source library here. we could optimize and maintain it later if required.
https://github.com/tianocore/edk2-platforms/tree/master/Silicon/Intel/CoffeelakeSiliconPkg/Pch/Library/BaseSmbusLib

mauricema:
OK, if so, let us stay with it for now.

PR #502, opened 2019-12-13
Read Config Data from SPI flash using signature header info

lsubash:
Configuration data is read based on info available from
Signature and Public key info headers when verified boot
is enabled.

### DIFF
--- a/Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
+++ b/Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
@@ -118,6 +118,29 @@ PlatformNameInit (
 }
 
 
+/*
+
+  Config Data Layout
+
+--> +---------------------------+
+|-S |       CFG HEADER          |
+| I +---------------------------+
+| G |                           |
+| N |       CFG DATA            |
+| E |                           |
+| D |                           |
+| - |---------------------------+
+|   |       SIGNATURE HDR       |
+|   |---------------------------+
+|   |       SIGNATURE           |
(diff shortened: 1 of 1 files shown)


lsubash:
> Why are we doing this ?
> I saw it was broken into multiple SPI reads, which should be avoided from perf perspective.

We would like to read Signature and Public key info exactly as per the lengths mentioned in header.
I understand there were multiple flash read operations with this approach. Other options could be,
1. Read Max PubKey signature sizes always from flash which would be RSA3K sizes. For RSA2K we would read 256 bytes more data. If you think perf impact would less, we could do this one.
2. Add a Sig+pubkey size parameter in CFG_BLOB where CFG file need to be changed for all platforms.

mauricema:
Can we keep it simple ?
For CFGD and KEYH signing, we  should have known which alg we used for signing.  It should be determined at build time with a fixed PCD in BoardConfig,.py.  And the length is then fixed already.  I did not see why we have to do runtime calculation.

sagraw2:
Can ConfigData be created by a separate tool (i.e. not during BIOS build process)?
If yes, I would suggest to just read max bytes (i.e. RSA3K) and keep the code logic simple.

lsubash:
@Sachin CfgData can be created and updated fwupdate tool. With recent changes to hashstore, we could have 2K and 3K hashes stored and user could use either of this.
Has sync with Maurice, current cfgdata can be either in Bios region and PDR. We would like chap cfgdata in PDR region as only BIOS is getting used.

I would add as separate PR to address PDR removal. Hence, Closing this request.

PR #507, opened 2019-12-18
Add MiniShell support for debug purpose

mauricema:
Soemtimes when JTAG based debug is not available, it might be easier
    to have Shell access in earlier stage to check lots of platform
    settings. Today it is impossible because full Shell has lots of other
    dependencies which might not satisfy in early stage. This patch added
    a PCD PcdMiniShellEnabled to build a mini Shell with very few
    dependencies. This mini Shell can be used in early debug phase for SBL.

    To use it, add the following to override the PCD for a specific stage
    in BootloaderCorePkg.dsc.
    EX:
        $(PLATFORM_PACKAGE)/Stage1B/Stage1B.inf {
          <PcdsFeatureFlag>
            gPlatformCommonLibTokenSpaceGuid.PcdMiniShellEnabled | TRUE
          ...
        }
    Then include ShellLib.h in stage C code and add ShellLib in related
    stage INF file. If adding it into Stage1A, it needs to be after the
    Stage1A banner print out.


BootloaderCommonPkg/Include/Library/ShellLib.h
```diff
@@ -11,7 +11,7 @@
 #include <PiPei.h>
 #include <Library/BaseLib.h>
 
-#define MAX_COMMAND_LINE_LEN 256
+#define MAX_COMMAND_LINE_LEN  80
```
gdong1: 80 may be not long enough for full shell. e.g. to append new command line parameters.
use 80 only when mini shell PCD is defined?

mauricema: Make sense.

mauricema:
Updated a new commit.

PR #513, opened 2019-12-19
Execute version check only for SBL update

raghavag:
This patch will run version check only for SBL update and will
skip for all other components.

Version check for all other components will be added in future
patches.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -362,6 +362,14 @@ VerifyFwVersion (
 
   CurrentBlVersion = NULL;
   CapsuleBlVersion = NULL;
+
+  //
+  // For now - Perform version check only for Slim Bootloader
+  //
+  if (CompareGuid(&ImageHdr->UpdateImageTypeId, &gSblFWUpdateImageFileGuid) != 0) {
+    return EFI_SUCCESS;
+  }
+
   //
   // Get base address of Stage 1A from current firmware
   //


PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -1130,10 +1130,12 @@ UpdateSystemFirmware (
...
-  Status = VerifyFwVersion (ImageHdr, FwPolicy);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, " VerifyFwVersion failed with Status = 0x%x\n", Status));
-    return Status;
+  if (CompareGuid(&ImageHdr->UpdateImageTypeId, &gSblFWUpdateImageFileGuid)) {
+    Status = VerifyFwVersion(ImageHdr, FwPolicy);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, " VerifyFwVersion failed with Status = 0x%x\n", Status));
+      return Status;
+    }
```
gdong1: VerifyFwVersion () will decide how to check the version, check which components, etc.
So better to update VerifyFwVersion ()  for this change.

mauricema: Agree with Guo

mauricema: Also CompareGuid() returns INTN.  So you need to use if (CompareGuid() != 0)

raghavag: Done, updated code according to the comments

PR #514, opened 2019-12-21
TPM and Measured boot support SHA384 and SM3

lsubash:
Current Slimbootloader supports only TPM for SHA256 PCR
This patch introduces flexibility for user to select TPM
PCR bank available while build. Support of SHA384 and
SM3 to TPM measured boot are added.

For first boot SHA256/default TPM PCR would be available which
is default for most supported TPM's. Selected TPM PCR will be
available after reboot.

TPM selection would be based on config data and support would
be added in subsequent patches.

### DIFF
--- a/BootloaderCommonPkg/BootloaderCommonPkg.dec
+++ b/BootloaderCommonPkg/BootloaderCommonPkg.dec
@@ -230,6 +230,16 @@
   # @Prompt Minimum length(in bytes) of the system preboot TCG event log area(LAML).
   gPlatformCommonLibTokenSpaceGuid.PcdTcgLogAreaMinLen|0x10000|UINT32|0x00010081
 
+[PcdsDynamic]
+  ## This PCD indicates the PCR bank to be enabled/supported by Slim Bootloader for measured boot
+  #  Based on the value set, PCR bank world be enabled and extended
+  #  This PCD could be updated based on TPM hash alg info from Config data blob
+  #     0x00000001    - SHA1.<BR>
+  #     0x00000002    - SHA2_256.<BR>
+  #     0x00000004    - SHA2_384.<BR>
+  #     0x00000008    - SHA2_512.<BR>
+  #     0x00000010    - SM3_256.<BR>
+  gPlatformCommonLibTokenSpaceGuid.PcdMeasuredBootHashMask    |         0x00000002 | UINT32  | 0x20000703
 
 
 [PcdsFeatureFlag]
--- a/BootloaderCommonPkg/Include/Library/ContainerLib.h
+++ b/BootloaderCommonPkg/Include/Library/ContainerLib.h
@@ -9,6 +9,8 @@
 #define _CONTAINER_LIB_H_
 
 #include <Library/PcdLib.h>
+#include <Library/CryptoLib.h>
+
 
 #define CONTAINER_LIST_SIGNATURE SIGNATURE_32('C','T','N', 'L')
 
@@ -33,7 +35,16 @@ typedef UINT8 AUTH_TYPE;
 // Attributes for COMPONENT_ENTRY
 #define COMPONENT_ENTRY_ATTR_RESERVED       BIT7
 
-typedef VOID (*LOAD_COMPONENT_CALLBACK) (UINT32 ProgressId);
+
+typedef struct {
+  UINT32           ComponentType;
+  UINT8           *CompBuf;
+  UINT32           CompLen;
+  HASH_ALG_TYPE    HashAlg;
+  UINT8           *HashData;
+} COMPONENT_CALLBACK_INFO;
+
+typedef VOID (*LOAD_COMPONENT_CALLBACK) (UINT32 ProgressId, COMPONENT_CALLBACK_INFO *CbInfo);
 
 typedef struct {
   UINT32           Signature;
--- a/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
+++ b/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
@@ -247,9 +247,9 @@ GetHashAlg(
 
   HashAlg = HASH_TYPE_NONE;
 
-  if(AuthType == AUTH_TYPE_SIG_RSA2048_SHA256){
+  if((AuthType == AUTH_TYPE_SIG_RSA2048_SHA256) || (AuthType == AUTH_TYPE_SHA2_256)) {
     HashAlg = HASH_TYPE_SHA256;
-  } else if (AuthType == AUTH_TYPE_SIG_RSA3072_SHA384){
+  } else if ((AuthType == AUTH_TYPE_SIG_RSA3072_SHA384) ||  (AuthType == AUTH_TYPE_SHA2_384)) {
     HashAlg = HASH_TYPE_SHA384;
   }
 
@@ -679,6 +679,10 @@ LoadComponentWithCallback (
   UINT32                    DstLen;
   UINT32                    ScrLen;
   BOOLEAN                   IsInFlash;
+  COMPONENT_CALLBACK_INFO   CbInfo;
+  UINT32                    ComponentId;
+
+  ComponentId = ContainerSig;
 
   if (ContainerSig < COMP_TYPE_INVALID) {
     // Check if it is component type
@@ -723,7 +727,7 @@ LoadComponentWithCallback (
   }
 
   if (LoadComponentCallback != NULL) {
-    LoadComponentCallback (PROGESS_ID_LOCATE);
+    LoadComponentCallback (PROGESS_ID_LOCATE, NULL);
   }
 
   // Component must have LOADER_COMPRESSED_HEADER
@@ -767,7 +771,7 @@ LoadComponentWithCallback (
     ScrBuf  = (UINT8 *)AllocBuf + ALIGN_UP (SignedDataLen, TEMP_BUF_ALIGN);
     CopyMem (CompBuf, CompData, SignedDataLen);
     if (LoadComponentCallback != NULL) {
-      LoadComponentCallback (PROGESS_ID_COPY);
+      LoadComponentCallback (PROGESS_ID_COPY, NULL);
     }
   } else {
     CompBuf = CompData;
@@ -778,7 +782,18 @@ LoadComponentWithCallback (
   Status = AuthenticateComponent (CompBuf, SignedDataLen, AuthType,
              CompData + ALIGN_UP(SignedDataLen, AUTH_DATA_ALIGN),  HashData, Usage);
   if (LoadComponentCallback != NULL) {
-    LoadComponentCallback (PROGESS_ID_AUTHENTICATE);
+    if(Status == EFI_SUCCESS){
+      // Update component Call back info after authenticaton is done
+      // This info will used by firmware stage to extend to TPM
+      CbInfo.ComponentType    = ComponentId;
+      CbInfo.CompBuf          = CompBuf;
+      CbInfo.CompLen          = SignedDataLen;
+      CbInfo.HashAlg          = GetHashAlg(AuthType);
+      CbInfo.HashData         = HashData;
+      LoadComponentCallback (PROGESS_ID_AUTHENTICATE, &CbInfo);
+    } else {
+      LoadComponentCallback (PROGESS_ID_AUTHENTICATE, NULL);
+    }
   }
   if (!EFI_ERROR (Status)) {
     CompressHdr = (LOADER_COMPRESSED_HEADER *)CompBuf;
@@ -791,7 +806,7 @@ LoadComponentWithCallback (
       Status = Decompress (CompressHdr->Signature, CompressHdr->Data, CompressHdr->CompressedSize,
                            CompBase, ScrBuf);
       if (LoadComponentCallback != NULL) {
-        LoadComponentCallback (PROGESS_ID_DECOMPRESS);
+        LoadComponentCallback (PROGESS_ID_DECOMPRESS, NULL);
       }
       if (EFI_ERROR (Status)) {
         if (ReqCompBase == NULL) {
--- a/BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
+++ b/BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
@@ -259,10 +259,28 @@ TpmLogSpecIDEvent (
   if ((ActivePcr & HASH_ALG_SHA256) != 0) {
     DigestSize[NumberOfAlgorithms].algorithmId = TPM_ALG_SHA256;
     DigestSize[NumberOfAlgorithms].digestSize = SHA256_DIGEST_SIZE;
-
-    DigestSize = &DigestSize[1];
-    ++NumberOfAlgorithms;
+    NumberOfAlgorithms++;
   }
+
+  if ((ActivePcr & HASH_ALG_SHA384) != 0) {
+    DigestSize[NumberOfAlgorithms].algorithmId = TPM_ALG_SHA384;
+    DigestSize[NumberOfAlgorithms].digestSize = SHA384_DIGEST_SIZE;
+    NumberOfAlgorithms++;
+  }
+
+  if ((ActivePcr & HASH_ALG_SHA512) != 0) {
+    DigestSize[NumberOfAlgorithms].algorithmId = TPM_ALG_SHA512;
+    DigestSize[NumberOfAlgorithms].digestSize = SHA512_DIGEST_SIZE;
+    NumberOfAlgorithms++;
+  }
+
+  if ((ActivePcr & HASH_ALG_SM3_256) != 0) {
+    DigestSize[NumberOfAlgorithms].algorithmId = TPM_ALG_SM3_256;
+    DigestSize[NumberOfAlgorithms].digestSize = SM3_DIGEST_SIZE;
+    NumberOfAlgorithms++;
+  }
+
+  DigestSize = &DigestSize[NumberOfAlgorithms];
 
   CopyMem (SpecIdEvent + 1, &NumberOfAlgorithms, sizeof (NumberOfAlgorithms));
 
@@ -345,13 +363,13 @@ TpmLogEvent (
 **/
 VOID
 TpmLogLocalityEvent (
-  IN UINT8 StartupLocality
+  IN UINT8 StartupLocality,
+  IN UINT32 ActivePcrBanks
   )
 {
   TCG_EfiStartupLocalityEvent     StartupLocalityEvent;
   TCG_PCR_EVENT2_HDR              PcrEventHdr;
   TPML_DIGEST_VALUES              *Digests;
-  UINT32                          ActivePcrBanks;
   TPM_LIB_PRIVATE_DATA            *TpmLibData;
 
   CopyMem (StartupLocalityEvent.Signature, TCG_EfiStartupLocalityEvent_SIGNATURE,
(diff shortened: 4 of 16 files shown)


BootloaderCommonPkg/BootloaderCommonPkg.dec
```diff
@@ -222,6 +222,15 @@
   # @Prompt Minimum length(in bytes) of the system preboot TCG event log area(LAML).
   gPlatformCommonLibTokenSpaceGuid.PcdTcgLogAreaMinLen|0x10000|UINT32|0x00010081
 
+[PcdsDynamic]
+  ## This PCD indicates the hash algorithm to be enabled by bootloader for measured boot
+  #  Based on the value set, PCR bank world be enabled and extended
```
sagraw2: "This PCD indicates the PCR bank to be enabled/supported by Slim Bootloader for measured boot"

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -387,7 +387,7 @@ SetupBootImage (
     if (LoaderPlatformInfo != NULL) {
       if (FeaturePcdGet (PcdMeasuredBootEnabled) && (LoaderPlatformInfo->LdrFeatures & FEATURE_MEASURED_BOOT)) {
         // Extend hash of the image into TPM.
-        TpmExtendPcrAndLogEvent (8, TPM_ALG_SHA256, LoadedImage->ImageHash,
+        TpmExtendPcrAndLogEvent (8, (TPMI_ALG_HASH) GetTpmHashAlgEnabled(), LoadedImage->ImageHash,
```
sagraw2: Will the ImageHash be of the same type as the PCR bank enabled in TPM?

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -110,15 +116,33 @@ PreparePayload (
       ComponentEntry = NULL;
       LocateComponentEntry (ContainerSig, ComponentName, NULL, &ComponentEntry);
       if (ComponentEntry != NULL)  {
-        if (ComponentEntry->HashSize == SHA256_DIGEST_SIZE) {
-          TpmExtendHash (ComponentEntry->HashData, HASH_TYPE_SHA256);
+
```
sagraw2: Would it be cleaner to create a separate method in this file to move all the hash extend logic into it?

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
```diff
@@ -17,11 +17,72 @@
...
+GetTcgHashEnabled (
+  VOID
+  )
+{
+  UINT32 TpmHashAlg = 0;
+  UINT8  PcdHashType;
+
+  PcdHashType = PcdGet8(PcdMeasuredBootHashType);
+
+  if(PcdHashType == HASH_TYPE_SHA256){
```
sagraw2: We have three different enums, HASH_TYPE_SHA256, HASH_ALG_SHA256 and TPM_ALG_SHA256.

I am afraid this will cause bugs later.
Can we atleast get rid of one?

BootloaderCommonPkg/BootloaderCommonPkg.dec
mauricema: Do we really need dynamic PCD ?  
We are trying to void if possible since it will increase the binary size.
Can this be a fixed config ?

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
```diff
@@ -604,33 +682,51 @@ TpmIndicateReadyToBoot (
...
   if(Status == EFI_SUCCESS) {
-    Status = TpmExtendPcrAndLogEvent ( 0, TPM_ALG_SHA256, Digest,
-      EV_POST_CODE, POST_CODE_STR_LEN, (UINT8 *)EV_POSTCODE_INFO_POST_CODE);
+    if (HashAlg == PcdGet8(PcdMeasuredBootHashType)){
+
+      Status = TpmExtendPcrAndLogEvent ( 0, (TPMI_ALG_HASH) GetTpmHashAlgEnabled(), Digest,
+        EV_POST_CODE, POST_CODE_STR_LEN, (UINT8 *)EV_POSTCODE_INFO_POST_CODE);
+    } else {
+      DEBUG ((DEBUG_INFO, "Calculate Hash for the components\n"));
+      Status = GetComponentInfo (Signature, &Src, &Length);
```
mauricema: Can we consider doing it at the loading time ?   The normal flow will copy it to memory and then do hash calculation, it is more efficient if the component is in flash.

mauricema:
For the reboot,  will it occur only once if the TPM bank is different ?

BootloaderCommonPkg/BootloaderCommonPkg.dec
lsubash: Platform configdata would define TPM PCR bank to be enabled. This dynamic pcd would be updated and used in TPMlib and necessary libs to check the PCR bank enabled. 
Incase if TPM Alg is NOT available in platform config data(for legacy platforms), default value in dynamic PDC would be used.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: Three hashes,
-  TCG Alg registry (bit mask) - Limited to TPM lib
-  TPM Alg registry - Limited to TPM Lib
- Another one is CryptoLib HASH_ALG_TYPE which is used across SBL.

We could think of aligning Crypto HASH_ALG_TYPE with TCG Alg registry. 
I would take this as separate one as it would involves many file changes including data type changes in API's.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: Thanks. I would check for doing hash calculation at load time and then extend.

BootloaderCorePkg/Stage2/Stage2.c
lsubash: Yes, I would add separate method.

PayloadPkg/OsLoader/OsLoader.c
lsubash: It would be different in some cases. As OsLoader uses container support, I could add condition to check hash type.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
```diff
@@ -162,26 +223,34 @@ TpmPcrBankCheck (
...
     // Update Active PCR banks info in TPM Library data
     TpmLibSetActivePcrBanks (ActivePcrBanks);
 
-    // Check if TPM supports SHA256 PCR bank
-    if ((TpmHashAlgorithmBitmap & HASH_ALG_SHA256) == 0) {
-      DEBUG ((DEBUG_ERROR, "SHA256 hash algorithm is NOT supported. This should never happen with TPM2.0 device. \n"));
+    // Check if TPM supports  PCR bank
+    if (!(TpmHashAlgorithmBitmap & HASH_ALG_SHA256) &&  !(TpmHashAlgorithmBitmap & HASH_ALG_SHA384)
+                                                    && !(TpmHashAlgorithmBitmap & HASH_ALG_SM3_256)){
+      DEBUG ((DEBUG_ERROR, "Hash algorithm is NOT supported. This should never happen with TPM2.0 device. \n"));
```
gdong1: Better uses below to make it follow coding style and clear.
if ((TpmHashAlgorithmBitmap  & (HASH_ALG_SHA256 | HASH_ALG_SHA384 | HASH_ALG_SM3_256))  == 0) {

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
```diff
@@ -17,11 +17,72 @@
...
+
+
+/**
+  Get Tpm Hash Algo type enabled.
+
+  @retval PcdHashType    TPM Algorithm Id.
+**/
+UINT32
+GetTpmHashAlgEnabled (
+  VOID
```
gdong1: Better let this function return TPMI_ALG_HASH or HASH_ALG_TYPE instead of UINT32?

BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
```diff
@@ -259,11 +259,20 @@ TpmLogSpecIDEvent (
   if ((ActivePcr & HASH_ALG_SHA256) != 0) {
     DigestSize[NumberOfAlgorithms].algorithmId = TPM_ALG_SHA256;
     DigestSize[NumberOfAlgorithms].digestSize = SHA256_DIGEST_SIZE;
-
-    DigestSize = &DigestSize[1];
-    ++NumberOfAlgorithms;
+  } else if ((ActivePcr & HASH_ALG_SHA384) != 0) {
```
sagraw2: Let's replace 'else if' with 'if' here.
This will allow in future to support multiple PCR banks simultaneously.

Also, do NumberOfAlgorithms++ in every 'if' section.

BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
```diff
@@ -376,6 +384,18 @@ TpmLogLocalityEvent (
     Digests->digests[Digests->count].hashAlg = TPM_ALG_SHA256;
     ZeroMem (& (Digests->digests[Digests->count].digest), GetHashSizeFromAlgo (TPM_ALG_SHA256));
     Digests->count = Digests->count + 1;
+  } else if ((ActivePcrBanks & HASH_ALG_SHA384) != 0) {
```
sagraw2: Same here. Replace 'else if' with 'if'.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
sagraw2: Although SBL does NOT support multiple active PCR banks, there may be use-case when customer wants to support that. To keep our code scalable, let's keep this as UINT32 itself to reflect a bit mask.

BootloaderCommonPkg/BootloaderCommonPkg.dec
```diff
@@ -222,6 +222,15 @@
...
 
+[PcdsDynamic]
+  ## This PCD indicates the hash algorithm to be enabled by bootloader for measured boot
+  #  Based on the value set, PCR bank world be enabled and extended
+  #  This PCD would be updated based on TPM hash alg info from Config data blob
+  #     0x01    - SHA2_256.<BR>
+  #     0x02    - SHA2_384.<BR>
+  #     0x03    - SHA2_512.<BR>
+  #     0x04    - SM3_256.<BR>
+  gPlatformCommonLibTokenSpaceGuid.PcdMeasuredBootHashType    |         0x1| UINT8  | 0x20000703
```
lsubash: this will be updated to hashmask to support multiple PCR's

BootloaderCommonPkg/Library/TpmLib/TpmEventLog.c
lsubash: yes sachin.

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: This function would be updated to support MeasuredBoot Hashmask .

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: will update.

lsubash:
Updated patch with review comments. Also added following support,
- Enhance API's to support multiple PCR's. Currently only one Active PCR will be supported.
  Added following API's.
  GetCryptoHashAlg 
  TpmLibGetActivePcrBanks
  TpmHashAndExtendPcrEventLog
  TpmExtendOsHash
  
  Other API's were updated.
  
- Updated  PcdMeasuredBootHashType to PcdMeasuredBootHashMask

BootloaderCommonPkg/Library/TpmLib/TpmLib.c
lsubash: Maurice, I will check this optimization as separate patch

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -387,8 +387,7 @@ SetupBootImage (
     if (LoaderPlatformInfo != NULL) {
       if (FeaturePcdGet (PcdMeasuredBootEnabled) && (LoaderPlatformInfo->LdrFeatures & FEATURE_MEASURED_BOOT)) {
         // Extend hash of the image into TPM.
-        TpmExtendPcrAndLogEvent (8, TPM_ALG_SHA256, LoadedImage->ImageHash,
-          EV_COMPACT_HASH, sizeof("LinuxLoaderPkg: OS Image"), (UINT8 *)"LinuxLoaderPkg: OS Image");
+        TpmExtendOsHash (LoadedImage->ImageHash, HASH_TYPE_SHA256);
```
lsubash: Currently HASH_TYPE_SHA256 is used. Will improve to use the hash type from container format.

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -36,6 +36,61 @@ LoadComponentCallback (
...
+    if (HashAlg == GetCryptoHashAlg(PcdGet32(PcdMeasuredBootHashMask))){
+      if (ComponentEntry->HashSize == HashSize) {
+        TpmExtendHash (ComponentEntry->HashData, HashAlg);
+      } else {
+        DEBUG ((DEBUG_INFO, "EPAYLOAD hash does not exist !"));
+        return RETURN_NOT_FOUND;
+      }
+    } else {
+      Status = GetComponentInfo (ContainerSig, &Src, &Length);
```
mauricema: why  do you need this ?    EPayload is always a container.    So if LocateComponentEntry() returned NULL, it is a failure condition already.

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -36,6 +36,61 @@ LoadComponentCallback (
...
+  Extend EPayload to TPM.
+
+  @param[in]  ContainerSig    Containetr Signature
+  @param[in]  ComponentName   Component Name
+
+  @retval RETURN_SUCCESS      Operation completed successfully.
+  @retval RETURN_NOT_FOUND    EPayload hash Not found
+**/
+RETURN_STATUS
+ExtendEPayload (
```
mauricema: ExtendEpayload

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -36,6 +36,61 @@ LoadComponentCallback (
...
+    } else if ((ComponentEntry->AuthType == AUTH_TYPE_SHA2_384) || (ComponentEntry->AuthType == AUTH_TYPE_SIG_RSA3072_SHA384)){
+      HashAlg = HASH_TYPE_SHA384;
+      HashSize = SHA384_DIGEST_SIZE;
+    }
+
+    if (HashAlg == GetCryptoHashAlg(PcdGet32(PcdMeasuredBootHashMask))){
+      if (ComponentEntry->HashSize == HashSize) {
+        TpmExtendHash (ComponentEntry->HashData, HashAlg);
+      } else {
+        DEBUG ((DEBUG_INFO, "EPAYLOAD hash does not exist !"));
```
mauricema: EPAYLOAD hash is invalid !

BootloaderCorePkg/Stage2/Stage2.c
lsubash: This case addresses if measured boot hash type differs from container hash type.
Eg: SM3 used for MB and RSA3K/SHA384 for signing process

BootloaderCorePkg/Stage2/Stage2.c
lsubash: will do

PayloadPkg/OsLoader/OsLoader.c
sagraw2: Instead of adding this new API here, I would recommend to file a separate patch and take care of complete change w.r.t OS hash.

BootloaderCorePkg/Stage2/Stage2.c
sagraw2: Please add this explanation in the code comment appropriately so that future readers 
will know what/why is happening.

PayloadPkg/OsLoader/OsLoader.c
lsubash: TpmExtendPcrAndLogEvent() would enforce to use TPM Hash Alg. Also this API do not check for Active PCR banks. 
New API TpmExtendOsHash() would take care of additional checks.

lsubash:
With recent version fixed comments.

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -36,6 +36,65 @@ LoadComponentCallback (
...
+  Extend EPayload to TPM.
+
+  @param[in]  ContainerSig    Containetr Signature
+  @param[in]  ComponentName   Component Name
+
+  @retval RETURN_SUCCESS      Operation completed successfully.
+  @retval RETURN_NOT_FOUND    EPayload hash Not found
+**/
+RETURN_STATUS
+ExtendEpayload (
```
mauricema: Let us move it into Stage2Support.c file to make Stage2 main flow clean.

BootloaderCorePkg/Stage2/Stage2.c
mauricema: If so, we are recalc the hash using the info from GetComponentInfo.  This is not ideal.   The base will point to data in flash.  The process will be very slow.
During loading, it has been copied into memory already.  Might be better to do it at that time to avoid performance issue.  All TPM extension can occur at the same place where we do hash verification.  Any thoughts ?

BootloaderCorePkg/Stage2/Stage2.c
lsubash: sure

BootloaderCorePkg/Stage2/Stage2.c
sagraw2: Agreed. That is important to avoid reading from flash for hash calculation.

lsubash: We are performing TPM operations for stage components(in stage1B and Stage2) immediately after LoadComponentWithCallback() in which hash verification happens. Yes, We could use the "memory Buffer addr" returned from LoadComponentWithCallback() in case HashCalculation need to be done.

mauricema: Yes, maybe we can enhance the callback to support this.  Then the flow will be simpler to just extend hash in the callback function.

BootloaderCommonPkg/Include/Library/ContainerLib.h
```diff
@@ -33,7 +35,16 @@ typedef UINT8 AUTH_TYPE;
...
 
-typedef VOID (*LOAD_COMPONENT_CALLBACK) (UINT32 ProgressId);
+
+typedef struct {
+  UINT32           ComponentType;
+  UINT8           *CompBuf;
+  UINT32           CompLen;
+  HASH_ALG_TYPE    HashAlg;
+  UINT8           *HashData;
+} COMPONENT_CB_INFO;
```
gdong1: Better use COMPONENT_CALLBACK_INFO if you have other things to update this patch.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -778,7 +782,14 @@ LoadComponentWithCallback (
...
              CompData + ALIGN_UP(SignedDataLen, AUTH_DATA_ALIGN),  HashData, Usage);
   if (LoadComponentCallback != NULL) {
-    LoadComponentCallback (PROGESS_ID_AUTHENTICATE);
+    // Update component Call back info afte authenticaton is done
+    // This info will used by firmware stage to extend to TPM
+    CbInfo.ComponentType    = ComponentId;
+    CbInfo.CompBuf          = CompBuf;
+    CbInfo.CompLen          = SignedDataLen;
+    CbInfo.HashAlg          = GetHashAlg(AuthType);
+    CbInfo.HashData         = HashData;
```
mauricema: How do you know the HashData is valid since the Status returned by Auth() could be error ?  
Should we consider only extending hash when AuthenticateComponent() returns  SUCCESS ?

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -34,6 +39,20 @@ LoadComponentCallback (
   default:
     break;
   }
+
+  if(CbInfo)
```
mauricema: I think you should define a new function for this.  And only call this with  PROGESS_ID_AUTHENTICATE case.

BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -31,6 +34,16 @@ LoadComponentCallback (
   default:
     break;
   }
+
+  //Check the boot mode
+  BootMode = GetBootMode();
```
mauricema: I think you should define a new function for this.  And only call this with  PROGESS_ID_AUTHENTICATE case.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
lsubash: This callback would be invoked only for
- Stage2 authentication where hashdata would be NULL
- Payload and Epayload where hashdata is valid only for Epayload
Incase authentication fails, boot would halt. 
Added a additional check to invoke CB only Auth success cases

BootloaderCorePkg/Stage2/Stage2.c
lsubash: done

PR #515, opened 2020-01-02
Pass Hash Alg param for verifying hash data with RsaVerify

lsubash:
Hash verification of Public key hash should be based on
hash alg used with Hash store. Previously hash alg in
signature info is used. There would be instances where
hashstore hash algorithm differs from signing alg.

### DIFF
--- a/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
+++ b/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
@@ -231,6 +231,32 @@ LocateComponentEntryFromContainer (
 }
 
 /**
+  This function returns the hash alg type from auth type.
+
+  @param[in] AuthType    Authorization Type.
+
+  @retval         Hash Algorithm Type.
+
+**/
+HASH_ALG_TYPE
+GetHashAlg(
+  AUTH_TYPE AuthType
+  )
+{
+  HASH_ALG_TYPE HashAlg;
+
+  HashAlg = HASH_TYPE_NONE;
+
+  if(AuthType == AUTH_TYPE_SIG_RSA2048_SHA256){
+    HashAlg = HASH_TYPE_SHA256;
+  } else if (AuthType == AUTH_TYPE_SIG_RSA3072_SHA384){
+    HashAlg = HASH_TYPE_SHA384;
+  }
+
+  return HashAlg;
+}
+
+/**
   Authenticate a container header or component.
 
   @param[in] Data         Data buffer to be authenticated.
@@ -273,7 +299,7 @@ AuthenticateComponent (
       SignHdr  = (SIGNATURE_HDR *) SigPtr;
       KeyPtr   = (UINT8 *)SignHdr + sizeof(SIGNATURE_HDR) + SignHdr->SigSize ;
       Status   = DoRsaVerify (Data, Length, Usage, SignHdr,
-                             (PUB_KEY_HDR *) KeyPtr, HashData, NULL);
+                             (PUB_KEY_HDR *) KeyPtr, GetHashAlg(AuthType), HashData, NULL);
     } else if (AuthType == AUTH_TYPE_NONE) {
       Status = EFI_SUCCESS;
     } else {
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -855,7 +855,7 @@ AuthenticateCapsule (
     }
   }
 
-  Status    = DoRsaVerify (FwImage, Header->SignatureOffset, HASH_USAGE_PUBKEY_FWU, SignatureHdr, PubKeyHdr, NULL, NULL);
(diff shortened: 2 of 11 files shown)


BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -230,6 +230,28 @@ LocateComponentEntryFromContainer (
...
+/**
+  This function returns the hash alg type from auth type.
+
+  @param[in] AuthType    Authorization Type.
+
+  @retval         Hash Algorithm Type.
+
+**/
+UINT8 GetHashAlg(AUTH_TYPE AuthType){
+  UINT8 HashAlg;
```
gdong1: Why not return HASH_ALG_TYPE instead of UINT8?

lsubash: I would do it.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -847,7 +847,7 @@ AuthenticateCapsule (
     }
   }
 
-  Status    = DoRsaVerify (FwImage, Header->SignatureOffset, HASH_USAGE_PUBKEY_FWU, SignatureHdr, PubKeyHdr, NULL, NULL);
+  Status    = DoRsaVerify (FwImage, Header->SignatureOffset, HASH_USAGE_PUBKEY_FWU, SignatureHdr, PubKeyHdr, PcdGet8(PcdCompSignHashAlg), NULL, NULL);
   if (EFI_ERROR (Status)) {
```
gdong1: To get hash alg, some places PcdCompSignHashAlg and some places use auth type and others. Later we need make it more clear.

lsubash: PcdCompSignHashAlg is used for component hashes and key hash store.
For RSAVerify called from Stage1B,  PubKeyHashAlg would be PcdCompSignHashAlg as configdata verification.

In ContainerFormat, we have AuthType in the header. PubKeyHashAlg would be derived from AuthType info.

gdong1: Understand current status. The CFG data should use container format already. so we could use AuthType from its header. And we would like to use container format for capsule too later.  then we could update capsule verification at that time.

sagraw2: I have a issue with this 'HashAlg' parameter itself.
We should not need it.

The caller should ONLY supply : data to be verified, signature, pub key and the key_type.
Implementation should retrieve the hash and the hash_type from the trusted store (i.e. SBL hash store) and then use it.

I discussed this with Subash and it seems the root problem is our DoHashVerify API .
That API is currently being overloaded and attempting to do multiple things.
We should separate it in two modes:
DoHashVerify (Data, Length,  HashAlg, ExpectedHash) : Generic hash verification.
DoHashVerifyUsingHashStore (Data, Length, ComponentType/KeyUsage) : Use HashStore to verify the hash.

gdong1: Agree to separate 2 usage cases for DoHashVerify to make them more clear. The component hashVerify impl could get component hash internally and call generic DoHashVerify.

lsubash: Agreed, we should not pass hashtype for PubKeyhash for ones in hashstore. We should query from hashstore for algtype. We need to think about cases for pub keys which might not be in hashstore.

Also We should use separate API's for as suggested by Sachin above.

lsubash:
DO NOT merge this patch,

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -230,6 +230,31 @@ LocateComponentEntryFromContainer (
...
 
+/**
+  This function returns the hash alg type from auth type.
+
+  @param[in] AuthType    Authorization Type.
+
+  @retval         Hash Algorithm Type.
+
+**/
+HASH_ALG_TYPE GetHashAlg(
```
mauricema: Coding style requires different line between HASH_ALG_TYPE and GetHashAlg.

lsubash:
This patch is still under discussion. Do not merge

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
mauricema: Can you fix this  before we merge ?

lsubash: @Maurice, Sachin and myslef think adding a parameter for pubkeyhash type  may not be required. DoHashVerify need to be separate methods. One method type can query hashalg type  from hash store for pubkey and stage components.

Let us discuss tomorrow and close it.

lsubash: Fixed the coding style issue.

lsubash:
Has discussion with Maurice and Sachin. Adding new pubkeyhash param is required.
Simplification of DoHashVerify functionality would be address later.

Updated patch with fixing code style comments

PR #516, opened 2020-01-06
Python3 Next() compatibility fix

lsubash:
In Python 3 built-in function next() is used to get the next result
from an iterator. next method causes post build failure.


jgutbub:
Can confirm GenContainer.py is able to work with Python3 (and generate same output binary compared with Python2) after this patch is applied.

PR #517, opened 2020-01-07
Create Internal firmware update file

raghavag:
FirmwareUpdate.c is a big file and for convenience
moved update routines to internal file. Also created a internal
header file.


PayloadPkg/FirmwareUpdate/FirmwareUpdate.inf
```diff
@@ -22,6 +22,7 @@
 
 [Sources]
   FirmwareUpdate.c
+  InternalFirmwareUpdate.c
```
gdong1: It is not clear what is inside InternalFirmwareUpdate.c and what this file is used for from the name.
From the function, all functions are related with bios region or cfg region update.  Not sure if container update will be support in this file. If yes, maybe rename it to BiosFwUpdate.c

raghavag: This file is not related to anything specific, going forward I need to add several more functions to support container and component updates and firmware update.c file is getting big. so I split all update routines and only firmware update internal functions to separate file

PayloadPkg/FirmwareUpdate/InternalFirmwareUpdate.c
```diff
@@ -0,0 +1,352 @@
...
+EFI_STATUS
+EFIAPI
+UpdateRegionLessthan4k (
+  IN  UINT64    Address,
+  IN  VOID      *Buffer,
+  IN  UINT32    Length
+  )
+{
+  EFI_STATUS    Status;
+  UINT8         ReadBuffer[0x1000];
```
mauricema: This array to too big to be on stack.  I think it might be better to allocate instead.

PayloadPkg/FirmwareUpdate/InternalFirmwareUpdate.c
```diff
@@ -0,0 +1,352 @@
...
+EFI_STATUS
+EFIAPI
+UpdateRegionBlock (
+  IN  UINT64    Address,
+  IN  VOID      *Buffer,
+  IN  UINT32    Length
+  )
+{
+  EFI_STATUS    Status;
+  UINT8         ReadBuffer[0x1000];
```
mauricema: Same comment as above.

PayloadPkg/FirmwareUpdate/InternalFirmwareUpdate.c
raghavag: Changed to allocate memory

PayloadPkg/FirmwareUpdate/InternalFirmwareUpdate.c
raghavag: Changed to allocate memory

PayloadPkg/FirmwareUpdate/FirmwareUpdate.inf
mauricema: Let us call it  FirmwareUpdateHelper.c

PayloadPkg/FirmwareUpdate/InternalFirmwareUpdate.c
```diff
@@ -0,0 +1,379 @@
...
+  UINT32        Index;
+  UINT8         *Src;
+
+  ReadBuffer = NULL;
+
+  if (Length == 0) {
+    return EFI_SUCCESS;
+  }
+
+  ReadBuffer = AllocatePool (SIZE_4KB);
```
mauricema: Use AllocatePage instead.

PayloadPkg/FirmwareUpdate/InternalFirmwareUpdate.c
```diff
@@ -0,0 +1,379 @@
+/** @file
+Internal functions to update firmware in boot media.
+
+Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
```
mauricema: change copy right for both header and c file

PR #519, opened 2020-01-08
GenContainer: Validate hash alg for RSA key type used

lsubash:
In GenContainer tool supported combinations are
 RSA2048SAH256 and RSA3072SHA384. Check hash type
 to allign with private key used.

### DIFF
--- a/BootloaderCorePkg/Tools/GenContainer.py
+++ b/BootloaderCorePkg/Tools/GenContainer.py
@@ -620,6 +620,18 @@ def gen_layout (comp_list, img_type, sign_hash_alg, out_file, key_dir, key_file)
     layout += end_layout
     return layout
 
+def validate_hash_type (key_file, hash_type):
+    if key_file:
+        key_type = get_key_type (key_file)
+        if key_type == 'RSA2048' and hash_type == 'SHA2_256':
+           return True
+        elif key_type == 'RSA3072' and hash_type == 'SHA2_384':
+           return True
+        else:
+            return False
+
+    return True
+
 def create_container (args):
     layout = ""
     # if '-l', get the layout content directly
@@ -646,10 +658,14 @@ def create_container (args):
             out_dir = os.path.dirname(args.out_path)
             out_file = os.path.basename(args.out_path)
 
+        if validate_hash_type(key_file, args.hash_type) == False:
+            raise Exception ("Invalid hash type combination with key used")
         layout = gen_layout (args.comp_list, args.img_type, args.hash_type, out_file, key_dir, key_file)
     comp_dir = args.comp_dir if args.comp_dir else def_inp_dir
     tool_dir = args.tool_dir if args.tool_dir else def_inp_dir
     container_list = eval ('[[%s]]' % layout.replace('\\', '/'))
+    if validate_hash_type(key_file, args.hash_type) == False:
+            raise Exception ("Invalid hash type combination with key used")
     gen_container_bin (container_list, out_dir, comp_dir, key_dir, tool_dir, args.hash_type)
 
 def extract_container (args):
@@ -665,6 +681,8 @@ def replace_component (args):
     data = get_file_data (args.image)
     container = CONTAINER (data)
     container.set_dir_path (args.out_dir, '.', '.', tool_dir)
+    if validate_hash_type(args.key_file, args.hash_type) == False:
+        raise Exception ("Invalid hash type combination with key used")
     file = container.replace (args.comp_name, args.comp_file, args.compress, args.key_file, args.hash_type, args.new_name)
     print ("Component '%s' was replaced successfully at:\n  %s" % (args.comp_name, file))
 
@@ -673,6 +691,8 @@ def sign_component (args):
     compress_alg = compress_alg[0].upper() + compress_alg[1:]
     lz_file = compress (args.comp_file, compress_alg, args.out_dir, args.tool_dir)
     data = bytearray(get_file_data (lz_file))
+    if validate_hash_type(args.key_file, args.hash_type) == False:
+        raise Exception ("Invalid hash type combination with key used")
     hash_data, auth_data = CONTAINER.calculate_auth_data (lz_file, args.auth, args.key_file, args.out_dir, args.hash_type)
     sign_file = os.path.join (args.out_dir, args.sign_file)
     data.extend (b'\xff' * get_padding_length(len(data)))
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -736,7 +745,6 @@ def main():
     cmd_display.add_argument('-c',  dest='compress', choices=['lz4', 'lzma', 'dummy'],  default='dummy', help='compression algorithm')
     cmd_display.add_argument('-a',  dest='auth', choices=['SHA2_256', 'SHA2_384', 'RSA2048', 'RSA3072' 'none'], default='none',  help='authentication algorithm')
     cmd_display.add_argument('-k',  dest='key_file',  type=str, default='', help='Private key file path to sign component')
-    cmd_display.add_argument('-ht', dest='hash_type', type=str,  choices=['SHA2_256', 'SHA2_384'], default='SHA2_256', help='Signing Hash Alg when auth type is RSA2048,RSA3072')
```
mauricema: Why do we remove all hash_type?  
There will be case that the component will not be signed individually.  Instead, it just does hash verification. In this case, there is no public key at all.  How do we know the hash type to use ?

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -736,7 +745,6 @@ def main():
     cmd_display.add_argument('-c',  dest='compress', choices=['lz4', 'lzma', 'dummy'],  default='dummy', help='compression algorithm')
     cmd_display.add_argument('-a',  dest='auth', choices=['SHA2_256', 'SHA2_384', 'RSA2048', 'RSA3072' 'none'], default='none',  help='authentication algorithm')
```
sagraw2: Let's be more specific in the help.

Update this line to this:
cmd_display.add_argument('-a',  dest='auth', choices=['SHA2_256', 'SHA2_384', 'RSA2048_SHA2_256', 'RSA3072_SHA2_384' 'none'], default='none',  help='authentication algorithm')

BootloaderCorePkg/Tools/GenContainer.py
lsubash: Agreed with comments. There were two cases,
- Hash type used in container generation.
- Hash type used in signing mechanism
By removal of hash_type param, it would generate error in cases where this tool used in standalone mode and private key is NULL.

I would update with error check for validating hash type to be used when private key is valid.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -665,15 +674,17 @@ def replace_component (args):
     data = get_file_data (args.image)
     container = CONTAINER (data)
     container.set_dir_path (args.out_dir, '.', '.', tool_dir)
-    file = container.replace (args.comp_name, args.comp_file, args.compress, args.key_file, args.hash_type, args.new_name)
+    hash_type = get_hash_type(args.key_file)
```
sagraw2: Do not use key_file to retrieve hash_type.
Instead use auth_type field.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -665,15 +674,17 @@ def replace_component (args):
...
+    file = container.replace (args.comp_name, args.comp_file, args.compress, args.key_file, hash_type, args.new_name)
     print ("Component '%s' was replaced successfully at:\n  %s" % (args.comp_name, file))
 
 def sign_component (args):
     compress_alg = args.compress
     compress_alg = compress_alg[0].upper() + compress_alg[1:]
     lz_file = compress (args.comp_file, compress_alg, args.out_dir, args.tool_dir)
     data = bytearray(get_file_data (lz_file))
-    hash_data, auth_data = CONTAINER.calculate_auth_data (lz_file, args.auth, args.key_file, args.out_dir, args.hash_type)
+    hash_type = get_hash_type(args.key_file)
```
sagraw2: Do not use key_file to retrieve hash_type.
Instead use auth_type field.

BootloaderCorePkg/Tools/GenContainer.py
sagraw2: For hash verification only,  'auth' variable will list that.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -620,6 +620,18 @@ def gen_layout (comp_list, img_type, sign_hash_alg, out_file, key_dir, key_file)
...
 
+def validate_hash_type (key_file, hash_type):
+    if key_file:
+        key_type = get_key_type (key_file)
+        if key_type == 'RSA2048' and hash_type == 'SHA2_256':
+           return True
+        elif key_type == 'RSA3072' and hash_type == 'SHA2_384':
+           return True
+        else:
+            return False
```
mauricema: remove extra leading space

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -703,7 +723,7 @@ def main():
     cmd_display.add_argument('-t', dest='img_type',  type=str, default='CLASSIC', help='Container Image Type : [NORMAL, CLASSIC, MULTIBOOT]')
     cmd_display.add_argument('-o', dest='out_path',  type=str, default='.', help='Container output directory/file')
     cmd_display.add_argument('-k', dest='key_path',  type=str, default='', help='Input key directory/file')
-    cmd_display.add_argument('-ht', dest='hash_type', type=str, choices=['SHA2_256', 'SHA2_384'], default='SHA2_256', help='Hash Alg for signing')
+    cmd_display.add_argument('-ht', dest='hash_type', type=str, choices=['SHA2_256', 'SHA2_384'], default='SHA2_256', help='Hash Algorithm Type')
```
mauricema: let us use lower cases  "Hash algorithm type" for all occurances

lsubash: Maurice, have question. Currently we support auth types param - 
(SHA2_256, SHA2_384, RSA2048, RSA3072).  Internally these are mapped to (SHA2_256, SHA2_384, RSA2048SHA256, RSA3072SHA384)

Instead of using two params **auth** and **hash_type**, could we use only **auth**  and recommend user to use auth (SHA2_256, SHA2_384, **RSA2048SHA256**, **RSA3072SHA384**) at container create() and Sign()? 
This change would be need an update in all boardconfig files and stitch configs where only "RSA2048" is used.

mauricema: Yes, that is fine with me.

BootloaderCorePkg/Tools/GenContainer.py
mauricema: if we can converge to use auth to cover both cases,  that will be fine.   Then we can remove hash_type.

lsubash:
Do Not Merge. I would come up to converge auth type and hash type.

lsubash:
I would generate an separate patch merging auth and hash types for container generation.

PR #526, opened 2020-01-09
[Stage1] Add Stage1 stack base setup offset

myguitar:
This patch allows to setup Stage1 stack/data in any CAR range.
By default, the stack base offset is 0 from CarBase.

### DIFF
--- a/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
+++ b/BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
@@ -19,6 +19,8 @@ SECTION .text
 
 extern  ASM_PFX(SecStartup)
 extern  ASM_PFX(PcdGet32(PcdStage1StackSize))
+extern  ASM_PFX(PcdGet32(PcdStage1DataSize))
+extern  ASM_PFX(PcdGet32(PcdStage1StackBaseOffset))
 extern  ASM_PFX(EarlyBoardInit)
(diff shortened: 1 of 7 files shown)


BootloaderCorePkg/Stage1A/Ia32/SecEntry.nasm
```diff
@@ -60,8 +62,17 @@ FspApiSuccess:
...
+        add     esp, dword [ASM_PFX(PcdGet32(PcdStage1StackBaseOffset))]
         add     esp, dword [ASM_PFX(PcdGet32(PcdStage1StackSize))]
 
+        ;
+        ; Check stage1 stack/data range
+        ;
+        mov     eax, esp
+        add     eax, dword [ASM_PFX(PcdGet32(PcdStage1DataSize))]
+        cmp     eax, edx
+        jg      $
```
mauricema: Instead of deadloop here.  Can we have a better solution ?
EX:  If it is outside of the CAR range, it might indicate StackBaseOffset is too big.   In case case, we can force to StackBaseOffset to be 0 and continue.  In Stage1A, once we  have DEBUG UART enabled, then we will check if we are using the required StackBaseOffset.  If not, then we can print error message and halt. In this way, it is easier to debug this kind of issue.

myguitar: Thanks for the comment. Let me handle early error status after enabling DEBUG UART.

PR #527, opened 2020-01-09
Added comp signature along with GUID in capsule

raghavag:
Currently each component inside capsule is identified with GUID
but going forward support is being added to update multiple components
inside sbl and container, so larger number of GUID's are required, also
GUID need to be passed for unknown components through command line.

Instead 4 character unique signature from flash map that is passed
through command line to indentify the component is added to the
component header. This signature is used during runtime to indentify
the component from flash map and container.


BootloaderCommonPkg/Include/Library/FirmwareUpdateLib.h
```diff
@@ -41,6 +41,10 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #define CAPSULE_FLAGS_CFG_DATA  BIT0
 
+#define FW_UPDATE_COMP_SBL  SIGNATURE_32('B', 'I', 'O', 'S')
+#define FW_UPDATE_COMP_CSME SIGNATURE_32('C', 'S', 'M', 'E')
+#define FW_UPDATE_COMP_CSMD SIGNATURE_32('C', 'S', 'M', 'D')
+
```
gdong1: How about update FW_UPDATE_COMP_SBL to FW_UPDATE_COMP_BIOS_REGION to match with its signature "BIOS"?
and update FW_UPDATE_COMP_CSME to be FW_UPDATE_COMP_CSME_REGION
FW_UPDATE_COMP_CSMD to FW_UPDATE_COMP_CSME_DRIVER?

raghavag: Okay I will change

gdong1: You acted quickly while I testing and reviewing. 
To make it shorter, better to use FWU_COMP_BIOS_REGION if you want :)

raghavag: Done

Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
```diff
@@ -289,89 +289,126 @@ GetFirmwareUpdateInfo (
...
-  //  +-------------------------+
-
-  FlashMap = GetFlashMapPtr();
-  if (FlashMap == NULL) {
-    return EFI_NOT_FOUND;
-  }
-
-  NonRedundantRegionOffset = \
-                             FlashMap->RomSize - (TopSwapRegionSize + RedundantRegionSize) * 2 - NonRedundantRegionSize;
+  if (CompareGuid(&ImageHdr->UpdateImageTypeId, &gSblCompFWUpdateImageFileGuid) == TRUE) {
```
gdong1: it is not necessary to compare with TRUE since Compare() itself is a BOOLEAN.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -863,18 +856,26 @@ ApplyFwImage (
...
   *ResetRequired = FALSE;
 
-  if (CompareGuid(&ImageHdr->UpdateImageTypeId, &gSblFWUpdateImageFileGuid)) {
-    Status = UpdateSystemFirmware(ImageHdr);
-  } else if (CompareGuid(&ImageHdr->UpdateImageTypeId, &gCfgFWUpdateImageFileGuid)) {
+  if (ImageHdr == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
+  Signature = (UINT32)ImageHdr->UpdateHardwareInstance;
```
gdong1: Later we could think about to optimize capsule header to use component signature directly.

Silicon/QemuSocPkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
```diff
@@ -285,31 +285,31 @@ GetFirmwareUpdateInfo (
   ASSERT (UpdatePartition != NULL);
   UpdatePartition->RegionCount   = RegionNumber;
 
-  if (CompareGuid(&ImageHdr->UpdateImageTypeId, &gCfgFWUpdateImageFileGuid) == TRUE) {
+  if (CompareGuid(&ImageHdr->UpdateImageTypeId, &gSblCompFWUpdateImageFileGuid) == TRUE) {
```
myguitar: No need to check with TRUE since CompareGuid() returns BOOLEAN.

BootloaderCommonPkg/Include/Library/FirmwareUpdateLib.h
myguitar: There are many type casts to check SIGNATURE_32 value. What about using SIGNATURE_64 here?

Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
raghavag: Done

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
raghavag: we might still need the GUID for ESRT compatibility

Silicon/QemuSocPkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
raghavag: Done

BootloaderCommonPkg/Include/Library/FirmwareUpdateLib.h
raghavag: We will use the same 32 byte signature for all components inside the flash map, even though we might change these to 64 but type cast is still required to match the signature to components from flash map.

We also cannot reduce the hardware instance to 32 bytes because we have container name : component name format. Higher bytes will have name of component inside container and lower bytes will have component name

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
gdong1: It is a little confusing for the name UpdateHardwareInstance assigned to Signature. 
Not sure if there is any benefit for ESRT compatibility.  we could think about it later.

PR #531, opened 2020-01-13
Added python 3 support for GenCapsuleFirmware.py

raghavag:
This patch updated GenCapsuleFirmware.py to Python 3
compatible.

### DIFF
--- a/BootloaderCorePkg/Tools/GenCapsuleFirmware.py
+++ b/BootloaderCorePkg/Tools/GenCapsuleFirmware.py
@@ -12,6 +12,7 @@ import sys
 import re
 import struct
 import uuid
+import binascii
 from ctypes import *
 
 sys.dont_write_bytecode = True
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -333,6 +334,8 @@ def DumpInfo(self):
             FmpCapsuleImageHeader.DumpInfo('  ')
             print('')
 
+def UnicodeStrToHexString(Str):
+    return (binascii.hexlify(Str.encode('utf-8'))).decode()
```
gdong1: Is the str really unicode string? or just utf-8 string.

raghavag: in Python 3 all strings are unicode strings.

PR #532, opened 2020-01-13
Extend external Hash store component digest

lsubash:
Extenal component hash is calculated if PcdMeasuredBootHashMask
differs from PcdCompSignHashAlg


BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -125,6 +125,35 @@ PrepareStage2 (
...
+VOID
+CheckExtHashStoreMbDigestExtend (
+    IN STAGE1B_PARAM   *Stage1bParam,
+    IN UINT8           *Src,
+    IN UINT32           Length
+  )
+{
+  // Calculate External hash srore Digest if PcdMeasuredBootHashMask differs with PcdCompSignHashAlg
+  if (GetCryptoHashAlg(PcdGet32(PcdMeasuredBootHashMask)) != PcdGet8(PcdCompSignHashAlg)){
+    CalculateHash ((UINT8 *)Src, Length, GetCryptoHashAlg(PcdGet32(PcdMeasuredBootHashMask)),
```
mauricema: If the required hash algo is not supported,  the CalculateHash will fail.  In this case, you cannot set the valid bit.

BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -627,15 +670,29 @@ ContinueFunc (
...
+        // Extend External Config Data hash
         if (Stage1bParam->ConfigDataHashValid == 1) {
           TpmExtendConfigData (Stage1bParam->ConfigDataHash,
             PcdGet8(PcdCompSignHashAlg),
             (UINT8 *) Stage1bParam->CfgDataAddr,
             ((CDATA_BLOB *) Stage1bParam->CfgDataAddr)->UsedLength);
         }
+
+        //Extend Key manifest hash
+        if (Stage1bParam->KeyHashManifestHashValid == 1) {
```
mauricema: Can we rework the flow to align between config data and hash store ?   Right, they are handled differently.

BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -125,6 +125,35 @@ PrepareStage2 (
...
+/**
+  Check digest for external hash store component.
+
+  @param[in] Stage1bParam   Stage1B parameters.
+  @param[in] Src            Extrenal hash store address
+  @param[in] Length         Used length of external hash store
+
+**/
+VOID
+CheckExtHashStoreMbDigestExtend (
```
sagraw2: This method is updating the hash of the KeyManifestHash store if the hash type does NOT matches with current PCR enabled bank. I would prefer to call this method as : UpdateKeyManifestHashForTPMExtend()

BootloaderCorePkg/Stage1B/Stage1B.c
lsubash: Yes, I will update the condition.

BootloaderCorePkg/Stage1B/Stage1B.c
lsubash: They are handled differently. In Config Data I have introduced an TPMLib API where we handle cases as calcluatehash.
Here in hash store -   external hash store memory read from flash is not available. It is getting overwritten.
So, we are trying handle all cases while hash store is verified. 

I could move additional hash checks done to TPMLib.

BootloaderCorePkg/Stage1B/Stage1B.c
lsubash: Thanks. Will do.

lsubash:
I am planning to move SBL specific functionality in functions as TpmExtendStageHash, TpmExtendConfigData to BootloaderCommon code which would be part of Stage boot.

Closing this to address HashStore Digest as part of it.

PR #533, opened 2020-01-14
TpmLib: Perform cold reset when PCR bank changes requested

lsubash:
PCR bank allocate would be performed when requested PCR
differ from default PCR bank. System reset has to be performed
after PCR bank allocation. By default SHA256 is enabled.
For SHA384 and SM3 measured boot usecases system reset has
to be performed for first boot.


lsubash:
Discussed with Sachin. As its would have impact to other platforms, closing this request.
We need to implement reset policy SBL.

PR #536, opened 2020-01-16
Add GIT tool check for build

myguitar:
GIT is required to get FSP binary, ucode patch and so on.
This patch will check GIT availability.

### DIFF
--- a/BootloaderCorePkg/Tools/BuildUtility.py
+++ b/BootloaderCorePkg/Tools/BuildUtility.py
@@ -798,30 +798,52 @@ def check_for_python():
 
     return version
 
+def print_tool_version_info(cmd, version):
+    try:
+        if os.name == 'posix':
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -798,30 +798,52 @@ def check_for_python():
...
 
+def check_for_git():
+    '''
+    Verify GIT executable is available
+    '''
+    cmd = 'git'
+    try:
+        version = subprocess.check_output([cmd, '--version']).decode().strip()
+    except:
+        print('ERROR: GIT not available. Please install GIT.')
```
gdong1: Git might be installed but not in env. 
change to "ERROR: GIT not found. Please check if GIT is in the PATH environment variable"

myguitar: Error message has been updated.

PR #541, opened 2020-01-23
Add a common api to patch PSS table

myguitar:
This will allow update PSS table in a common way.
For a platform specific power calculation,
a function pointer can be provided.

### DIFF
--- a/Platform/CommonBoardPkg/Include/Library/BoardSupportLib.h
+++ b/Platform/CommonBoardPkg/Include/Library/BoardSupportLib.h
@@ -12,6 +12,22 @@
 #include <Library/FirmwareUpdateLib.h>
 
 /**
+  A function pointer to get relative power number in mW
+
+  @param[in]  BaseRatio     Base bus ratio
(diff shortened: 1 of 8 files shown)


Platform/CommonBoardPkg/Include/Library/BoardSupportLib.h
```diff
@@ -72,4 +88,34 @@ CheckStateMachine (
...
+  @param[in]      PackageMinPower   Minimum package power
+  @param[in]      GetRelativePower  A func pointer to get relative power
+  @param[in]      DoListAll         List all from LFM to Turbo
+
+  @retval         EFI_SUCCESS       Patch done successfully
+  @retval         others            Error occured during patching the table
+
+**/
+EFI_STATUS
+AcpiPatchPssTable (
```
mauricema: Can we pass in a data structure pointer instead ?   
It is too many parameters to pass.

myguitar: Thanks for the feedback. Let me make a following up PR.

PR #545, opened 2020-01-29
Avoid hardcoded flashmap address

mauricema:
Current FLASH_MAP address is hard-coded at 0xFFFFFFF8. It will work
in most of the cases.  However, if region is added on top of the
Stage1A FV, it will cause FLASH_MAP address shift.  Instead, the code
can use relative address to locate FLASH_MAP.  The address can be
calculated with (Stage1AFvBase + Stage1AFvSize + 0xFFFFFFF8).

### DIFF
--- a/BootloaderCorePkg/Stage1A/Stage1A.c
+++ b/BootloaderCorePkg/Stage1A/Stage1A.c
@@ -47,8 +47,8 @@ GetFlashMapBufInfo (
   UINT32         Stage1aFvBase;
 
   FlashMap      = NULL;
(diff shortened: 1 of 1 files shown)


rprangar:
Cant we just Stage1AFvBase + Stage1AFvSize - 4?

mauricema:
yes, it will be the same.  We can do "-8", it is the same as 0xfffffff8.   I just want to keep that address since in most of the cases, that will still be the correct flash map address.   If you do have a big concern, we can already change it back.

PR #546, opened 2020-01-29
Allocate BN Buf with Rsa Verification

lsubash:
Currently BN buf is statically defined and this
would increase stack sizes. Allocate the required memory.

### DIFF
--- a/BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
+++ b/BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
@@ -13,6 +13,7 @@
 #include "pcptool.h"
 
 #include <Library/CryptoLib.h>
+#include <Library/BlMemoryAllocationLib.h>
 
 /* Wrapper function for RSAVerify to make the inferface consistent.
  * Returns non-zero on failure, 0 on success.
@@ -22,18 +23,19 @@ int VerifyRsaSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *Signa
   int    sz_n;
   int    sz_e;
   int    sz_rsa;
-  int    sz_adj;
+  int    sz_scratch;
   int    signature_verified;
 
   Ipp8u  *rsa_n;
   Ipp8u  *rsa_e;
   Ipp16u  mod_len;
-  Ipp8u  bn_buf[5200];
+  Ipp8u  *bn_buf;
   IppsBigNumState *bn_rsa_n;
   IppsBigNumState *bn_rsa_e;
   Ipp8u *scratch_buf;
   IppStatus err;
   IppsRSAPublicKeyState *rsa_key_s;
+  Ipp8u *bn_buf_ptr;
   const IppsHashMethod  *pHashMethod = NULL;
 
   rsa_n = (Ipp8u *) PubKeyHdr->KeyData;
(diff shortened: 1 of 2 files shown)


BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
```diff
@@ -54,6 +56,10 @@ int VerifyRsaSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *Signa
     return err;
   }
 
+  // Allocate BN Buf of size 6K bytes
+  bn_buf_sz = 6*1024;
+  bn_buf = AllocatePool (bn_buf_sz);
```
myguitar: Please check NULL return. No need to do ZeroMem?

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
```diff
@@ -63,37 +69,44 @@ int VerifyRsaSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *Signa
   bn_rsa_e = (IppsBigNumState *)(bn_buf + sz_adj);
   sz_adj = sz_adj + ((sz_e + sizeof(Ipp32u)) & ~(sizeof(Ipp32u)-1));
 
-  if (sz_adj > sizeof(bn_buf)) {
+  if (sz_adj > bn_buf_sz) {
```
myguitar: Why don't you calculate sz_adj first, then allocate necessary memory?

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
mauricema: also there is API available to get the required buffer size in IPP.
Please try to use API to get it and then allocate it.

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
mauricema: let us try to allocate the required size from beginning.

lsubash:
Thanks Maurice and Aiden. I would check for IPP API's  to get required size.

lsubash:
> Pls use ippsRSA_GetBufferSizePublicKey() to get the buffer size.

Thanks Maurice. I have done scratch buf allocation using this API.

myguitar:
One concern. FreeMem does not de-allocate memory. Multiple MemAlloc (somewhat big here) may cause an issue before memory init.
@mauricema Isn't AllocateTemporaryMemory() better option?

lsubash:
> One concern. FreeMem does not de-allocate memory. Multiple MemAlloc (somewhat big here) may cause an issue before memory init.
> @mauricema Isn't AllocateTemporaryMemory() better option?

This API can be used. We should follow an order while deallocation using FreeTemporaryMemory.

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
```diff
@@ -54,49 +56,71 @@ int VerifyRsaSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *Signa
...
+  bn_buf_ptr   = bn_buf;
+  rsa_key_s    = (IppsRSAPublicKeyState*) bn_buf_ptr;
+  bn_buf_ptr   = bn_buf_ptr + sz_rsa;
+  bn_rsa_n     = (IppsBigNumState *) bn_buf_ptr;
+  bn_buf_ptr   = bn_buf_ptr + sz_n;
+  bn_rsa_e     = (IppsBigNumState *) bn_buf_ptr;
+
   err = ippsBigNumInit(mod_len / sizeof(Ipp32u), bn_rsa_n);
   if (err != ippStsNoErr) {
+    FreeTemporaryMemory (bn_buf);
```
mauricema: can we have a better flow to handle error ?  too many duplicated free memory here.

mauricema:
> One concern. FreeMem does not de-allocate memory. Multiple MemAlloc (somewhat big here) may cause an issue before memory init.
> @mauricema Isn't AllocateTemporaryMemory() better option?

Yes, we should use AllocateTemporaryMemory().

lsubash:
Updated error handling with this PR.

PR #550, opened 2020-01-31
Support booting to 64-bit kernel entry point

myguitar:
This will allow OsLoader payload to boot to 64-bit kernel entry point.
If CPU supports 64-bit mode and a kernel image has 64-bit entry point,
OsLoader will switch to 64-bit long mode and jump to the 64-bit entry
point. Otherwise, continue to boot to 32-bit entry point.
- Ported necessary code from VitualMemory.c in MdeModulePkg
- Moved PagingLib from BootloaderCorePkg to BootloaderCommonPkg
- TBD: 64-bit IDT

Next step is to support 64-bit Payload.
- 32-bit compatible mode
- 64-bit CryptoLib
- etc.


BootloaderCommonPkg/Library/PagingLib/Ia32/Paging32.nasm
```diff
@@ -77,18 +77,18 @@ ASM_PFX(DisablePaging32):
 global ASM_PFX(FlushCacheLine)
```
mauricema: This FlushCacheLine can be removed now since we have AsmFlushCacheRange in ExtraBaseLib.  That should be used instead.

BootloaderCommonPkg/Library/PagingLib/Paging32To64.c
```diff
@@ -0,0 +1,658 @@
...
+    UINT64  Global:1;                 // 0 = Not global page, 1 = global page TLB not cleared on CR3 write
+    UINT64  Available:3;              // Available for use by system software
+    UINT64  PAT:1;                    //
+    UINT64  MustBeZero:17;            // Must be zero;
+    UINT64  PageTableBaseAddress:22;  // Page Table Base Address
+    UINT64  AvabilableHigh:11;        // Available for use by system software
+    UINT64  Nx:1;                     // 0 = Execute Code, 1 = No Code Execution
+  } Bits;
+  UINT64    Uint64;
+} PAGE_TABLE_1G_ENTRY;
```
mauricema: Let us move all related page structures into PageingLib.h so that it can be used by everybody.

BootloaderCommonPkg/Library/PagingLib/Ia32/Paging32.nasm
myguitar: Okay. Let me remove it.

BootloaderCommonPkg/Library/PagingLib/Paging32To64.c
myguitar: Sure, I will move them to PagingLib.h.

BootloaderCommonPkg/Library/LinuxLib/LinuxLib.c
```diff
@@ -316,11 +330,24 @@ LinuxBoot (
...
 {
   BOOT_PARAMS   *Bp;
+  UINTN          KernelStart;
 
   Bp = GetLinuxBootParams ();
   if (Bp != NULL) {
     UpdateLinuxBootParams (Bp);
-    JumpToKernel ((VOID *)Bp->Hdr.Code32Start, Bp);
+    KernelStart = (UINTN)Bp->Hdr.Code32Start;
+    if (!IsLongModeEnabled ()) {
```
gdong1: Is it a typo for "!IsLongModeEnabled () " here? from logic it looks should be IsLongModeEnabled ()

myguitar: This means current processor mode. IsLongModeEnabled() will return TRUE if current processor mode is long mode, otherwise return FALSE.
If current processor mode is protected mode, processor supports long mode and Linux has 64-bit entry point, JumpToLongMode will switch to long mode and jump to 64-bit entry point.

gdong1: if IsLongModeEnabled() is TRUE, it would CpuDeadLoop(). Is this the expected behavior?

myguitar: It's TBD since there is no 64-bit code to run at this moment. After enabling Long Mode in previous stage, Long Mode kernel jump code (64-to-64 code, not 32-to-64 jump code) will be added.

gdong1: Got it. That's the confusing part. Ideally IsLongModeEnabled () should be added later when we support 64 bit code.

PR #551, opened 2020-02-02
Add missing header files in INF

mauricema:
In order to sync up with EDK201911 stable release, it is required
to add missing header files in the INF file. Otherwise, the build
will throw warnings. This patch added the missing headers in INF
files.


gdong1:
without adding header files to INF, it still builds success with the warning message, , right?

mauricema:
that is right.

PR #554, opened 2020-02-03
Fix several minor issues in the code

raghavag:
1) GetComponentInfoByPartition does not check FlashMapPtr, added code
   to return error if FlashMapPtr is NULL

2) In GetComponentHash function, there is no check for if HashEntryPtr is
   greater than or equal to HashEndPtr. In this case HashEntryData will be
   corrupted.

3) In InitFirmwareUpdate function, if we could not find corresponding image in
   capsule, there is an error message but after that we continue to update
   reserved region based on ImageHdr which is not valid. Added code to continue
   if image is not found in capsule.

4) In UpdateSblComponent function, Status is uninitialized. Initialized Status to
   not found.

### DIFF
--- a/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
+++ b/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
@@ -390,6 +390,10 @@ GetComponentInfoByPartition (
   }
 
   FlashMapPtr = GetFlashMapPtr ();
+  if (FlashMapPtr == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
   RomBase = (UINT32) (0x100000000ULL - FlashMapPtr->RomSize);
 
   //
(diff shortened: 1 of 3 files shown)


BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
```diff
@@ -390,6 +390,10 @@ GetComponentInfoByPartition (
   }
 
   FlashMapPtr = GetFlashMapPtr ();
+  if (FlashMapPtr == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
```
myguitar: Is this required? GetComponentEntryByPartition() returns NULL if FlashMapPtr is NULL.
What about changing GetComponentEntryByPartition() API to get FlashMapPtr as a parameter?
FlashMapPtr = GetFlashMapPtr ();
if (FlashMapPtr == NULL) {
  return EFI_NOT_FOUND;
}
Entry = GetComponentEntryByPartition(FlashMapPtr, Signature, IsBackupPartition);
if (Entry == NULL) {
  return EFI_NOT_FOUND;
}

mauricema: I agree it is not a real issue.   But let us not change too much code here to make Klocwork happy.
If line 393 to 396 can resolve the klocwork issue,  how about we just go with it.

myguitar: Okay let's go with this if this change resolves KW report.

PR #555, opened 2020-02-04
KlocWork issue fixes in security libraries

lsubash:

### DIFF
--- a/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
+++ b/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
@@ -93,6 +93,10 @@ DumpHex (
   UINTN Size;
   UINTN Index;
 
+  if (UserData == NULL){
+    return;
+  }
+
   Data = UserData;
   while (DataSize != 0) {
     Size = 16;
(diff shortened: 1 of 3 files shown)


BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
```diff
@@ -93,28 +93,32 @@ DumpHex (
...
 
-    Data += Size;
-    Offset += Size;
-    DataSize -= Size;
+      Data += Size;
+      Offset += Size;
+      DataSize -= Size;
+    }
+  } else {
+    DEBUG ((DEBUG_INFO, "DumpHex: UserData is NULL \n"));
```
mauricema: I don't think we need any debug print here.

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
```diff
@@ -93,28 +93,32 @@ DumpHex (
   UINTN Size;
   UINTN Index;
 
-  Data = UserData;
-  while (DataSize != 0) {
-    Size = 16;
-    if (Size > DataSize) {
-      Size = DataSize;
-    }
+  if (UserData != NULL) {
```
mauricema: let us just do:
if (UserData == NULL) {
  return;
}

lsubash:
Updated PR with review comments.

PR #558, opened 2020-02-04
Fix FPDT size issue

gdong1:
FPDT size should be FIRMWARE_PERFORMANCE_TABLE. And updated
this logic to correct the size and adjust next ACPI table
starting address.

### DIFF
--- a/BootloaderCorePkg/Library/AcpiInitLib/AcpiFpdt.c
+++ b/BootloaderCorePkg/Library/AcpiInitLib/AcpiFpdt.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2019 - 2020, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -183,13 +183,15 @@ UpdateFpdtS3Table (
   Update Firmware Performance Data Table (FPDT).
 
   @param[in] Table          Pointer of ACPI FPDT Table.
+  @param[out] ExtraSize     Extra size the table needed.
 
   @retval EFI_SUCCESS       Update ACPI FPDT table successfully.
   @retval Others            Failed to update FPDT table.
  **/
 EFI_STATUS
 UpdateFpdt (
-  IN VOID                             *Table
+  IN  UINT8                           *Table,
+  OUT UINT32                          *ExtraSize
   )
 {
   FIRMWARE_PERFORMANCE_TABLE          *Fpdt;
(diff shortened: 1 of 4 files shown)


BootloaderCorePkg/Library/AcpiInitLib/AcpiFpdt.c
```diff
@@ -214,8 +216,10 @@ UpdateFpdt (
     CopyMem (S3PerfTable, &mS3PerformanceTableTemplate, sizeof (mS3PerformanceTableTemplate));
     UpdateFpdtBootTable (BootPerfTable);
 
-    // Fixup FPDT table length
-    Fpdt->Header.Length = sizeof (INTERNAL_FIRMWARE_PERFORMANCE_TABLE);
+    if (ExtraSize != NULL) {
+      *ExtraSize = (UINT8 *) (S3PerfTable + 1) - Table - Fpdt->Header.Length;
+    }
+
```
myguitar: Without extra parameter, below would be enough.
Fpdt->Header.Length = (UINT32)(UINTN)((UINT8 *) (S3PerfTable + 1) - Table);

gdong1: FPDT table size should be sizeof (FIRMWARE_PERFORMANCE_TABLE). And later ACPI table checksum will be calculated by this size. we should not update ACPI table size to another value. ExtraSize is required to save the record which is not part of FIRMWARE_PERFORMANCE_TABLE.

myguitar: Got it. Fpdt has only pointers to the records. Then, what about relocating Fpdt and putting records before Fpdt table?
size = bootperftable + s3perftable
relocate Fpdt to (Fpdt + size)
copy record before new Fpdt
*Table = (Fpdt + size)

gdong1: It doesn't matter put perf record before or after FPDT table. We still need return the size of record to caller since *Table could not update current pointer.
We could update the parameter to UINT8 **, but this logic is not as clear as using ExtraSize since the callee would update Current pointer in caller.

myguitar: My bad on invalid pointer operation. Let's do this in simple way. Later we may need to introduce acpi reclaim memory allocation. Some other tables like DMAR also requires some memory. I don't think using ExtraSize parameter in all ACPI table update function is good idea.

gdong1: yes. we could update it as long as we had better options.

PR #559, opened 2020-02-04
Add PCI Enumeration policy info

myguitar:
PciEnumeration() scans a single PCI root bridge currently.
The PCI_ENUM_POLICY_INFO structure will be generated at build time,
and this will allow PCI enumeration more flexible.

typedef struct {
  UINT8           DowngradeIo32;	// default:1
  UINT8           DowngradeMem64; // default:1
  UINT8           DowngradePMem64;// default:1
  UINT8           Reserved;
  UINT8           BusScanType;    // default:0 (0: list, 1: range)
  UINT8           NumOfBus;       // the number of BusScanItems
  UINT8           BusScanItems[0];
} PCI_ENUM_POLICY_INFO;

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2020, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -12,11 +12,25 @@
 #include <Library/PciExpressLib.h>
 #include <Library/SortLib.h>
 #include <InternalPciEnumerationLib.h>
+#include <Library/BootloaderCommonLib.h>
 
 #define  DEBUG_PCI_ENUM    0
-#define  DOWNGRADE_IO32    1
-#define  DOWNGRADE_PMEM64  1
-#define  DOWNGRADE_MEM64   1
+
+typedef enum {
+  BusScanTypeList     = 0,
+  BusScanTypeRange    = 1,
+  BusScanTypeInvalid  = 0xFF
+} BUS_SCAN_TYPE;
+
+typedef struct {
+  UINT8           DowngradeIo32;
+  UINT8           DowngradeMem64;
+  UINT8           DowngradePMem64;
+  UINT8           Reserved;
+  UINT8           BusScanType;
+  UINT8           NumOfBus;
+  UINT8           BusScanItems[0];
+} PCI_ENUM_POLICY_INFO;
 
 UINT8   *mPoolPtr;
 
@@ -176,11 +190,12 @@ PciParseBar (
   IN UINTN          BarIndex
   )
 {
-  UINT32        Value;
-  UINT32        OriginalValue;
-  UINT32        Mask;
-  EFI_STATUS    Status;
-  PCI_BAR_TYPE  BarType;
+  UINT32                Value;
+  UINT32                OriginalValue;
+  UINT32                Mask;
+  EFI_STATUS            Status;
+  PCI_BAR_TYPE          BarType;
+  PCI_ENUM_POLICY_INFO  *EnumPolicy;
 
   BarType       = PciBarTypeUnknown;
   OriginalValue = 0;
@@ -205,6 +220,7 @@ PciParseBar (
     return Offset + 4;
   }
 
+  EnumPolicy = (PCI_ENUM_POLICY_INFO *)PcdGetPtr (PcdPciEnumPolicyInfo);
   PciIoDevice->PciBar[BarIndex].Offset = (UINT8) Offset;
   if ((Value & 0x01) != 0) {
     //
@@ -216,25 +232,22 @@ PciParseBar (
       //
       // It is a IO32 bar
       //
-      BarType = PciBarTypeIo32;
-#ifdef DOWNGRADE_IO32
-      PciIoDevice->PciBar[BarIndex].BarType   = PciBarTypeIo16;
-#else
-      PciIoDevice->PciBar[BarIndex].BarType   = PciBarTypeIo32;
-#endif
+      if ((EnumPolicy != NULL) && (EnumPolicy->DowngradeIo32 == 0)) {
+        BarType = PciBarTypeIo32;
+      } else {
+        BarType = PciBarTypeIo16;
+      }
       PciIoDevice->PciBar[BarIndex].Length    = ((~ (Value & Mask)) + 1);
-      PciIoDevice->PciBar[BarIndex].Alignment = PciIoDevice->PciBar[BarIndex].Length - 1;
-
     } else {
       //
       // It is a IO16 bar
       //
       BarType = PciBarTypeIo16;
-      PciIoDevice->PciBar[BarIndex].BarType   = PciBarTypeIo16;
       PciIoDevice->PciBar[BarIndex].Length    = 0x0000FFFF & ((~ (Value & Mask)) + 1);
-      PciIoDevice->PciBar[BarIndex].Alignment = PciIoDevice->PciBar[BarIndex].Length - 1;
-
     }
+    PciIoDevice->PciBar[BarIndex].BarType   = BarType;
+    PciIoDevice->PciBar[BarIndex].Alignment = PciIoDevice->PciBar[BarIndex].Length - 1;
+
     //
     // Workaround. Some platforms inplement IO bar with 0 length
     // Need to treat it as no-bar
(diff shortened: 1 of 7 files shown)


BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -669,25 +669,26 @@ PciSearchDevice (
 EFI_STATUS
 PciScanBus (
   IN PCI_IO_DEVICE                      *Bridge,
-  IN UINT8                              StartBusNumber,
-  OUT UINT8                             *SubBusNumber,
-  OUT UINT8                             *PaddedBusRange
+  IN UINT16                             StartBusNumber,
+  OUT UINT16                            *SubBusNumber,
+  OUT UINT16                            *PaddedBusRange
```
gdong1: UINT8 should be enough for bus number, do we really need change it to UINT16?

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -1136,43 +1144,62 @@ PciEnumeration (
...
 
   SetAllocationPool (MemPool);
 
-  SubBusNumber = 0;
-  Root = CreatePciIoDevice (NULL, NULL, 0, 0, 0);
-  PciScanBus (Root, 0, &SubBusNumber, NULL);
+  IoBase = PcdGet32 (PcdPciResourceIoBase);
+  Mem32Base = PcdGet32 (PcdPciResourceMem32Base);
+  for (PrimaryBus = 0; PrimaryBus <= PCI_MAX_BUS; PrimaryBus = SubBusNumber + 1) {
+    SubBusNumber = PrimaryBus;
```
gdong1: Not sure the boot performance impact to core platform. If big impact, maybe we could add a policy to decide if we need  scan all the possible buses.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: This is because of "PrimaryBus <= PCI_MAX_BUS". PrimaryBus can be 256 and it causes UINT8 overflow, then 0 again.
This loop can be infinite with UINT8.
for (PrimaryBus = 0; PrimaryBus <= PCI_MAX_BUS; PrimaryBus = SubBusNumber + 1)

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: We can use Pcd to limit max bus iteration. How do you think?
One concern is we need to update start/end bus number in MCFG table. But, all boards has hard-coded value 0 - 0xFF currently. For MCFG table, whole iteration is required and the MCFG field needs to be updated with the detected bridges. This is also TBD.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
gdong1: Bus number range should be 0 ~ 255, totally at most 256 buses. How it could be 256?

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
gdong1: Did you get the boot perf impact for this patch? If the impact is obvious, yes, we could add a PCD for max bus number.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: In PciEnumeration(),
for (PrimaryBus = 0; PrimaryBus <= PCI_MAX_BUS; PrimaryBus = SubBusNumber + 1) {
}
PrimaryBus should be 256(0x100) and then exit the for loop. But, UINT8 makes overflow.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: Let me check perf number.

myguitar: There is huge gap in perf number - 3ms -> 85ms. Let me define a Pcd to limit the number of bus iteration.

gdong1: OK. make the default value to same with the current behavior for core platform.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
gdong1: Got it.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
mauricema: I think the better solution is to align wiht EDK2 to use root bridge concept.   In this way, we don't have to scan the non-existing buses.  Each root bridge will have a start bus.  The PCI scan bus will start from this bus.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -1136,43 +1144,62 @@ PciEnumeration (
...
+    DumpDeivceTree  (Root);
+    DumpResourceBar (Root);
+    DEBUG ((DEBUG_INFO, "MEM Pool Used: 0x%08X\n", (UINT32)GetAllocationPool() - (UINT32)MemPool));
 #endif
 
+    //
+    // Io/Mem32 Base for next bus
+    //
+    IoBase = ALIGN (Root->PciBar[0].BaseAddress + Root->PciBar[0].Length, Root->PciBar[0].Alignment);
+    Mem32Base = MAX (Mem32Base, ALIGN (Root->PciBar[2].BaseAddress + Root->PciBar[2].Length, Root->PciBar[2].Alignment));
```
mauricema: MEM and PMEM should not be allocated in interleaved manner.  All MEM needs to be a single range.  And all PMEM needs to be a different range.   It is to allow WB cache type for preferable memory.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: Let me try to scan only root bridge first, then iterate the bridges and compare perf number.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: Thanks for the feedback. I am going to re-visit this PR.

myguitar:
Added PcdPciEnumPolicyInfo for DowngradeIo32, DowngradeMem64, DowngradePMem64 and RootRridgeEnumMask.
By default, DowngradeIo32=1, DowngradeMem64=1, DowngradePMem64 =1 and enumerate Root Bus 0 only unless a BoardConfig.py overrides PcdPciEnumPolicyInfo.
No perf impact for the existing targets.

mauricema:
Looks good.  
The only comment I have is :  Can you use root bus list directly instead of bit map ?  In this way, it is more straight forward for people to config it.

myguitar:
Thanks for your comment. Let me improve this PR a little more.

myguitar:
PR has been updated to generate PCI_ENUM_POLICY_INFO at build time.

Platform/QemuBoardPkg/BoardConfig.py
```diff
@@ -35,6 +35,18 @@ def __init__(self, *args, **kwargs):
...
+        # _PCI_ENUM_DOWNGRADE_IO32  : 1
+        # _PCI_ENUM_DOWNGRADE_MEM64 : 1
+        # _PCI_ENUM_DOWNGRADE_PMEM64: 1
+        # _PCI_ENUM_BUS_SCAN_TYPE   : 0 (0: list, 1: range)
+        #
+        #self._PCI_ENUM_DOWNGRADE_IO32   = 0
+        #self._PCI_ENUM_DOWNGRADE_MEM64  = 0
+        #self._PCI_ENUM_DOWNGRADE_PMEM64 = 0
+        #self._PCI_ENUM_BUS_SCAN_TYPE    = 1
+        self._PCI_ENUM_BUS_SCAN_ITEMS = '0,10,0x35,128,255,0xff'
```
mauricema: what is the 255, 0xff ?   Are they same ?

myguitar: This is to handle duplicated bus number in gen_pci_enum_policy_info(). Not an issue, but let me remove duplicated number to avoid confusion.

PR #568, opened 2020-02-08
[APL] Enable firmware update trigger through sticky register

mauricema:
This patch added code to:
  - Trigger FWU using sticky register in SBL Shell
  - Detect FWU mode using combination of sticky register and state
    machine
  - Clear the trigger flag at the end of FWU

### DIFF
--- a/Platform/ApollolakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
+++ b/Platform/ApollolakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
@@ -832,15 +832,28 @@ EarlyBootDeviceInit (
 BOOLEAN
 IsFirmwareUpdate ()
 {
-  UINT16          FirmwareUpdateStatus;
+  EFI_STATUS        Status;
+  UINT16            FwuReq;
+  FLASH_MAP        *FlashMapPtr;
+  FW_UPDATE_STATUS  FwUpdStatus;
+  UINT32            OffsetInBiosRgn;
+  UINT32            RsvdBase;
 
-  //
-  // This is platform specific method. Here just use COMS.
-  //
-  IoWrite8 (0x70, FWU_BOOT_MODE_OFFSET);
-  FirmwareUpdateStatus = IoRead8 (0x71);
+  // Check if state machine is set to capsule processing mode.
+  FlashMapPtr = GetFlashMapPtr ();
+  Status = GetComponentInfoByPartition (FLASH_MAP_SIG_BLRESERVED, FALSE, &RsvdBase, NULL);
+  if (!EFI_ERROR (Status) && (FlashMapPtr != NULL)) {
+    OffsetInBiosRgn = FlashMapPtr->RomSize + RsvdBase;
+    ZeroMem (&FwUpdStatus, sizeof(FwUpdStatus));
+    SpiFlashRead (FlashRegionBios, OffsetInBiosRgn, sizeof(FwUpdStatus), (VOID *)&FwUpdStatus);
+    if (CheckStateMachine (&FwUpdStatus) == EFI_SUCCESS) {
+      return TRUE;
+    }
+  }
 
-  if (FirmwareUpdateStatus == FWU_BOOT_MODE_VALUE) {
+  // Check if platform firmware update trigger is set.
+  FwuReq = (MmioRead32 (PMC_BASE_ADDRESS + R_PMC_BIOS_SCRATCHPAD) >> 16) & 0xFF;
+  if ((FwuReq & 0x0F) != 0) {
     return TRUE;
   }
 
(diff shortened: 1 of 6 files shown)


Platform/ApollolakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -832,15 +832,28 @@ EarlyBootDeviceInit (
...
+  FlashMapPtr = GetFlashMapPtr ();
+  Status = GetComponentInfoByPartition (FLASH_MAP_SIG_BLRESERVED, FALSE, &RsvdBase, NULL);
+  if (!EFI_ERROR (Status) && (FlashMapPtr != NULL)) {
+    OffsetInBiosRgn = FlashMapPtr->RomSize + RsvdBase;
+    ZeroMem (&FwUpdStatus, sizeof(FwUpdStatus));
+    SpiFlashRead (FlashRegionBios, OffsetInBiosRgn, sizeof(FwUpdStatus), (VOID *)&FwUpdStatus);
+    if (CheckStateMachine (&FwUpdStatus) == EFI_SUCCESS) {
+      return TRUE;
+    }
+  }
```
gdong1: Should we check only 2 methods (CMOS and scratch bit) here to know if it is firmware update mode?
After firmware update done, the flags in CMOS or scratch bit would be cleared. and it is not necessary to check firmware update machine here.

Silicon/ApollolakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
```diff
@@ -508,5 +511,8 @@ ClearFwUpdateTrigger (
   VOID
   )
 {
+  // Clear platform firmware update trigger.
+  MmioAnd32 (PMC_BASE_ADDRESS + R_PMC_BIOS_SCRATCHPAD, 0xFF00FFFF);
```
myguitar: Instead of PMC_BASE_ADDRESS,
PmcBar = MmioRead32 (MM_PCI_ADDRESS (0, PCI_DEVICE_NUMBER_PMC, PCI_FUNCTION_NUMBER_PMC, R_PMC_BASE)) & ~0x0F;

Platform/ApollolakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
mauricema: We don't use CMOS for APL at all.   For scratch bit, it needs to be converted into state machine since the flag will be lost in case of power failure.

Silicon/ApollolakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
mauricema: PMC device might be hidden.   But the BAR is programmed inside FSP-T.   
Since FSP used fixed base for PMC base (not resource reallocation).  I guess it is fine to do this.

Platform/ApollolakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
gdong1: oh, CMOS method was deleted in this patch.  do we need add CSME method detection here for CSME method?
With this update, it will read SPI block in every boot which should impact boot performance. Maybe we could add a PCD to disable this feature when register method is not required.

mauricema: It will only read less than 16 bytes.   So the time is very limited.   I measured the perf impact, it is less than 1ms.   So it should be fine.
Going forward, register based trigger is the standard way.    And it is also supported from ACPI level.  HECI could be optional now.

gdong1: OK. In the firmware update command, if using "-c" parameter, it would not work if here we don't check CSME method, right? so we will add it in a separate patch?

mauricema: we check both.

gdong1: OK. CSME method is checked in another function after MemoryInit.

Silicon/ApollolakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
myguitar: Got it. I am okay with this. Thanks.

mauricema: thx!

PR #570, opened 2020-02-14
Fix klockwork critical issue

koktong-ong:
Fix the critical issue triggered from Klocwork as shown below:
Function returns address of local variable in line 122 of IasImage.c

### DIFF
--- a/BootloaderCommonPkg/Library/IasImageLib/IasImage.c
+++ b/BootloaderCommonPkg/Library/IasImageLib/IasImage.c
@@ -68,7 +68,7 @@ IsIasImageValid (
 
   if (!FeaturePcdGet (PcdVerifiedBootEnabled)) {
     DEBUG ((DEBUG_INFO, "IAS image verification is skipped!\n"));
-    return Hdr;
+    return (IAS_HEADER *) ImageAddr;
   }
 
   if (!IAS_IMAGE_IS_SIGNED (Hdr->ImageType)) {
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/IasImageLib/IasImage.c
```diff
@@ -68,7 +68,7 @@ IsIasImageValid (
 
   if (!FeaturePcdGet (PcdVerifiedBootEnabled)) {
     DEBUG ((DEBUG_INFO, "IAS image verification is skipped!\n"));
-    return Hdr;
+    return (IAS_HEADER *) ImageAddr;
   }
```
gdong1: How it could solve any issue since Hdr is set to (IAS_HEADER *) ImageAddr in line 49, they are exactly same.

koktong-ong: ImageAddr is the input pointer instead of local variable so it should resolve the issue right? Local pointer will be deleted when exiting the function.

mauricema: Same question here.  The change seems to be equivalent to the original code.

mauricema: I think the fix should 
change:
  IasImageInfo->HashData = DigestHash;
to:
  CopyMem (IasImageInfo->HashData, DigestHash, sizeof(IasImageInfo->HashData));

koktong-ong: Guys I've confused about the input pointer and local pointer. Will update the pull request once again.

gdong1: There is no DigestHash, is the KW based on old code?

koktong-ong: I guess the code built in the server is out dated, I will check again at KW side.

PR #572, opened 2020-02-14
End firmware update incase of failure

raghavag:
Current code does not clear firmware update trigger incase
of failures like, capsule image not found or capsule image
authentication failed.

Code is modified to fix this issue. As part of this change,
EndFirmwareUpdate function is moved from platform code to
core firmwareupdate.c. EndFirmwareUpdate function will call
ClearFwUpdateTrigger and this will end firmware update.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -703,10 +703,10 @@ ProcessCapsule (
       return Status;
     }
 
-    Status = ClearFwUpdateTrigger();
-    if (EFI_ERROR(Status)) {
-      DEBUG((DEBUG_ERROR, "Clearing firmware update trigger failed with error: %r\n", Status));
-    }
+    //
+    // Clear firmware update trigger
+    //
+    ClearFwUpdateTrigger();
   } else {
     return EFI_SUCCESS;
   }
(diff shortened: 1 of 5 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -1073,6 +1073,29 @@ GetRegionInfo (
...
+EndFirmwareUpdate (
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+
+  Status = ClearFwUpdateTrigger ();
+
+  Status = PlatformEndFirmwareUpdate ();
+  return Status;
```
mauricema: There are two status above.   Either of the failure should result in the final failure.

gdong1: From the original design, clear fwu flag is part of the EndFirmwareUpdate from platform.
do we have the case for the core (FirmwareUpdate driver) to only call one of them? want to understand why we need have API ClearFwUpdateTrigger () for the core driver.

mauricema: ClearFwUpdateTrigger() will need to be done at the beginning of the FWU flow as well after the sticky register is detected and translated into the state machine.  So it needs to be called from core side.

raghavag: Since ClearFwUpdateTrigger does not potentially return an error, changed return from status to VOID.

PR #583, opened 2020-03-06
 PSD Fixed memory overwrite issue.

pchand20:
platform security discovery driver having bug it was wrongly doing
  pointer arithmetic and causing memory overwrite. this may cause intermittent
  fault.

  removing ZeroMem call since its not required, ACPI table is already zero memory allocated.

Signed-off-by: Prakash Chandra <prakash1.chandra@intel.com>


pchand20:
seems something wrong with this commit. will Initiate another one.

bnevis-i:
@myguitar Please see https://github.com/slimbootloader/slimbootloader/pull/585

PR #584, opened 2020-03-06
[CFL] SgxLib moved to CFL Silicon lib

jgutbub:
Since SgxLib has a dependency on platform
specific FSP header files it should not be
used as a common platform library. This
patch preserves the same functionality as
before but just moves the library to be
platform specific.


myguitar:
This has been already merged, but SgxLib must not use UPDs directly as a Library. This is invalid usage, so SgxLib must be fixed.

PR #591, opened 2020-03-17
Add DMA protection in core code

mauricema:
This patch added DMA memory type into memory allocation pool for payloads.
This DMA memory buffer with PcdDmaBufferSize is located at address
aligned at PcdDmaBufferAlignment after Payload reserved memory. Memory
type EfiRuntimeServicesData is used to indicate DMA memory type.

Stage1B calculates the DMA memory location using fixed PCDs so that
platform can set up DMA protection as early as possible after memory is
ready. In Stage1B or Stage2 platform code should use platform VTd
information to setup PMR to protect all low memory except for the DMA
buffer range. DMA memory will be added into memory pool at the entry
point of the payload. Before transfering to OS, the DMA memory protection
can be disabled, and the DMA memory pool can be reclaimed for OS usage.

Currently only boot media device will utilize the DMA buffer range for
block access operations. So it should only be required by payloads. GFX,
when enabled, will also use DMA. It will be targeted to the system stolen
memory which is not protected by PMR.

### DIFF
--- a/PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
+++ b/PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
@@ -69,6 +69,8 @@ PayloadInit (
   // +--------------------------------------------+ RsvdBase + RsvdSize
   // |   Reserved memory for Payload              |
   // +--------------------------------------------+ RsvdBase
+  // |   + DMA buffer                             |
+  // +--------------------------------------------+ DmaBase
   // |   + Payload heap                           |
   // +--------------------------------------------+ HeapBase
   // |   + Payload stack                          |
--- a/PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
+++ b/PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
@@ -50,13 +50,15 @@ PayloadInit (
   UINT32                    HeapSize;
   UINT64                    RsvdBase;
   UINT64                    RsvdSize;
+  UINT32                    DmaBase;
+  UINT32                    DmaSize;
   UINT32                    StackBase;
   UINT32                    StackSize;
   LOADER_PLATFORM_DATA      *LoaderPlatformData;
   EFI_STATUS                PcdStatus1;
   EFI_STATUS                PcdStatus2;
   CONTAINER_LIST            *ContainerList;
-  EFI_MEMORY_RANGE_ENTRY    MemoryRanges[2];
+  EFI_MEMORY_RANGE_ENTRY    MemoryRanges[3];
 
   PcdStatus1 = PcdSet32S (PcdPayloadHobList, (UINT32)HobList);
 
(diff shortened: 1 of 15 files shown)


gdong1:
Do we have plan to support UEFI payload with DMA protection? not sure if UEFI BIOS has any method to do it.

mauricema:
For UEFI payload, we can do it as next step. SBL should work with UEFI payload well regarding to the DMA protection.  We will do SBL first, later on do UEFI payload.

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
```diff
@@ -78,20 +80,32 @@ PayloadInit (
...
   ASSERT ((RsvdBase & EFI_PAGE_MASK) == 0);
 
+  if (FeaturePcdGet (PcdDmaProtectionEnabled)) {
+    DmaSize = ALIGN_UP (PcdGet32 (PcdDmaBufferSize), EFI_PAGE_SIZE);
+  } else {
+    DmaSize = 0;
+  }
+  DmaBase  = (UINT32)RsvdBase - DmaSize;
+  DmaBase  = ALIGN_DOWN(DmaBase, PcdGet32 (PcdDmaBufferAlignment));
+
```
gdong1: do we need update SplitMemroyMap() in the Stage2Support.c to add DMA memory range in the memory map, so that all the payload could get DMA memory range from memory map?

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
```diff
@@ -78,20 +80,32 @@ PayloadInit (
   GetPayloadReservedRamRegion (&RsvdBase, &RsvdSize);
   ASSERT ((RsvdBase & EFI_PAGE_MASK) == 0);
 
+  if (FeaturePcdGet (PcdDmaProtectionEnabled)) {
```
gdong1: Need update comments in line 69 to match new changes.

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
```diff
@@ -78,20 +80,32 @@ PayloadInit (
...
   MemoryRanges[0].NumberOfPages = EFI_SIZE_TO_PAGES (HeapSize);
   MemoryRanges[0].Type          = EfiBootServicesData;
   MemoryRanges[1].BaseAddress   = RsvdBase;
   MemoryRanges[1].NumberOfPages = EFI_SIZE_TO_PAGES ((UINT32)RsvdSize);
   MemoryRanges[1].Type          = EfiReservedMemoryType;
-  AddMemoryResourceRange (MemoryRanges, 2);
+  MemoryRanges[2].BaseAddress   = DmaBase;
+  MemoryRanges[2].NumberOfPages = EFI_SIZE_TO_PAGES (DmaSize);
+  MemoryRanges[2].Type          = EfiRuntimeServicesData;
+  AddMemoryResourceRange (MemoryRanges, 3);
```
gdong1: if SplitMemroyMap() is updated, I think we don't need update it here in the payload.

Silicon/CommonSocPkg/Library/VtdPmrLib/IntelVtdPmr.c
```diff
@@ -22,6 +23,7 @@
   @return protected low memory alignment.
 **/
 UINT32
+STATIC
```
gdong1: For internal function, I remember  it is not necessary to add "STATIC" although I don't know if really have difference.

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
mauricema: DMA memory is free memory to OS and we don't reserve it.  So if we report it in memory type, which type to use ?
From my perspective, payload should not need to know where is the DMA memory base/size since they will just use AllocateRuntimePages/Pools to get it.

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
mauricema: Good point,  let me update that.

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
mauricema: This is still required.  It needs to be added into memory pool as a different memory type.

Silicon/CommonSocPkg/Library/VtdPmrLib/IntelVtdPmr.c
mauricema: It is recommended to use STATIC for internal function to avoid potential conflicts on same function names in different libs.

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
gdong1: this is my thought to avoid to build another HOB to payload:
Memory map was updated in SBL:
If Stage2 directly boots OS, the DMA size should be set to 0.
If using UEFI payload to boot OS, the SblParseLib could find DMA memory range from memory map to use for DMA later. And convert to free memory after use.
If using Osloader, it could do similar with UEFI payload.

gdong1: After chat, we know UEFI payload will use page table to protect DMA which is different SBL methods. so let's use this method firstly in SBL.

PayloadPkg/Library/PayloadEntryLib/PayloadEntryLib.c
mauricema: updated patch

PR #595, opened 2020-03-18
Add IoMmu support for XHCI library

mauricema:
This patch synced up to the latest EDKII XHCI library and then
added support for IoMmu interfaces. This will allow the library
to use DMA buffer for I/O transactions.

### DIFF
--- a/BootloaderCommonPkg/Library/XhciLib/UsbHcMem.h
+++ b/BootloaderCommonPkg/Library/XhciLib/UsbHcMem.h
@@ -1,7 +1,7 @@
 /** @file
 Private Header file for Usb Host Controller PEIM
 
-Copyright (c) 2014 - 2017, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2014, Intel Corporation. All rights reserved.<BR>
 
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
@@ -11,6 +11,9 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define _EFI_PEI_XHCI_MEM_H_
 
 #include <PiPei.h>
+
+#define  PeiServicesAllocatePages(a, b, c)   \
+         (((*(c) = (UINTN)AllocatePages (b)) != 0) ? EFI_SUCCESS : EFI_OUT_OF_RESOURCES)
 
 #define USBHC_MEM_DEFAULT_PAGES 16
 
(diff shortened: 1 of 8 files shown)


BootloaderCommonPkg/Library/XhciLib/UsbHcMem.h
```diff
@@ -1,7 +1,7 @@
 /** @file
 Private Header file for Usb Host Controller PEIM
 
-Copyright (c) 2014 - 2017, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2014, Intel Corporation. All rights reserved.<BR>
```
gdong1: This change looks not right, please check other files too.

BootloaderCommonPkg/Library/XhciLib/XhcPeim.h
```diff
@@ -20,9 +20,10 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/TimerLib.h>
 #include <Library/IoLib.h>
 #include <Library/MemoryAllocationLib.h>
+#include <Library/IoMmuLib.h>
 
-#define  PeiServicesAllocatePages(a, b, c)   \
-         (((*(c) = (UINTN)AllocatePages (b)) != 0) ? EFI_SUCCESS : EFI_OUT_OF_RESOURCES)
+#undef   DEBUG_INFO
+#define  DEBUG_INFO   DEBUG_VERBOSE
```
gdong1: Is this for debug purpose? I saw several places changing code to DEBUG_INFO from DEBUG_VERBOSE

BootloaderCommonPkg/Library/XhciLib/UsbHcMem.h
mauricema: This was from EDK2 without any change.

BootloaderCommonPkg/Library/XhciLib/XhcPeim.h
mauricema: Ahh, yes. It was for debug and should be removed.  Good catch!

mauricema: Most of the code was directly synced from EDK2.

mauricema:
pushed a new patch to address the comments.

BootloaderCommonPkg/Library/XhciLib/XhcPeim.h
mauricema: sorry, I forgot that I added these code to remove too much debug message in XHCI driver.  I did not try to change the debug level one by one since that is the original EDK2 code.  To address this, we can refine the INFO to VERBOSE level.  And only show error or warning messages.

PR #609, opened 2020-03-25
Add SBL x64 build for VTF

mauricema:
This an initial patch to add x64 build for SBL.  A new build flag
'-x64' is added to indicate x64 arch build. This cannot be fully
used at this moment because it has many dependencies on other x64
libraries. Only VTF reset vector x64 build is tested.

VTF x64 flow is different from IA32.  It switches to 32 bit
mode as usual and then calls into FspTempRamInit to set up CAR.
Once CAR is ready, it builds 4GB identical mapping page table for
x64 and then switches to x64 long mode. Finally, it locates the
STAGE1A entry point and tranfers the control to STAGE1A in pure
64 bit mode.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -269,6 +269,7 @@ class Build(object):
         self._workspace                    = os.environ['WORKSPACE']
         self._board                        = board
         self._image                        = "SlimBootloader.bin"
+        self._arch                         = board.BUILD_ARCH
         self._target                       = 'RELEASE' if board.RELEASE_MODE  else 'NOOPT' if board.NO_OPT_MODE else 'DEBUG'
         self._fsp_basename                 = 'FspDbg'  if board.FSPDEBUG_MODE else 'FspRel'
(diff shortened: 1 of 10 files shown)


BuildLoader.py
```diff
@@ -1279,6 +1289,7 @@ def cmd_build(args):
     buildp.add_argument('-v',  '--usever',  action='store_true', help='Use board version file')
     buildp.add_argument('-fp', dest='fsppath', type=str, help='FSP binary path relative to FspBin in Silicon folder', default='')
     buildp.add_argument('-fd', '--fspdebug', action='store_true', help='Use debug FSP binary')
+    buildp.add_argument('-x64','--x64',  action='store_true', help='Build X64 image. Default is to build IA32 image.')
```
myguitar: To consider other architecture later, string argument would be extensible. 
buildp.add_argument('-a',  '--arch' , dest='arch', type=str, help='IA32, X64', default ='IA32')

mauricema: very valid point, let me change that.

mauricema:
Pushed a new commit to address the review comments.

PR #610, opened 2020-03-26
Add X64 MdePkg Library

myguitar:
This adds necessary X64 Library from MdePkg
- Copied from EDK2 Stable201911


MdePkg/Library/BaseIoLibIntrinsic/X64/IoFifoSev.nasm
```diff
@@ -0,0 +1,282 @@
+;------------------------------------------------------------------------------
```
mauricema: This file is not required. It is for AMD SEV implementation.

myguitar: Let me remove this. Thanks.

PR #615, opened 2020-03-26
Pointer type cast for both 32/64-bit operation

myguitar:
This patch allows both 32/64-bit addressing properly.
- Pointer type cast with UINTN
- Add missing EFIAPI for APIs


myguitar:
> Thank you for taking time to make it x64 compiling ready.

No problem. I will send additional PRs to make X64 target buildable on to 64bit_support branch.

PR #617, opened 2020-03-27
Fix pointer type cast errors from Visual Studio

myguitar:
Visual Studio reports more pointer type cast errors with 64-bit build.
This will cover the issue on the existing targets.


BootloaderCommonPkg/Library/Lz4DecompressLib/Lz4DecompressLib.c
```diff
@@ -582,7 +582,7 @@ Lz4Decompress (
   DecompressedSize  = * (UINT32 *)Source;
   Source = (VOID *) ((CHAR8 *)Source + sizeof (UINT32));
   SourceSize -= sizeof (UINT32);
-  Size = (UINT32)LZ4_decompress_safe (Source, Destination, SourceSize, DecompressedSize);
+  Size = (UINT32)LZ4_decompress_safe (Source, Destination, (int)SourceSize, DecompressedSize);
```
mauricema: (INT32) ?

myguitar: Thanks. Updated as commented.

PR #618, opened 2020-03-27
Add support for 64bit compiling map file

mauricema:
This patch enhanced the map format parsing for map files generated
by x64 MSFT compiler.


myguitar:
Let me take a look Travis.

myguitar:
It looks there is a failure - https://travis-ci.org/github/slimbootloader/slimbootloader/builds/667807120.
Travis does not post the result.

mauricema:
Thx!  Let me take a look!

mauricema:
Yes, it was GCC map file issue.  Updated the patch to deal with MSFT compiler only at this moment.

PR #625, opened 2020-04-01
Move stage common code into StageLib

mauricema:
There are lots of common code in Stage1A/1B/2, so it makes sense
to put those into a StageLib so that the same code can be reused.
As part of it, PagingLib is also restructured to align with Stage
mapping/remapping. It will consider both X64 mode and IA32 mode.


myguitar:
Need to rebase to resolve a conflict.

mauricema:
Thanks.  Rebased and addressed the conflicts.

PR #640, opened 2020-04-05
RSA PSS verification scheme

lsubash:
This patch adds support for RSA PSS verification.
Ported IPP Crypto for PSS from latest IPP github repo.
Secure boot lib and RSA wrappers functions are
updated. RSA verfication sheme is would be
based on SigType in Signature Header.

### DIFF
--- a/BootloaderCommonPkg/Include/Library/CryptoLib.h
+++ b/BootloaderCommonPkg/Include/Library/CryptoLib.h
@@ -1,7 +1,7 @@
 /** @file
   Provides sha256 and RSA2048 verify functions.
 
-Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2017-2020, Intel Corporation. All rights reserved.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -57,6 +57,9 @@ typedef UINT8 HASH_ALG_TYPE;
 #define IPP_HASHLIB_SHA2_384             0x0004
 #define IPP_HASHLIB_SHA2_512             0x0008
 #define IPP_HASHLIB_SM3_256              0x0010
+
+#define IPP_RSALIB_PKCS_1_5              0x0001
+#define IPP_RSALIB_PSS                   0x0002
 
 
 typedef UINT8 HASH_CTX[IPP_HASH_CTX_SIZE];   //IPP Hash context buffer
@@ -171,11 +174,9 @@ Sm3 (
 /**
   Verifies the RSA signature with PKCS1-v1_5 encoding scheme defined in RSA PKCS#1.
 
-  @param[in]  PubKey         Pointer to a pre-processed RSA key data.
-  @param[in]  Signature      Pointer to RSA PKCS1-v1_5 signature to be verified.
-  @param[in]  SignatureLen   Length of the signature in bytes.
-  @param[in]  SignatureType  Now only support signature type SIG_TYPE_RSA2048SHA256.
-  @param[in]  Hash           Pointer to octet message hash to be checked.
+  @param[in]  PubKeyHdr         Pointer to a PubKey data.
+  @param[in]  SignatureHdr      Pointer to signature data to be verified.
+  @param[in]  Hash              Pointer to octet message hash to be checked.
 
   @retval  RETURN_SUCCESS             Valid signature.
   @retval  RETURN_INVALID_PARAMETER   Key or signature format is incorrect.
@@ -183,11 +184,37 @@ Sm3 (
 
 **/
 RETURN_STATUS
+EFIAPI
 RsaVerify_Pkcs_1_5 (
-  CONST PUB_KEY_HDR        *PubKeyHdr,
-  CONST SIGNATURE_HDR      *SignatureHdr,
-  CONST UINT8              *Hash
+  IN CONST PUB_KEY_HDR        *PubKeyHdr,
+  IN CONST SIGNATURE_HDR      *SignatureHdr,
+  IN CONST UINT8              *Hash
   );
+
+
+/**
+  Verifies the RSA signature with PSS encoding scheme defined in RSA PSS.
+
+  @param[in]  PubKeyHdr         Pointer to a PubKey data.
+  @param[in]  SignatureHdr      Pointer to signature data to be verified.
+  @param[in]  Src               Pointer to meassage.
+  @param[in]  SrcSize           Size of the message.
+
+  @retval  RETURN_SUCCESS             Valid signature.
+  @retval  RETURN_INVALID_PARAMETER   Key or signature format is incorrect.
+  @retval  RETURN_SECURITY_VIOLATION  Invalid signature.
+
+**/
+
+RETURN_STATUS
+EFIAPI
+RsaVerify_PSS (
+  IN CONST PUB_KEY_HDR        *PubKeyHdr,
+  IN CONST SIGNATURE_HDR      *SignatureHdr,
+  IN CONST UINT8              *Src,
+  IN CONST UINT32             SrcSize
+  );
+
 
 /**
   Computes the HMAC SHA-256 message digest of a input data buffer.
--- a/BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
+++ b/BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2018, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2018-2020, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -15,10 +15,10 @@
 #include <Library/CryptoLib.h>
 #include <Library/BlMemoryAllocationLib.h>
 
-/* Wrapper function for RSAVerify to make the inferface consistent.
+/* Wrapper function for RSA PKCS_1.5 Verify to make the inferface consistent.
  * Returns non-zero on failure, 0 on success.
  */
-int VerifyRsaSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *SignatureHdr,  CONST UINT8  *Hash)
+int VerifyRsaPkcs1Signature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *SignatureHdr,  CONST UINT8  *Hash)
 {
   int    sz_n;
   int    sz_e;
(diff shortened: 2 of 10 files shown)


BootloaderCommonPkg/Include/Library/CryptoLib.h
```diff
@@ -189,6 +187,30 @@ RsaVerify_Pkcs_1_5 (
...
+  @param[in]  SrcSize           Size of the message.
+
+  @retval  RETURN_SUCCESS             Valid signature.
+  @retval  RETURN_INVALID_PARAMETER   Key or signature format is incorrect.
+  @retval  RETURN_SECURITY_VIOLATION  Invalid signature.
+
+**/
+
+RETURN_STATUS
+RsaVerify_PSS (
```
mauricema, line 206: Should we consider using generic name for function ?    EX:  SignatureVerify (...)
Other info should be indicated in the parameter, PUB_KEY_HDAR and SIGNATURE_HDR.

lsubash:
Sizes calculated for recent platform release build:
Without this path,
Stage1A - 0x5170
Stage1B - 0xdf48
Stage2 - 0x58ea8

With PSS patch,
Stage1A - 0x5170
Stage1B - 0xe688
Stage2 - 0x595a8

Increase in FV sizes for stage 1B is **1856** bytes and Stage2 is **1792** bytes

mauricema:
Thanks!
Can we still use PCD to control the compiler so that size will not be increased ?

lsubash:
> Can we still use PCD to control the compiler so that size will not be increased ?
We would need to support both signing schemes. For example, we would set RSA_PSS config in our signing tools. Still, we need to support boot images signed with PCKS1. Eg: Booting PKSC1 yocto image.

mauricema:
Yes, understood.  I think it is just very similar to PcdIppHashLibSupportedMask.  It is possible to enable multiple.

lsubash:
> Yes, understood. I think it is just very similar to PcdIppHashLibSupportedMask. It is possible to enable multiple.

Yes, I will add Pcd flag support and enable both schemes by default.

BootloaderCommonPkg/Include/Library/CryptoLib.h
```diff
@@ -171,11 +171,9 @@ Sm3 (
 /**
   Verifies the RSA signature with PKCS1-v1_5 encoding scheme defined in RSA PKCS#1.
 
-  @param[in]  PubKey         Pointer to a pre-processed RSA key data.
-  @param[in]  Signature      Pointer to RSA PKCS1-v1_5 signature to be verified.
-  @param[in]  SignatureLen   Length of the signature in bytes.
-  @param[in]  SignatureType  Now only support signature type SIG_TYPE_RSA2048SHA256.
-  @param[in]  Hash           Pointer to octet message hash to be checked.
+  @param[in]  PubKeyHdr         Pointer to a PubKey data.
```
myguitar, line 174: Can you also add IN in function parameters?

BootloaderCommonPkg/Include/Library/CryptoLib.h
```diff
@@ -189,6 +187,30 @@ RsaVerify_Pkcs_1_5 (
...
+  @retval  RETURN_SECURITY_VIOLATION  Invalid signature.
+
+**/
+
+RETURN_STATUS
+RsaVerify_PSS (
+  CONST PUB_KEY_HDR        *PubKeyHdr,
+  CONST SIGNATURE_HDR      *SignatureHdr,
+  CONST UINT8              *Src,
+  CONST UINT32             SrcSize
```
myguitar, line 210: Same here. Please add IN in function parameters.

BootloaderCommonPkg/Include/Library/CryptoLib.h
```diff
@@ -171,11 +171,9 @@ Sm3 (
...
-  @param[in]  Signature      Pointer to RSA PKCS1-v1_5 signature to be verified.
-  @param[in]  SignatureLen   Length of the signature in bytes.
-  @param[in]  SignatureType  Now only support signature type SIG_TYPE_RSA2048SHA256.
-  @param[in]  Hash           Pointer to octet message hash to be checked.
+  @param[in]  PubKeyHdr         Pointer to a PubKey data.
+  @param[in]  SignatureHdr      Pointer to signature data to be verified.
+  @param[in]  Hash              Pointer to octet message hash to be checked.
 
   @retval  RETURN_SUCCESS             Valid signature.
   @retval  RETURN_INVALID_PARAMETER   Key or signature format is incorrect.
```
myguitar, line 179: Can you add EFIAPI?

BootloaderCommonPkg/Include/Library/CryptoLib.h
```diff
@@ -189,6 +187,30 @@ RsaVerify_Pkcs_1_5 (
...
+  @param[in]  Src               Pointer to meassage.
+  @param[in]  SrcSize           Size of the message.
+
+  @retval  RETURN_SUCCESS             Valid signature.
+  @retval  RETURN_INVALID_PARAMETER   Key or signature format is incorrect.
+  @retval  RETURN_SECURITY_VIOLATION  Invalid signature.
+
+**/
+
+RETURN_STATUS
```
myguitar, line 205: Can you add EFIAPI?

BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
```diff
@@ -1,7 +1,7 @@
 /** @file
   Secure boot library routines to provide RSA signature verification.
 
-  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017-2019, Intel Corporation. All rights reserved.<BR>
```
myguitar, line 4: 2020?

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
```diff
@@ -157,6 +286,23 @@ RsaVerify_Pkcs_1_5 (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *Signature
                   ((SignatureHdr->SigSize != RSA2048_MOD_SIZE) && (SignatureHdr->SigSize != RSA3072_MOD_SIZE))) {
     return RETURN_INVALID_PARAMETER;
   } else {
-    return VerifyRsaSignature (PubKeyHdr, SignatureHdr, Hash) ? RETURN_SECURITY_VIOLATION : RETURN_SUCCESS ;
+    return VerifyRsaPkcs1Signature (PubKeyHdr, SignatureHdr, Hash) ? RETURN_SECURITY_VIOLATION : RETURN_SUCCESS ;
```
myguitar, line 289: VerifyRsaPkcs1Signature returns int type. Better to check the return value with ippStsNoErr.

BootloaderCommonPkg/Include/Library/CryptoLib.h
lsubash: Thanks. I will add. 
Also they need to be added public Hashfunctions as SHA256 and so. Would do as separate patch

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
lsubash: ippStsNoErr defitions are limited to IPP Crypto lib. 
ippStsNoErr would be same as RETURN_SUCCESS

myguitar: Understood. It returns int type, but the if condition looks like BOOLEAN. Assigning it to EFI_STATUS would be better.

lsubash:
Pushed another version of patch fixing review comments.

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
lsubash: Aiden, I would address in another patch,

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
```diff
@@ -146,17 +146,173 @@ int VerifyRsaSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *Signa
...
+  if (FixedPcdGet8(PcdCompSignSchemeSupportedMask) & IPP_RSALIB_PSS) {
+    if ((SignatureHdr->SigType != SIGNING_TYPE_RSA_PSS) ||
+                    ((SignatureHdr->SigSize != RSA2048_MOD_SIZE) && (SignatureHdr->SigSize != RSA3072_MOD_SIZE))) {
+      return RETURN_INVALID_PARAMETER;
+    } else {
+      return VerifyRsaPssSignature (PubKeyHdr, SignatureHdr, Src, SrcSize) ? RETURN_SECURITY_VIOLATION : RETURN_SUCCESS ;
+    }
   } else {
-    return VerifyRsaSignature (PubKeyHdr, SignatureHdr, Hash) ? RETURN_SECURITY_VIOLATION : RETURN_SUCCESS ;
+      return RETURN_UNSUPPORTED;
```
myguitar, line 315: indent

lsubash:
> A minor indent issue. Other than that, looks good to me.



> A minor indent issue. Other than that, looks good to me.

Done.

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
```diff
@@ -146,17 +146,173 @@ int VerifyRsaSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *Signa
 }
 
 
-/* Wrapper function for RSAVerify to make the inferface consistent.
+/* Wrapper function for RSA PSS verify to make the inferface consistent.
+ * Returns non-zero on failure, 0 on success.
+ */
+int VerifyRsaPssSignature (CONST PUB_KEY_HDR *PubKeyHdr, CONST SIGNATURE_HDR *SignatureHdr,  CONST UINT8  *Src, CONST UINT32  Size)
```
sagraw2, line 152: Have we done some negative testing for this new method?
What happens when we manually corrupt a PSS signature in the stitched SBL image?
Does the implementation gracefully returns failure and block the boot?

BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
```diff
@@ -71,21 +71,36 @@ DoRsaVerify (
...
+
     Status = RsaVerify_Pkcs_1_5 (PublicKey, SignatureHdr, Digest);
-  } else {
+
+  } else if(SignatureHdr->SigType == SIGNING_TYPE_RSA_PSS) {
+
+    // Calculate Hash only when OutHash is valid
+    // RSA PSS requires to pass message to be verified
+    if (OutHash != NULL) {
+      Status = CalculateHash  (Data, Length, SignatureHdr->HashAlg, Digest);
```
sagraw2, line 94: Hashing will be done twice for RSA-PSS. Do we know where it will impact most?
What are the large components which will follow this path and will experience twice hashing?

lsubash: Hashing in case of PSS would be done for  CfgDatablob and KeyHashManifest verification would be twice.  Rest of the calls as Container verification for Boot image, hash calculation would  be done only only in IPP.

BootloaderCommonPkg/Library/IppCryptoLib/rsa_verify.c
lsubash: Yes, checked with corrupting the CfgData and PSS verification fails with an graceful exit.
RSA verification used for ext components -
CFG Data, KeyhashManifest, external component loaded using container and Boot images.
Boot halts only when boot image verification fails. For rest of the components, boot would continue.  Corresponding features would fails.

sagraw2: Thanks Subash for running these tests.

BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
sagraw2: CfgDataBlob and KeyHashMan sizes are pretty small...so double hashing may not have a impact.
For kernel image verification, won't that happen twice...as we extend the measurement of kernel image in TPM?

lsubash: For Kernel Image using container, 
- Container header is RSA verified. While extending to TPM, hash is recalculated. (size of header is small)
- Kernel image (component in container) is hash verified.  When extended to TPM it uses the hash in Container -Component header. In cases as SM3, hash would be calculated for all TPM measurements as its not supported while signing.

For IAS image method, yes there would double calc. But its getting deprecated,

sagraw2: Sounds good.

PR #668, opened 2020-04-13
Fix XHCI library memory de-allocation issue

mauricema:
This patch added code to XHCI de-initialization funciton to free
all used memory.

### DIFF
--- a/BootloaderCommonPkg/Library/XhciLib/XhcPeim.c
+++ b/BootloaderCommonPkg/Library/XhciLib/XhcPeim.c
@@ -1392,34 +1392,46 @@ XhcPeiGetRootHubPortStatus (
   return EFI_SUCCESS;
 }
 
+
 /**
-  One notified function to stop the Host Controller at the end of PEI
+  Free the USB device context.
 
-  @param[in]  PeiServices        Pointer to PEI Services Table.
-  @param[in]  NotifyDescriptor   Pointer to the descriptor for the Notification event that
-                                 caused this function to execute.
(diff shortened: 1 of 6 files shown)


BootloaderCommonPkg/Library/XhciLib/XhcPeim.c
```diff
@@ -1392,34 +1392,46 @@ XhcPeiGetRootHubPortStatus (
...
+        UsbDevContext->ConfDesc = NULL;
+      }
 
-  return EFI_SUCCESS;
+      for (Index2 = 0; Index2 < 31; Index2++) {
+        if (UsbDevContext->EndpointTransferRing[Index2] != NULL) {
+          FreePool (UsbDevContext->EndpointTransferRing[Index2]);
+          UsbDevContext->EndpointTransferRing[Index2] = NULL;
+        }
+      }
```
myguitar, line 1432: Do we need to free UsbDevContext->EndpointTransferRing here?
```
FreePool (UsbDevContext->EndpointTransferRing);
UsbDevContext->EndpointTransferRing = NULL;
```

mauricema: I don't think we need.  EndpointTransferRing itself is defined as an array in USB_DEV_CONTEXT.  The space is allocated as part of the full structure USB_DEV_CONTEXT.

myguitar: You are right. It's static array. Thanks.

PR #675, opened 2020-04-15
Support 64-bit XHCI MMIO address

stalamudupula:
If Platform code assigns 64-bit BAR address to XHCI,
get the full 64-bit address to access MMIO space.
Behavior is undefined if building IA32 and assigning
64-bit XHCI resources.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -266,6 +266,7 @@ class BaseBoard(object):
         self.HASH_STORE_SIZE       = 0x200  #Hash store size to be allocated in bootloader
 
         self.PCI_MEM64_BASE        = 0
+        self.BUILD_ARCH            = 'IA32'
 
         for key, value in list(kwargs.items()):
             setattr(self, '%s' % key, value)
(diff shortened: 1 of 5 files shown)


BuildLoader.py
```diff
@@ -266,6 +266,7 @@ def __init__(self, *args, **kwargs):
         self.HASH_STORE_SIZE       = 0x200  #Hash store size to be allocated in bootloader
 
         self.PCI_MEM64_BASE        = 0
+        self.BUILD_ARCH            = 'IA32'
```
mauricema, line 269: do you need this.  BUILD_ARCH will be initialized in line 271-272.

stalamudupula: BUILD_ARCH is not available until cmd_build() routine and  I want to use BUILD_ARCH in BoardConfig.py to enforce 64-bit PCI resource allocation iff it is X64 build.

mauricema: OK.  Understood now.

PR #676, opened 2020-04-15
Simplified stitchifwi.py for CoffeLake

raghavag:
This patch simplified ingredient folder structure
required for stitchifwi. Also created a config file
with all the configuration in one file and this file
will now be input to stitchifwi

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Script/StitchIfwiConfig.py
+++ b/Platform/CoffeelakeBoardPkg/Script/StitchIfwiConfig.py
@@ -1,0 +1,168 @@
+## @ StitchIfwiConfig.py
+#  This is an IFWI stitching configuration script for Slim Bootloader
+#
+# Copyright (c) 2019 - 2020, Intel Corporation. All rights reserved.<BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+def get_bpmgen2_params_change_list ():
+    params_change_list = []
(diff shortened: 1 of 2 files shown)


Platform/CoffeelakeBoardPkg/Script/StitchIfwiConfig.py
```diff
@@ -0,0 +1,202 @@
+## @ StitchIfwi.py
+#  This is an IFWI stitch config script for Slim Bootloader
+#
+# Copyright (c) 2019, Intel Corporation. All rights reserved. <BR>
```
mauricema, line 4: copyright

Platform/CoffeelakeBoardPkg/Script/StitchIfwiConfig.py
```diff
@@ -0,0 +1,202 @@
+## @ StitchIfwi.py
+#  This is an IFWI stitch config script for Slim Bootloader
+#
+# Copyright (c) 2019, Intel Corporation. All rights reserved. <BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+extra_usage_txt = \
```
myguitar, line 9: Usage text should be here instead of StitchIfwi.py?

raghavag: okay I will move it back to stitchifwi

PR #677, opened 2020-04-15
Upgrade LZ4 to 1.7.4

mauricema:
This patch upgraded the LZ4 from 1.4.0 to 1.7.4. The size will
increase around 300 bytes. Performance is still very similar.
But when trying to use more recent LZ4 version 1.9, noticed
significant performance degration. So keep to 1.7.4 for now.


mauricema:
These are from original LZ4 github.  Only Lz4Decompress.c was re-orgnized.

PR #679, opened 2020-04-16
Add 'acpi_rsdp' Linux kernel parameter in cmdline

myguitar:
Recent Linux kernel accepts acpi_rsdp=0x.. in kernel command line.
This will make Linux kernel look for ACPI RSDP address in the kernel
commad line first prior to in DMI or F-segment.

### DIFF
--- a/BootloaderCommonPkg/Library/LinuxLib/LinuxLib.c
+++ b/BootloaderCommonPkg/Library/LinuxLib/LinuxLib.c
@@ -13,8 +13,12 @@
 #include <Library/HobLib.h>
 #include <Library/LinuxLib.h>
 #include <Library/PagingLib.h>
+#include <Library/PrintLib.h>
 #include <Guid/GraphicsInfoHob.h>
 #include <Guid/MemoryMapInfoGuid.h>
+#include <Guid/SystemTableInfoGuid.h>
+
+#define ACPI_RSDP_CMDLINE_STR         "acpi_rsdp="
 
 /**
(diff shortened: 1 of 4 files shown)


BootloaderCommonPkg/Library/LinuxLib/LinuxLib.c
```diff
@@ -314,6 +318,15 @@ UpdateLinuxBootParams (
     E820Entry++;
   }
   Bp->E820Entries = (UINT8)MapInfo->Count;
+
+  GuidHob = GetFirstGuidHob (&gLoaderSystemTableInfoGuid);
+  if (GuidHob != NULL) {
+    SystemTableInfo = (SYSTEM_TABLE_INFO *)GET_GUID_HOB_DATA (GuidHob);
+
+    AsciiSPrint (ParamValue, sizeof (ParamValue), " acpi_rsdp=0x%lx", SystemTableInfo->AcpiTableBase);
```
mauricema, line 326: Will this prevent user from overriding acpi_rsdp in command line in case they want ?

myguitar: You are right, this will prevent user option since kernel command parameters are parsed sequentially. To avoid this, let me add acpi_rsdp only if the kernel command line does not have it.

myguitar: Updated to check "acpi_rsdp=" existence before appending it.

PR #681, opened 2020-04-17
Add Converged BootGuard Library to Soc common

lsubash:
Currently BootGuard library is maintained for     every supported silicon. Most of Slimboot supported     platforms support CBnT standard. Adding to SOC common  for platforms to consume.

Rename BootGuardLib header in APL.  APL follows BootGuard 2.0 and other supported platforms
follows CBnT standad. Its conflicting with CBnT header.

### DIFF
--- a/Platform/ApollolakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
+++ b/Platform/ApollolakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
@@ -11,7 +11,7 @@
 #include <Library/GpioLib.h>
 #include <Library/SerialPortLib.h>
 #include <Library/PlatformHookLib.h>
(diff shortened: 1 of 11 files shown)


myguitar:
What about CFL/WHL? Already using the common lib?

lsubash:
> What about CFL/WHL? Already using the common lib?
In CFL/WHL , its separate BTG and TXT ACM's.
It follows Btg 1.0 (legacy). 

CBnT added in Commonlib is for platforms later than CFL.
--

lsubash:
> re you moving the files from CoffeeLakeSoc folder or adding new file

Moving from recent platform CBnT lib(SblInternal) to SOC Common.
Once this patch gets merged, current & upcoming platforms would be cleaned up and consume common lib. 

CoffeeLake & APL platforms would maintain their own BootGuardLib's as its different from CBnT lib.

PR #687, opened 2020-04-20
Support 64-bit ELF loading

myguitar:
This will load and execute 64-bit ELF image.
- Load image from ELF program header
- 32-bit ELF on IA32 only. 64-bit ELF on X64 only
- TBD: Relocate ELF

### DIFF
--- a/BootloaderCommonPkg/Library/ElfLib/ElfLib.inf
+++ b/BootloaderCommonPkg/Library/ElfLib/ElfLib.inf
@@ -22,6 +22,8 @@
 [Sources]
   ElfCommon.h
   Elf32.h
+  Elf64.h
+  ElfLibInternal.h
   ElfLib.c
 
 [Packages]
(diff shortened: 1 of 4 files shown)


BootloaderCommonPkg/Library/ElfLib/ElfLib.inf
```diff
@@ -22,8 +22,18 @@
 [Sources]
   ElfCommon.h
   Elf32.h
+  Elf64.h
+  ElfLibInternal.h
   ElfLib.c
 
+[Sources.IA32]
```
mauricema, line 29: Do you think we can make it simpler to just use a same for both IA32 and X64 ?
If required,  IS_X64 can be used to determine x86 or x64 build.  The compiler will remove unused code if size is the concern.

myguitar: Sure, let me make it common for both IA32 and X64.

PR #688, opened 2020-04-21
Use PcdCompSignHash alg for pubkey verification for non zero usage

lsubash:
PublicKey hashes stored in HashStore use hash alg type of
PcdCompSignHash defined with Build config. In container we support cases where hash type
could differ from Sbl signing hash. For example Slimboot could
be configured to RSA3K and SHA384 at build. Container Boot image
could happen to be signed with RSA2K and SHA256.

Assigning pubkey verification hash type to PcdCompSignHash
for valid USAGE cases.

### DIFF
--- a/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
+++ b/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
@@ -288,6 +288,7 @@ AuthenticateComponent (
   UINT8                    *SigPtr;
   UINT8                    *KeyPtr;
   SIGNATURE_HDR            *SignHdr;
+  UINT8                     HashAlg;
 
   if (!FeaturePcdGet (PcdVerifiedBootEnabled)) {
     Status = EFI_SUCCESS;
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -301,8 +302,15 @@ AuthenticateComponent (
       SigPtr   = (UINT8 *) AuthData;
       SignHdr  = (SIGNATURE_HDR *) SigPtr;
       KeyPtr   = (UINT8 *)SignHdr + sizeof(SIGNATURE_HDR) + SignHdr->SigSize ;
+
+      if (Usage == 0x0) {
+        HashAlg =  GetHashAlg(AuthType);
+      }  else {
+        HashAlg =  PcdGet8(PcdCompSignHashAlg);
```
mauricema, line 309: If  GetHashAlg(AuthType) is different from  PcdCompSignHahsAlg in this case, should we just return failure instead ?  
In what condition, this two will not match if  (Usage != 0)  ?

lsubash: For example,
- Consider SBL is configured with SHA384 signing hash. Hash Store would have pub key hashes with SHA384 alg. (Though we have a capability to support multiple hash alg in HashStore data structures, we do not have functionality in place to generate multiple hashes)
- In GetKeyHashList() for external KEYH, Configured RSA3K for all components except for Boot image.
- Yocto boot image container is generated with RSA2048_SHA256 is used to boot.
- Container Hash Alg (GetHashAlg(AuthType) would be SHA256 where as PcdCompSignHashAlg would be SHA384. In this case Usage!=0.

When Usage is 0, we are providing Public Key hash and hash alg would be same as authtype,

lsubash: Currently Yocto images supports only RSA2048_SHA256 container format. Also, same case is applicable for external components as TSN signinh hash  differs from SBL signing hash configured.

mauricema: My question is more on what will not work if we always use GetHashAlg(AuthType) ?
For container case, the AuthType indicated in the container header can be used to dertermine hash, right ?  In the example you give,  the authentication will fail due to inconsistent hash length.

lsubash:
Instead of changing in SBL, modified on tools to adjust the hash algorithm in external KeyHashStore manifest based on key size. Use SHA256 for size 2K and SHA384 for 3K.

https://github.com/slimbootloader/slimbootloader/pull/690

PR #690, opened 2020-04-23
Adjust hash algorithm used for KEYH based on key size

lsubash:
PublicKey hashes stored in HashStore use hash alg type of
PcdCompSignHash defined with Build config. In container we
support cases where hash type could differ from Sbl default
signing hash.

Adjust the hash algorithm in external KeyHashStore manifest
based on key size. Use SHA256 for size 2K and SHA384 for 3K.

### DIFF
--- a/BootloaderCorePkg/Tools/CommonUtility.py
+++ b/BootloaderCorePkg/Tools/CommonUtility.py
@@ -276,8 +276,14 @@ def rsa_sign_file (priv_key, pub_key, hash_type, sign_scheme, in_file, out_file,
         gen_file_from_object (out_file, bins)
 
 def get_key_type (in_key):
-    pub_key = gen_pub_key (in_key)
-    pub_key_hdr = PUB_KEY_HDR.from_buffer(pub_key)
+
+    # Check for public key in binary format.
+    key = bytearray(get_file_data(in_key))
+    pub_key_hdr = PUB_KEY_HDR.from_buffer(key)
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -275,8 +275,13 @@ def rsa_sign_file (priv_key, pub_key, hash_type, sign_scheme, in_file, out_file,
     if len(bins) != len(out_data):
         gen_file_from_object (out_file, bins)
 
-def get_key_type (in_key):
-    pub_key = gen_pub_key (in_key)
+def get_key_type (in_priv_key, in_pub_key = None):
```
mauricema, line 278: how about using get_key_type (in_key) ?
in_key can be  PEM  private or public key file or binary file for public key.
if in_key.endswith('.pem') then  do  gen_pub_key(in_key),  else do the file read.

lsubash: I think there should not be restriction for private key file end with pem format.
Alternatively we can maintain separate function getting key size from public key

mauricema: in_priv_key is not accurate since it could be private or public key. both are supported.

lsubash: Thanks. Added check Public key identifier for public key in bin format.
Updated PR.

Issue #136, opened 2019-04-12
PatchFv.py does not display usage properly.


mauricema:
Run  "python PatchFv.py" will display:
Traceback (most recent call last):
  File "W:\SlimBoot\SlimFork\BootloaderCorePkg\Tools\PatchFv.py", line 951, in <module>
    sys.exit(main())
  File "W:\SlimBoot\SlimFork\BootloaderCorePkg\Tools\PatchFv.py", line 852, in main
    Usage()
NameError: global name 'Usage' is not defined

PR #692, opened 2020-04-24
Support paging for Above4Gb addresses

stalamudupula:
This patch enhances MapMemoryRegion subroutine to
add PDP entries for mapping addresses > 4GiB.
Only 1:1 mapping is provided for Above4Gb addresses.
And linear addresses are mapped to 1GiB pages.

### DIFF
--- a/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
+++ b/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
@@ -11,19 +11,27 @@
 #include <Library/MemoryAllocationLib.h>
 #include <Library/PagingLib.h>
 
+#define PDE_PG_LVL  2
+#define PDP_PG_LVL  3
 
 /**
   Get shift bit number from address to page index
 
-  @retval    Bit to shift to get page index.
+  @param[in] Level    Page level.
+
+  @retval             Bit to shift to get specific page level index.
 
 **/
 UINT32
 GetPageShiftBits (
-  VOID
+  IN  UINT8   Level
 )
 {
-  return IsLongModeEnabled() ? 21 : 22;
+  if (Level == PDE_PG_LVL) {
+    return IsLongModeEnabled() ? 21 : 22;
+  } else {
+    return IsLongModeEnabled() ? 30 : 31;
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -47,13 +106,20 @@ MapMemoryRange (
...
+  UINTN          Address;
   UINT32         PteOffset;
   UINT32         PdeOffset;
-  UINT32         Alignment;
+  UINTN          Alignment;
   UINT32         PageBits;
+  EFI_STATUS     Status;
+
+  if (Ranges[0].Start >= 0x180000000ULL) {
+    DEBUG ((DEBUG_INFO, "SAI : Calling PdpEntryadd\n"));
```
mauricema, line 119: Remove this ??

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -47,13 +106,20 @@ MapMemoryRange (
...
-  UINT32         Address;
+  UINTN          Address;
   UINT32         PteOffset;
   UINT32         PdeOffset;
-  UINT32         Alignment;
+  UINTN          Alignment;
   UINT32         PageBits;
+  EFI_STATUS     Status;
+
+  if (Ranges[0].Start >= 0x180000000ULL) {
```
mauricema, line 118: Why do we do this ?
I think we can always check to see if the entry is valid for a given address.  If it is not there, then we can try to add entry.

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -20,10 +22,67 @@
...
-  return IsLongModeEnabled() ? 21 : 22;
+  if (Level == PDE_PG_LVL) {
+    return IsLongModeEnabled() ? 21 : 22;
+  } else {
+    return IsLongModeEnabled() ? 30 : 31;
+  }
+}
+
+EFI_STATUS
+AddPdpEntry (
```
mauricema, line 36: Add function description

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -20,10 +22,67 @@
...
+  UINT32        PdpOffset;
+  UINT32        Attribute;
+  UINTN         EntryNum;
+  UINTN         Alignment;
+  UINTN         PageBase;
+
+  Attribute    = IA32_PG_P | IA32_PG_RW | IA32_PG_AC;
+
+  ASSERT (Ranges[0].Start == Ranges[0].Mapping);
+  if (((Ranges[0].Start & 0xFFF) != 0) || ((Ranges[0].Limit & 0xFFF) != 0xFFF)) {
```
mauricema, line 55: since you are doing 1GB page, so your start and limit should be 1GB aligned.

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
mauricema: And the function name needs to be more specific
Maybe  MapAddressByHugePage () ?

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -20,10 +22,67 @@
...
+  PageBase    = Ranges[0].Start & ~Alignment;
+  if (((Ranges[0].Start & ~Alignment) != PageBase) || ((Ranges[0].Limit & ~Alignment) != PageBase)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Address     = PageBase;
+  PdpOffset   = 1 * EFI_PAGE_SIZE;
+  PageTable1G = (UINTN *)((UINTN)PageBuffer + PdpOffset);
+  Idx = (UINT32) RShiftU64 (PageBase, PageBits);
+  EntryNum    = 512;
```
mauricema, line 70: why hard code to 512 ?

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -20,10 +22,67 @@
...
+  for (; Idx < EntryNum; Idx++) {
+    if (Address <= Ranges[0].Limit) {
+      PageTable1G[Idx] = Address + (Attribute | IA32_PG_PD);
+      Address += SIZE_1GB;
+    } else {
+      break;
+    }
+  }
+
+  if ((Idx == EntryNum) && ((Address + SIZE_1GB) <= Ranges[0].Limit)) {
```
mauricema, line 81: if you check the limit is within the range at the beginning.

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -20,10 +22,67 @@
...
+
+  ASSERT (Ranges[0].Start == Ranges[0].Mapping);
+  if (((Ranges[0].Start & 0xFFF) != 0) || ((Ranges[0].Limit & 0xFFF) != 0xFFF)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  PageBits    = GetPageShiftBits (PDP_PG_LVL);
+  Alignment   = (UINTN) LShiftU64 (1, PageBits) - 1;
+  PageBase    = Ranges[0].Start & ~Alignment;
+  if (((Ranges[0].Start & ~Alignment) != PageBase) || ((Ranges[0].Limit & ~Alignment) != PageBase)) {
```
myguitar, line 62: It looks ((Ranges[0].Start & ~Alignment) != PageBase) is not required.

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -20,10 +22,67 @@
...
+
+  Address     = PageBase;
+  PdpOffset   = 1 * EFI_PAGE_SIZE;
+  PageTable1G = (UINTN *)((UINTN)PageBuffer + PdpOffset);
+  Idx = (UINT32) RShiftU64 (PageBase, PageBits);
+  EntryNum    = 512;
+
+  for (; Idx < EntryNum; Idx++) {
+    if (Address <= Ranges[0].Limit) {
+      PageTable1G[Idx] = Address + (Attribute | IA32_PG_PD);
```
myguitar, line 74: if no one uses Attibute,
```Attribute    = IA32_PG_P | IA32_PG_RW | IA32_PG_AC | IA32_PG_PD;```

PageTable1G[Idx] = Address + (Attribute
->
PageTable1G[Idx] = Address | (Attribute

File: BootloaderCommonPkg/Library/PagingLib/PagingMap.c [hunk 9]
```diff
@@ -44,60 +50,100 @@ MapMemoryRange (
...
+    Idx = (UINT32) RShiftU64 (PageBase, PageBits);
+    PageTable2M[Idx] = (UINTN)PageTable4K + Attribute;
+
+    break;
+  case SIZE_1GB:
+    // Get the start and ending index into 1G table and add 1G entries
+    Idx = (UINT32) RShiftU64 (PageBase, PageBits);
+    End = (UINT32) RShiftU64 (Ranges[0].Limit + 1, PageBits);
+    for (; Idx < End; Idx++) {
+      PageTable1G[Idx] = Address + (Attribute | IA32_PG_PD);
```
mauricema, line 139: Address should be set to start from "Ranges[0].Mapping" above line #136.

stalamudupula: What happens for a 1:1 mapping? Should the caller set Mapping = Start before passing Ranges param into this function? What about lines #108, 109?

mauricema: yes, for 1:1, the caller needs to make them identical.
#109 should only be used for 4Kb page case.

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -44,60 +50,100 @@ MapMemoryRange (
...
+      // 32 bit mode
+      PteOffset = 1 * EFI_PAGE_SIZE;
+      PdeOffset = 0;
+      EntryNum  = 1024;
     }
-    Address += SIZE_4KB;
+    PageTable4K = (UINTN *)((UINTN)PageBuffer + PteOffset);
+    PageTable2M = (UINTN *)((UINTN)PageBuffer + PdeOffset);
+    break;
+  case SIZE_1GB:
```
myguitar, line 96: Where are we using SIZE_1GB?

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -44,60 +50,100 @@ MapMemoryRange (
...
-    PteOffset = 1 * EFI_PAGE_SIZE;
-    PdeOffset = 0;
-    EntryNum  = 1024;
-  }
-
-  Alignment = (UINT32) LShiftU64 (1, PageBits) - 1;
-  PageBase  = Ranges[0].Start & ~Alignment;
-  if (((Ranges[0].Start & ~Alignment) != PageBase) || ((Ranges[0].Limit & ~Alignment) != PageBase)) {
+  // Check for Mapping PageSize alignment
+  if ( (Ranges[0].Mapping & (Ranges[0].PageSize - 1)) != 0) {
```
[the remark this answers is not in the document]
stalamudupula: I think we miss the IsLongModeEnabled() functionality if we calculate alignment based on pagesize and dont calculate PageBits as here. Isnt this more robust?

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
stalamudupula: Platform code can call this function with Range[].PageSize = SIZE_1GB, while trying to map > 4GiB PCI resources.

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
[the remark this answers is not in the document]
mauricema: we should not use IsLongModeEnabled() here since 64bit paging might be required in both 32 bit mode or 64 bit mode. It has to be determined using input parameter.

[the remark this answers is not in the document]
stalamudupula: Got it. Aiden is just asking to use the variable in the if condition. I was confusd with the Alignment  that we set again using PageBits to get the PageBase. Agree, at this point, we need to use PageSize input param. Thanks.

BootloaderCommonPkg/Library/PagingLib/PagingMap.c
myguitar: Got it. Thanks.

PR #693, opened 2020-04-27
Add MTRR print function

gdong1:
To help debug boot performance, add MTRR print function.
This function could be invoked multiple times with different
string to know where this MTRR data is printed.

### DIFF
--- a/BootloaderCommonPkg/Library/ShellLib/CmdMtrr.c
+++ b/BootloaderCommonPkg/Library/ShellLib/CmdMtrr.c
@@ -83,9 +83,21 @@ ShellCommandMtrrFunc (
   IN CHAR16 *Argv[]
   )
 {
-  UINT32  Index, BaseMsr, MaskMsr, Type;
-  BOOLEAN MtrrEnabled, FixedMtrrEnabled, Valid, Verbose;
-  UINT64  DefType, BaseVal, Base, MaskVal, Mask, Range, Limit;
+  UINT32      Index;
+  UINT32      BaseMsr;
+  UINT32      MaskMsr;
+  UINT32      Type;
+  BOOLEAN     MtrrEnabled;
+  BOOLEAN     FixedMtrrEnabled;
+  BOOLEAN     Valid;
+  UINT64      DefType;
+  UINT64      BaseVal;
+  UINT64      Base;
+  UINT64      MaskVal;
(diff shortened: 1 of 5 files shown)


myguitar:
Can you also update CmdMtrr.c?

BootloaderCommonPkg/Library/ShellLib/CmdMtrr.c
```diff
@@ -83,9 +83,21 @@ ShellCommandMtrrFunc (
...
+  UINT64      DefType;
+  UINT64      BaseVal;
+  UINT64      Base;
+  UINT64      MaskVal;
+  UINT64      Mask;
+  UINT64      Range;
+  UINT64      Limit;
+  BOOLEAN     Verbose;
 
   Verbose = FALSE;
```
myguitar, line 103: Please use PrintMtrr() here.

gdong1: the shell command should not use PrintMtrr() since it uses DEBUG to output MTRR message. It means there is output in release build for PrintMtrr(). The shell command should always works regardless debug or release build.

myguitar: The PrintMtrr and CmdMtrr is almost identical except of printing method. Can you make it a little more common? I think MtrrLib needs to have more primitive APIs.

gdong1: do you have any suggestions on how to make it more common?

myguitar: Let's make primitive APIs later when we really need external APIs.
For PrintMtrr(), I think we can add a output_function parameter in PrintMtrr() like ListDir().
if output_function is NULL, we can use DEBUG(). Otherwise, re-route to output_function.

gdong1: I don't get the exact idea on how to do this. how about check this patch firstly, then you help impl this @myguitar ?

myguitar: Sure. Let's merge this first. I will make a follow-up patch.

PR #697, opened 2020-04-29
[X64] Fix XHCI init failure

myguitar:
This issue is seen on a WHL board on X64 build when high 32-bit
BAR does not exist. In that case, MmioRead64 returns (UINT64)(-1).
To avoid this, read low and high 32-bit BAR separately.

### DIFF
--- a/BootloaderCommonPkg/Library/UsbInitLib/UsbInitLib.c
+++ b/BootloaderCommonPkg/Library/UsbInitLib/UsbInitLib.c
@@ -95,9 +95,10 @@ InitUsbDevices (
 {
   EFI_STATUS  Status;
   UINTN       PcieAddress;
-  UINTN       XhciMmioBase;
+  UINT64      XhciMmioBase;
   UINT32      Data;
   UINT8      *Class;
+  UINT8       BarType;
 
   Status = EFI_SUCCESS;
   if (mUsbInit.UsbHostHandle == NULL) {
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/UsbInitLib/UsbInitLib.c
```diff
@@ -109,7 +110,16 @@ InitUsbDevices (
...
       MmioOr8 (PcieAddress + PCI_COMMAND_OFFSET, EFI_PCI_COMMAND_MEMORY_SPACE | EFI_PCI_COMMAND_BUS_MASTER);
-      XhciMmioBase = (UINTN) MmioRead64 (PcieAddress + PCI_BASE_ADDRESSREG_OFFSET) & ~0xF;
+
+      //
+      // Read high 32-bit BAR only if 64-bit operation and 64-bit BAR memory space
+      //
+      XhciMmioBase = (UINT32)MmioRead32 (PcieAddress + PCI_BASE_ADDRESSREG_OFFSET);
+      BarType = (UINT8)XhciMmioBase & 0xF;
+      XhciMmioBase &= (UINT32)(~BarType);
+      if (IS_X64 && ((BarType & 0x04) != 0)) {
```
mauricema, line 120: Do we really need IX_X64 check here ?
in 32 bit mode I think it is also valid to read the full 64bit bar.  The higher 32 bit should be always 0 though.

myguitar: This was to avoid unnecessary high 32-bit BAR reading in IA32. Let me declare XhciMmioBase as UINT64 instead of UINTN and remove IS_X64.

myguitar:
Updated according to review comment.

BootloaderCommonPkg/Library/UsbInitLib/UsbInitLib.c
```diff
@@ -109,9 +109,12 @@ InitUsbDevices (
         (Class[2] == PCI_CLASS_SERIAL))) {
       // Enable XHCI controller
       MmioOr8 (PcieAddress + PCI_COMMAND_OFFSET, EFI_PCI_COMMAND_MEMORY_SPACE | EFI_PCI_COMMAND_BUS_MASTER);
-      XhciMmioBase = (UINTN) MmioRead64 (PcieAddress + PCI_BASE_ADDRESSREG_OFFSET) & ~0xF;
 
-      Status = UsbInitCtrl (XhciMmioBase, &mUsbInit.UsbHostHandle);
+      // Read low and high BAR separately
+      XhciMmioBase  = (UINT32)MmioRead32 (PcieAddress + PCI_BASE_ADDRESSREG_OFFSET) & (UINT32)(~0xF);
+      XhciMmioBase |= LShiftU64 ((UINT64)MmioRead32 (PcieAddress + PCI_BASE_ADDRESSREG_OFFSET + 0x4), 32);
```
mauricema, line 115: I think your code change is still valid, just need to remove IS_X64 from the condition. 
We still need to check the type type before reading the higher 32 bit.  Only 64 bit bar needs to read higher 32 bit.

myguitar: Added 64-bit BAR type check.

PR #699, opened 2020-04-30
Enhance debug log buffer as ring buffer

myguitar:
This will allow debug log buffer to record logs in ring buffer
if the buffer is full.

### DIFF
--- a/BootloaderCommonPkg/Include/Guid/LoaderPlatformDataGuid.h
+++ b/BootloaderCommonPkg/Include/Guid/LoaderPlatformDataGuid.h
@@ -14,26 +14,14 @@
 ///
 extern EFI_GUID gLoaderPlatformDataGuid;
 
-#define  DEBUG_LOG_BUFFER_SIGNATURE         SIGNATURE_32 ('D', 'L', 'O', 'G')
-
 #define  DEBUG_OUTPUT_DEVICE_LOG_BUFFER     BIT0
 #define  DEBUG_OUTPUT_DEVICE_SERIAL_PORT    BIT1
 #define  DEBUG_OUTPUT_DEVICE_CONSOLE        BIT2
 
 typedef struct {
-  UINT32  Signature;
-  UINT8   HeaderLength;
-  UINT8   Attribute;
-  UINT8   Reserved[2];
-  UINT32  UsedLength;
-  UINT32  TotalLength;
-  UINT8   Buffer[0];
-} DEBUG_LOG_BUFFER_HEADER;
-
-typedef struct {
   UINT8                     Revision;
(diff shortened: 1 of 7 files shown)


mauricema:
do you know why the original code will cause memory corruption ?

myguitar:
> do you know why the original code will cause memory corruption ?

In DebugLogBufferWrite(), UsedLength keeps increasing regardless of maximum buffer size. Therefore, it could corrupt memory if the length of total logs exceeds max buffer size, most possibly in Debug build.

```
if (NumberOfBytes > 0) {
  CopyMem ((UINT8 *)LogBufHdr + LogBufHdr->UsedLength, Buffer, NumberOfBytes);
  LogBufHdr->UsedLength += (UINT32)NumberOfBytes;
}
```

mauricema:
The old code did have a check on buffer overflow as below,  I am wondering why it does not work as expected:
 if (LogBufHdr->UsedLength + NumberOfBytes > LogBufHdr->TotalLength) {
    NumberOfBytes = LogBufHdr->TotalLength - LogBufHdr->UsedLength;
 }
 if (NumberOfBytes > 0) {
    CopyMem ((UINT8 *)LogBufHdr + LogBufHdr->UsedLength, Buffer, NumberOfBytes);
    LogBufHdr->UsedLength += (UINT32)NumberOfBytes;
 }

myguitar:
> The old code did have a check on buffer overflow as below, I am wondering why it does not work as expected:
> if (LogBufHdr->UsedLength + NumberOfBytes > LogBufHdr->TotalLength) {
> NumberOfBytes = LogBufHdr->TotalLength - LogBufHdr->UsedLength;
> }
> if (NumberOfBytes > 0) {
> CopyMem ((UINT8 *)LogBufHdr + LogBufHdr->UsedLength, Buffer, NumberOfBytes);
> LogBufHdr->UsedLength += (UINT32)NumberOfBytes;
> }

Oh you are right. If buffer index reaches at the maximum length, it does not record logs any more. My bad.
Let me replicate the memory corruption issue again. It looks MP synchronization issue due to global log buffer access when debug messages are printed inside ApFunc. Let me double check.

mauricema:
I see.  If it is called in MP, it might have issue because we don't have a lock for this at this point.  
Anyway, for your change, I think it is good to change the buffer to circular ring buffer.  We need to address the MP print issue separately.

myguitar:
> I see. If it is called in MP, it might have issue because we don't have a lock for this at this point.
> Anyway, for your change, I think it is good to change the buffer to circular ring buffer. We need to address the MP print issue separately.

Sure, let me take a look MP print issue separately.

PR #702, opened 2020-05-01
TpmLib: Add appropriate safety comparison checks

CunningLearner:
*Safety comparison checks in accordance with TPM2.0 Specification data structures
*Fix comment typos

### DIFF
--- a/BootloaderCommonPkg/Library/TpmLib/Tpm2Capability.c
+++ b/BootloaderCommonPkg/Library/TpmLib/Tpm2Capability.c
@@ -44,7 +44,7 @@ typedef struct {
   NOTE:
   To simplify this function, leave returned CapabilityData for caller to unpack since there are
   many capability categories and only few categories will be used in firmware. It means the caller
-  need swap the byte order for the feilds in CapabilityData.
+  need swap the byte order for the fields in CapabilityData.
 
   @param[in]  Capability         Group selection; determines the format of the response.
   @param[in]  Property           Further definition of information.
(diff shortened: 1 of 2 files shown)


CunningLearner:
> do you really need DEBUG_ERROR for every error ? Consider limiting it because it has size impact on final image.

There was only one, still it results in error.
Well, that's too okay, as I believe EFI_D_ERROR is nowhere exercised in our source code - limiting completely!

mauricema:
> > do you really need DEBUG_ERROR for every error ? Consider limiting it because it has size impact on final image.
> 
> There was only one, still it results in error.
> Well, that's too okay, as I believe EFI_D_ERROR is nowhere exercised in our source code - limiting completely!

The newly added code added around 5 "DEBUG ((DEBUG_ERROR, ...)" statements.  In most of the cases, I think error status code should be good enough.  We don't have to print every detailed error string.  SBL is very sensitive to the final code size.

CunningLearner:
> > > do you really need DEBUG_ERROR for every error ? Consider limiting it because it has size impact on final image.
> > 
> > 
> > There was only one, still it results in error.
> > Well, that's too okay, as I believe EFI_D_ERROR is nowhere exercised in our source code - limiting completely!
> 
> The newly added code added around 5 "DEBUG ((DEBUG_ERROR, ...)" statements. In most of the cases, I think error status code should be good enough. We don't have to print every detailed error string. SBL is very sensitive to the final code size.

Yes sure, we can remove every debug statements (all 5). For my knowledge, Is there any verbose mechanism available which can help in such scenarios facilitating both needs?

mauricema:
> > > > do you really need DEBUG_ERROR for every error ? Consider limiting it because it has size impact on final image.
> > > 
> > > 
> > > There was only one, still it results in error.
> > > Well, that's too okay, as I believe EFI_D_ERROR is nowhere exercised in our source code - limiting completely!
> > 
> > 
> > The newly added code added around 5 "DEBUG ((DEBUG_ERROR, ...)" statements. In most of the cases, I think error status code should be good enough. We don't have to print every detailed error string. SBL is very sensitive to the final code size.
> 
> Yes sure, we can remove every debug statements (all 5). Consider it done!

BTW,  if you like,  you can also use DEBUG_VERBOSE level. This level of debug will not be built into the final image by default.  But still can be enabled for debug purpose.

PR #707, opened 2020-05-11
Fix Klockwork issue flagged in BootloaderCommonPkg

vegnish:
Fix for: Klockwork flags multiple possible null pointer variable being dereferenced.

Signed-off-by: Vegnish Rao <vegnish.rao.paramesura.rao@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
+++ b/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
@@ -122,6 +122,9 @@ MapMemoryRange (
 
     Address = PageBase;
     // Create 4KB pages and remap the CAR region into a different memory location
+    if (PageTable4K == NULL) {
+      return EFI_NOT_FOUND;
+    }
     for (Idx = 0; Idx < EntryNum; Idx++) {
       if ((Address >= Ranges[0].Start) && (Address <= Ranges[0].Limit)) {
         PageTable4K[Idx]   = Ranges[0].Mapping + (Attribute | IA32_PG_PD);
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -122,6 +122,9 @@ MapMemoryRange (
 
     Address = PageBase;
     // Create 4KB pages and remap the CAR region into a different memory location
+    if (PageTable4K == NULL) {
+      return EFI_NOT_FOUND;
+    }
```
gdong1, line 127: The issues was caused by 2 “switch (Ranges[0].PageSize) {}” in same function.
Instead of add NULL check, better to merge them into one switch?

stalamudupula:
I'd prefer it to stay this way. First switch case does some init , and the second one actually creates page table entries. If we club them together, we might have to do some duplicate calculations.

gdong1:
> I'd prefer it to stay this way. First switch case does some init , and the second one actually creates page table entries. If we club them together, we might have to do some duplicate calculations.

I saw it only saves 2 lines of code (line 113 and 114) to have 2 switches.  is there any other benefit?

stalamudupula:
Yes, those are the 2 lines which I was talking about. No other benefit. But , I thought no it is more readable this way. I am OK if we want to have a single switch case too.

gdong1:
> Yes, those are the 2 lines which I was talking about. No other benefit. But , I thought no it is more readable this way. I am OK if we want to have a single switch case too.

We could save ~20 lines of code by merging them.  And it is more readable for me by putting code together for same purpose :)

vegnish:
@stalamudupula Could you help take a look and see if you'd like to make the changes suggested by Guo?
Otherwise, we could proceed to merge it.

stalamudupula:
Hi @vegnish , let me create a new PR for taking care of this kw issue. Please go ahead and close this PR. Thanks,

vegnish:
Closing this PR.
Issue addressed in https://github.com/slimbootloader/slimbootloader/pull/708

PR #708, opened 2020-05-11
Fix Klockwork issue flagged in PagingMap lib

stalamudupula:
Klocwork reports a potential dereferencing of a NULL
pointer. This patch addresses the issue.

### DIFF
--- a/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
+++ b/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
@@ -66,10 +66,6 @@ MapMemoryRange (
   UINTN          Alignment;
   UINT32         PageBits;
 
-  PageTable1G = 0;
-  PageTable2M = 0;
-  PageTable4K = 0;
-
   Attribute = IA32_PG_P | IA32_PG_RW | IA32_PG_AC;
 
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/PagingLib/PagingMap.c
```diff
@@ -98,23 +98,10 @@ MapMemoryRange (
...
-  case SIZE_1GB:
-    ASSERT (Ranges[0].Limit < SIZE_512GB);
-    PageBits    = GetPageShiftBits (PDP_PG_LVL);
-    PdpOffset   = 1 * EFI_PAGE_SIZE;
-    EntryNum    = 512;
-    PageTable1G = (UINTN *)((UINTN)PageBuffer + PdpOffset);
-    break;
-  default:
-    return EFI_INVALID_PARAMETER;
-  }
```
gdong1, line 111: The changes looks good to me.
Had better remove the lines 69 ~ 71: They should be initialized to NULL instead 0,  and now it looks they are not required.

stalamudupula: Yes, agree. let me make that change.

PR #710, opened 2020-05-12
Config Data name for SPI boot is Updated.

naveenm18:
Boot Options updated for selecting boot device as SPI.
Use the “Memory” setting in the ConfigEditor to select
SPI as boot device.

Signed-off-by: Mutha <naga.naveen.mutha@intel.com>


Platform/CommonBoardPkg/CfgData/Template_BootOption.dsc
```diff
@@ -34,7 +34,7 @@
     gCfgData.Reserved_$(1)                  |      * | 0x01 | 0x5A
     # !BSF NAME:{Boot Device type}
     # !BSF TYPE:{Combo}
-    # !BSF OPTION:{0:SATA, 1:SD, 2:EMMC, 3:UFS, 4:SPI, 5:USB, 6:NVME, 7:MAX}
+    # !BSF OPTION:{0:SATA, 1:SD, 2:EMMC, 3:UFS, 4:SPI_PDR, 5:USB, 6:NVME, 7:MEMORY}
```
mauricema, line 37: Let us keep it as "SPI". It does not have to be PDR region.

PR #720, opened 2020-05-19
Allow using bootloader stack to call FspMemoryInit

mauricema:
FSP 2.1 introduced new requirement to use bootloader stack for FSP-M. It
will cause issue for SBL since SBL only uses a small stack in Stage1. To
address this isue, a new PCD PcdFSPMStackTop is added to control the
stack settings for FSP-M.
  - If it is 0, it will not switch stack before calling FspMemoryInit API.
  - If it is 0xffffffff, it will switch to the new default FSP before
    calling FspMemoryInit API.
  - For other values, it will switch to the new stack at specified value
    before calling FspMemoryInit API.
This PCD will be set automatically by FSP_M_STACK_TOP variable in
BoardConfig.py file.

This code has been tested on UP Extreme board with latest FSP version.

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dec
+++ b/BootloaderCorePkg/BootloaderCorePkg.dec
@@ -144,6 +144,9 @@
   # If static patching is not required, zero default value can be used with no issue.
   #
   gPlatformModuleTokenSpaceGuid.PcdFSPSBase               | 0xFF000000 | UINT32 | 0x20000100
+  # Switch FSP-M stack top to new value.
+  # 0: do not switch,  0xffffffff: switch to default,  others: switch to specific value
+  gPlatformModuleTokenSpaceGuid.PcdFSPMStackTop           | 0x00000000 | UINT32 | 0x20000101
   gPlatformModuleTokenSpaceGuid.PcdAcpiTablesMaxEntry     |         32 | UINT32 | 0x2000000C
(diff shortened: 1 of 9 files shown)


BootloaderCorePkg/BootloaderCorePkg.dec
```diff
@@ -144,6 +144,9 @@
   # If static patching is not required, zero default value can be used with no issue.
   #
   gPlatformModuleTokenSpaceGuid.PcdFSPSBase               | 0xFF000000 | UINT32 | 0x20000100
+  # Switch FSP-M stack top to new value.
+  # 0: do not switch,  0xffffffff: wwitch to default,  others: switch to specific value
```
rprangar, line 148: Typo. wwitch->Switch.

mauricema: will fix it.

BuildLoader.py
```diff
@@ -219,6 +219,7 @@ def __init__(self, *args, **kwargs):
         self.BPM_SIZE              = 0x1000 # valid only if ACM_SIZE > 0
         self.CFG_DATABASE_SIZE     = 0
 
+        self.FSP_M_STACK_TOP       = 0
         self.STAGE1A_XIP           = 1
```
gdong1, line 223: Had better update core DSC file to override PcdFSPMStackTop by this value.

mauricema: Good catch.  Added it in the new commit.

PR #722, opened 2020-05-21
Add OEM container verification support

mauricema:
Current SBL supports container header verification. If the container
signature is BOOT, it will use HASH_USAGE_PUBKEY_OS. Otherwise, it
will use HASH_USAGE_PUBKEY_CONTAINER_DEF. This patch added OEM signed
container support. If a container signature between OEM0 to OEM7 is
found, it will be verified use HASH_USAGE_PUBKEY_OEM(x) where x is 0
to 7. To add an OEM public key hash, it can be done by updating
pub_key_list in GetKeyHashList() in file BoardConfig.py.


BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -333,11 +333,20 @@ GetContainerKeyUsageBySig (
...
 {
+  UINT8  Idx;
+
   if (ContainerSig == CONTAINER_BOOT_SIGNATURE) {
     return HASH_USAGE_PUBKEY_OS;
-  } else {
-    return HASH_USAGE_PUBKEY_CONTAINER_DEF;
   }
+
+  if ((ContainerSig & 0x00FFFFFF) == CONTAINER_OEM_BASE_SIGNATURE) {
```
lsubash, line 342: Does this condition works when container uses signature as 
"SIGNATURE_32 ('O', 'E', 'M',   **1)"**?

mauricema: Container signature is expected to be 'OEM0' to 'OEM7'.  We will match 'OEM' part first, then decode the 0-7 number.   SIGNATURE_32 ('O', 'E', 'M', 1) will not be considered as a valid OEM signture.  it needs to be SIGNATURE_32 ('O', 'E', 'M', '1').

lsubash: Thanks. I overlooked 0x**00**FFFFFF.

lsubash:
Looks good,

PR #724, opened 2020-05-21
Validate board config hash types in pre-build

lsubash:
SIGN_HASH_TYPE and IPP_HASH_LIB_SUPPORTED_MASK are derived from
_SIGN_HASH. AT times only _SIGN_HASH is configured in
BoardConfig.py which causes in incorrect hash set to
respective PCDs.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -1029,6 +1029,14 @@ class Build(object):
         if not os.path.exists(self._fv_dir):
             os.makedirs(self._fv_dir)
 
+        # Validate HASH_TYPE_VALUE config
+        if (HASH_TYPE_VALUE[self._board._SIGN_HASH] != self._board.SIGN_HASH_TYPE):
+            raise Exception  ('SIGN_HASH_TYPE is not set correctly!!')
+
+        # Validate IPP_HASH_LIB_SUPPORTED_MASK config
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -1029,6 +1029,14 @@ def pre_build(self):
         if not os.path.exists(self._fv_dir):
             os.makedirs(self._fv_dir)
 
+        # Validate HASH_TYPE_VALUE config
+        if (HASH_TYPE_VALUE[self._board._SIGN_HASH] != self._board.SIGN_HASH_TYPE):
```
mauricema, line 1033: which one do we expect user to change  _SIGN_HASH or  SIGN_HASH_TYPE ?

lsubash: Ideally user is expected to configure all three,

- _SIGN_HASH (string),  SIGN_HASH_TYPE (value to set PCD)

- IPP_HASH_LIB_SUPPORTED_MASK  based on their needs.

For ease, we have derived SIGN_HASH_TYPE and IPP_HASH_LIB_SUPPORTED_MASK from _SIGN_HASH.  This happens fine if default buildloader configs are used.
Incase if user configures only _SIGN_HASH in boardconfig.py, it would cause a mismatch.

PR #727, opened 2020-05-28
Segregate sign utilities to an separate script

lsubash:
SignUtility.py – It contains common functionality
    for signing and extraction Of public key info.
 	It adds the necessary structures for signed data.
SingleSign.py – It contains core functionality related
    to openssl for sign operations. This script will
	be enhanced for accessing key store.


lsubash:
@sagraw2 for review

BootloaderCorePkg/Tools/SignUtility.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
```
mauricema, line 1: why do we need SignUtiltiy.py and SingleSign.py ?  Can we use just one file to handle that ?  
I think the higher level functions in SignUtiltiy.py can be kept inside CommonUtility.

BootloaderCorePkg/Tools/SingleSign.py
```diff
@@ -0,0 +1,141 @@
...
+    # Temporary files to store hash generated
+    hash_file_tmp = out_file+'.hash.tmp'
+    hash_file     = out_file+'.hash'
+
+    # Generate hash using openssl dgst in hex format
+    cmdargs = [get_openssl_path(), 'dgst', '-'+'%s' % _hash_type_string[hash_type], '-out', '%s' % hash_file_tmp, '%s' % in_file]
+    run_process (cmdargs)
+
+    #sample hash data generated by dgst command
+    #SHA256()= 7658f28ae53f61b6dbaeac5ea13d0a169a3f325773bfa5c3cdfc2d0cad4a74d3
```
mauricema, line 56: pls remove the comment.

BootloaderCorePkg/Tools/SignUtility.py
lsubash: Functionality in SingleSign.py import some functionality from CommonUtility.
Keeping higher level functions in SignUtiltiy.py to CommonUtility (Which exists now) creates cyclic dependency between modules. Thought of keeping as a separate file.

BootloaderCorePkg/Tools/SingleSign.py
lsubash: will do

BootloaderCorePkg/Tools/SingleSign.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python
+## @ SingleSing.py
```
mauricema, line 2: SingleSign

BootloaderCorePkg/Tools/SignUtility.py
mauricema: OK

BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -23,6 +23,8 @@
 import ntpath
 from   CommonUtility import *
 from   IfwiUtility   import FLASH_MAP, FLASH_MAP_DESC, FIT_ENTRY, UCODE_HEADER
+from   SignUtility import *
```
sagraw2, line 26: If a customer needs to use two separate utilities for production and non-production, he/she will have to modify/update all the tools.
Can SignUtility.py be included only from CommonUtility.py ?
And 'CommonUtility.py' should be the only interface for 'SignUtility' with SBL code.

lsubash:
Update PR - Removed dependencies for SingleSign.py with other scripts. get_openssl_path() and run_process() got duplicated.
Removed SignUitility.py and common sign utilities are moved back to CommonUtility.py.

BootloaderCorePkg/Tools/SingleSign.py
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+
+    # Temporary files to store hash generated
+    hash_file_tmp = out_file+'.hash.tmp'
+    hash_file     = out_file+'.hash'
+
+    # Generate hash using openssl dgst in hex format
+    cmdargs = [get_openssl_path(), 'dgst', '-'+'%s' % _hash_type_string[hash_type], '-out', '%s' % hash_file_tmp, '%s' % in_file]
+    run_process (cmdargs)
```
sagraw2, line 99: Does run_process throw user friendly error string if any of below conditions happens?
'priv_key' file is not present
'openssl.exe' is not present

lsubash: run_process() for openssl.exe not present - it would throw an exception. We would need to add additional checks for better handling.

For Priv_key not present - it would capture error.
FileNotFoundError: [Errno 2] No such file or directory: './BootloaderCorePkg/Tools/Keys/TestSigningPrivateKey1.pem'

BootloaderCorePkg/Tools/SingleSign.py
sagraw2: is the above comment removed ?

lsubash: Fixed it.

PR #735, opened 2020-06-02
Remove TestSigningPrivateKey as default in GenContainer

lsubash:
TestSigningPrivateKey is defaulted for container
creation in non-layout form and key dir is used.
In CfgDataStitch also TestSigningPrivateKey is
defaulted when key dir is specified.

### DIFF
--- a/BootloaderCorePkg/Tools/CfgDataStitch.py
+++ b/BootloaderCorePkg/Tools/CfgDataStitch.py
@@ -36,7 +36,7 @@ def check_file_exist (chk_files):
     return ''
 
 
-def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_dir, script_dir, tool_dir, platform_id):
+def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_file, script_dir, tool_dir, platform_id):
     if len (dlt_files) == 0:
         raise Exception("Please run the generated CfgDataStitch.py script instead of the original one in source tree !")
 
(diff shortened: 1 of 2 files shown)


lsubash:
@sagraw2 for review

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -797,7 +797,7 @@ def main():
     group.add_argument('-cl', dest='comp_list',nargs='+', help='List of each component files, following XXXX:FileName format')
     cmd_display.add_argument('-t', dest='img_type',  type=str, default='CLASSIC', help='Container Image Type : [NORMAL, CLASSIC, MULTIBOOT]')
     cmd_display.add_argument('-o', dest='out_path',  type=str, default='.', help='Container output directory/file')
-    cmd_display.add_argument('-k', dest='key_path',  type=str, default='', help='Input key directory/file')
+    cmd_display.add_argument('-k', dest='key_path',  type=str, default='', help='Input key directory/file. Use directoy for Container layout as intput file ')
```
sagraw2, line 800: If there is no layout file, then you expect a 'key_file_name' to be provided here. Isn't it?
In that case, this help text is mis-leading.

Also, too many syntax errors.

"Input key directory/file. Use directoy for Container layout as intput file "

BootloaderCorePkg/Tools/CfgDataStitch.py
```diff
@@ -69,16 +69,8 @@ def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_dir, script_dir, tool_
...
-            key_dir = os.path.realpath(os.path.join(fv_dir, '../../../../BootloaderCorePkg/Tools/Keys'))
-        else:
-            key_dir = get_script_dir()
-
-    if os.path.isdir(key_dir):
-        key_file = os.path.join(key_dir, 'TestSigningPrivateKey.pem')
-    else:
-        key_file = key_dir
+    if key_file == '':
+        raise Exception("Key file is not specified!!")
```
sagraw2, line 73: Can't we use CFGDATA_KEY by default here....if one is not provided ?

BootloaderCorePkg/Tools/GenContainer.py
lsubash: Updated decsriptions.

BootloaderCorePkg/Tools/CfgDataStitch.py
lsubash: We are planning to provide KEY_ID or complete key path for standalone scripts.
I think we could default CFGDATA_KEY at later stage.

PR #736, opened 2020-06-02
Add key ids to single sign script

lsubash:
This patch adds support to key ids in single sign script.
Following pre-requistes to enable usage of key ids,
- Generate required RSA keys as per GenerateKeys.py
- SLIMBOOT_KEY_DIR env variable set to key folder
- Set private key paths to respective ids in
  buildloader.py and boardconfigs files
- Update key hash store generation to use respective key ids

Enabling keyids in slimboot would be done subsequent patches.

### DIFF
--- a/BootloaderCorePkg/Tools/CommonUtility.py
+++ b/BootloaderCorePkg/Tools/CommonUtility.py
@@ -211,6 +211,7 @@ def run_process (arg_list, print_cmd = False, capture_out = False):
 
     return output
 
+
 def rsa_sign_file (priv_key, pub_key, hash_type, sign_scheme, in_file, out_file, inc_dat = False, inc_key = False):
 
     bins = bytearray()
@@ -236,8 +237,13 @@ def rsa_sign_file (priv_key, pub_key, hash_type, sign_scheme, in_file, out_file,
 
 def get_key_type (in_key):
 
-    # Check for public key in binary format.
-    key = bytearray(get_file_data(in_key))
+    # Check in_key is file or key Id
+    if not os.path.exists(in_key):
+        key = bytearray(gen_pub_key (in_key))
+    else:
+        # Check for public key in binary format.
+        key = bytearray(get_file_data(in_key))
+
     pub_key_hdr = PUB_KEY_HDR.from_buffer(key)
     if pub_key_hdr.Identifier != b'PUBK':
         pub_key = gen_pub_key (in_key)
--- a/BootloaderCorePkg/Tools/SingleSign.py
+++ b/BootloaderCorePkg/Tools/SingleSign.py
@@ -19,6 +19,42 @@ import struct
 import hashlib
 import string
 
+# KEY_SIZE_TYPE defines the key sizes to be used for signing
+# KEY_SIZE_TYPE = RSA2048 uses RSA 2K size keys
+# KEY_SIZE_TYPE = RSA3072 uses RSA 3K size keys
+KEY_SIZE_TYPE = 'RSA2048'
+
+SIGNING_KEY = {
+    # Key Id                    | Key File Name start |
+    # ===========================================================
+    # MASTER_KEY_ID is used for signing Slimboot Key Hash Manifest (KEYH Component)
+    "MASTER_KEY_ID"          :    "MasterTestKey_Priv",
+
+    # CFGDATA_KEY_ID is used for signing external Config data blob)
+    "CFGDATA_KEY_ID"         :    "ConfigTestKey_Priv",
+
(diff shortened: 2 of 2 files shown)


lsubash:
Adding @sagraw2 for review

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -211,8 +211,24 @@ def run_process (arg_list, print_cmd = False, capture_out = False):
...
+    # Check for KEY_ID at end of string.
+    if (key_name[-6:] == "KEY_ID"):
+        return key_name
+    else:
+        return None
+
+
 def rsa_sign_file (priv_key, pub_key, hash_type, sign_scheme, in_file, out_file, inc_dat = False, inc_key = False):
 
+    # Check priv_key is path to private key
```
sagraw2, line 225: We probably do not need this logic here. 
It's better to encapsulate all key logic in single_sign script itself.

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -236,8 +252,17 @@ def rsa_sign_file (priv_key, pub_key, hash_type, sign_scheme, in_file, out_file,
 
 def get_key_type (in_key):
 
-    # Check for public key in binary format.
-    key = bytearray(get_file_data(in_key))
+    # Check priv_key is path to private key
+    # Extract key_id if present
+    if not os.path.exists(in_key):
```
sagraw2, line 257: Similarly, we probably do not need this here.

BootloaderCorePkg/Tools/SingleSign.py
```diff
@@ -89,6 +118,14 @@ def single_sign_file (priv_key, hash_type, sign_scheme, in_file, out_file):
         "RSA_PSS"      : 'pss',
     }
 
+    # Check for Key Id or key path
+    if not os.path.exists(priv_key):
```
sagraw2, line 122: I think the logic should be something like this:

if (os.path.exists(priv_key)):
  use_it

else {
 //Get SLIMBOOT_KEY_DIR path.
//if (priv_key is simple file name) search for file 'priv_key' in SLIMBOOT_KEY_DIR
//if (priv_key is KEY_ID) search for corresponding keyfile  in SLIMBOOT_KEY_DIR
}

BootloaderCorePkg/Tools/CommonUtility.py
lsubash: In container cases specially with layout, key can be passed in either ways,
- $Path\CONTAINER_KEY_ID
- $Path\ContainerTestKey_Priv_RSA2048.pem

In case of $Path\CONTAINER_KEY_ID - we are extracting keyid and pass to single sign script.

I would move this to single sign script to have all handling at one place.

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -211,8 +211,24 @@ def run_process (arg_list, print_cmd = False, capture_out = False):
 
     return output
 
+def get_key_id (priv_key):
+    key_name = os.path.basename(priv_key)
+    # Check for KEY_ID at end of string.
+    if (key_name[-6:] == "KEY_ID"):
```
mauricema, line 217: use  key_name.endswith ('KEY_ID')  instead.

BootloaderCorePkg/Tools/SingleSign.py
```diff
@@ -19,6 +19,35 @@
...
+    "FIRMWAREUPDATE_KEY_ID"  :    "FirmwareUpdateTestKey_Priv",
+
+    # CONTAINER_KEY_ID is used for signing container header with mono signature
+    "CONTAINER_KEY_ID"       :    "ContainerTestKey_Priv",
+
+    # CONTAINER_COMP1_KEY_ID is used for signing container components
+    "CONTAINER_COMP_KEY_ID" :    "ContainerCompTestKey_Priv",
+
+    # OS1_PUBLIC_KEY_ID, OS2_PUBLIC_KEY_ID is used for referencing Boot OS public keys
+    "OS1_PUBLIC_KEY_ID"      :    "OS1_TesKey_Pub",
```
mauricema, line 46: OS1_TesKey_Pub should be OS1_TestKey_Pub, same for OS2_TesKey_Pub

lsubash:
Update PR,
- As per comments provided by Sachin & Maurice
- Encapsulated  key logic in single_sign script.
  priv can be - Key id, complete key path, just key file in pem format.
- Improved error handling

PR #737, opened 2020-06-03
WIP: TEST Travis signing

lsubash:


Platform/QemuBoardPkg/BoardConfig.py
```diff
@@ -197,12 +206,12 @@ def GetContainerList (self):
...
           # ============================================================================================================================================
-          ('IPFW',      'SIIPFW.bin',    '',           'RSA2048_PSS_SHA2_256',   'TestSigningPrivateKey.pem',    0,             0     ),   # Container Header
+          ('IPFW',      'SIIPFW.bin',    '',           'RSA2048_PSS_SHA2_256',   'CONTAINER_KEY_ID',    0,             0     ),   # Container Header
           ('TST1',      '',              'Dummy',      '',                   '',                                 0,             0x2000),   # Component 1
           ('TST2',      '',              'Lz4',        '',                   '',                                 0,             0x3000),   # Component 2
-          ('TST3',      '',              'Lz4',        'RSA2048_PSS_SHA2_256',   'TestSigningPrivateKey.pem',    0,             0x3000),   # Component 3
+          ('TST3',      '',              'Lz4',        'RSA2048_PSS_SHA2_256',   'CONTAINER_KEY_ID',    0,             0x3000),   # Component 3
```
mauricema, line 212: This should be CONTAINER_COMP_KEY_ID ?

lsubash: This patch is not complete. I am trying to validate travis build changes.
Yes it should be CONTAINER_COMP_KEY_ID.

.travis.yml
```diff
@@ -67,7 +67,7 @@ script:
   - |
     if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       if [[ "$BUILD_TARGET" == "qemu" && "$BUILD_ARCH" == "" ]]; then
-        docker run --rm -v ${PWD}:/tmp/sbl -w /tmp/sbl --network=host sbl bash -c "python3 ./BuildLoader.py build ${BUILD_REL} ${BUILD_TARGET} && python3 ./Platform/QemuBoardPkg/Script/qemu_test.py";
+        docker run --rm -v ${PWD}:/tmp/sbl -w /tmp/sbl --network=host sbl bash -c "python3 BootloaderCorePkg/Tools/GenerateKeys.py -k ./Sbl_Key_Dir && python3 ./BuildLoader.py build ${BUILD_REL} ${BUILD_TARGET} && python3 ./Platform/QemuBoardPkg/Script/qemu_test.py";
```
mauricema, line 70: Can we avoid adding '-k' for the root level build script ?
We can do the following in BuildLoader.py
Set an env variable SBL_KEY_DIR to default '../SblKeyDir' if this does not exist.  If exists, then use the value instead.  Then everything will use this path for key dir.

PR #745, opened 2020-06-11
Add PciGetMaxBusNumber for PCI Bridge Subordinate Bus

myguitar:
At PciScanBus, a PCI bridge sets PCI Bridge Subordinate Bus to 0xFF
temporary to go thru any PPB. But, a platform has some reserved buses
(ex. 0xFB-0xFF) on PCI hierarchy, and writing 0xFF regardless of
reserved bus ranges causes system hang.

Therefore, PciGetMaxBusNumber will be used for PCI Bridge Subordinate
Bus and it gets the number of buses from PCI Enum Policy to skip the
reserved buses.

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
@@ -81,6 +81,11 @@ struct _PCI_BAR_RESOURCE {
   PCI_BAR                                  *PciBar;
 };
 
+typedef struct {
+  UINT8                                     BusBase;
+  UINT8                                     BusLimit;
+} PCI_BUS_NUM_RANGE;
+
 typedef struct _PCI_IO_DEVICE              PCI_IO_DEVICE;
 
 struct _PCI_IO_DEVICE {
@@ -100,6 +105,11 @@ struct _PCI_IO_DEVICE {
   // BAR for this PCI Device
   //
   PCI_BAR                                   PciBar[PCI_MAX_BAR];
+
+  //
+  // Bus number ranges for a PCI Root Bridge device
+  //
+  PCI_BUS_NUM_RANGE                         BusNumberRanges;
 
   //
   // ARI (Alternative Routing ID)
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
```diff
@@ -101,6 +108,11 @@ struct _PCI_IO_DEVICE {
   //
   PCI_BAR                                   PciBar[PCI_MAX_BAR];
 
+  //
+  // Bus number ranges for a PCI Root Bridge device
+  //
+  PCI_BUS_NUM_RANGE                         *BusNumberRanges;
```
mauricema, line 114: Is this an array ?   if it is not,  is it better to use PCI_BUS_NUM_RANGE instead of PCI_BUS_NUM_RANGE*  ?   It will avoid dynamic allocation.

BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
```diff
@@ -81,6 +81,13 @@ struct _PCI_BAR_RESOURCE {
   PCI_BAR                                  *PciBar;
 };
 
+typedef struct _PCI_BUS_NUM_RANGE          PCI_BUS_NUM_RANGE;
+
+struct _PCI_BUS_NUM_RANGE {
+  UINT8                                     BusBase;
+  UINT8                                     BusLimit;
+};
```
gdong1, line 89: It looks it is not necessary to define _PCI_BUS_NUM_RANGE, it would be simple as below:
typedef struct {
  UINT8                                     BusBase;
  UINT8                                     BusLimit;
} PCI_BUS_NUM_RANGE;

BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
myguitar: Addressed.

BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
myguitar: Addressed.

PR #747, opened 2020-06-12
Fix invalid PPB check in PciGetMaxBusNumber()

myguitar:
A PPB PCI_IO_DEVICE instance has BIT31 in its Address field to identify
the device as PPB type. But, the bit is set after scanning the PPB.
This skips PPB type check in PciGetMaxBusNumber() and let a caller
guarantee PPB type check instead of adding a field in PCI_IO_DEVICE
for PPB device.


mauricema:
looks good to me.

PR #749, opened 2020-06-15
OS loader Optimization for Raw Partition Boot.

naveenm18:
Removed Checking for MBR/GPT partitions.
In FindPartitions() identifying memory device
using MediaGetInterfaceType().

Signed-off-by: Mutha <naga.naveen.mutha@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
+++ b/BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
@@ -522,6 +522,7 @@ FindPartitions (
   LOGICAL_BLOCK_DEVICE        *BlockDev;
   EFI_PARTITION_TABLE_HEADER  *Gpt;
   PART_BLOCK_DEVICE           *PartBlockDev;
+  OS_BOOT_MEDIUM_TYPE          CurrentMediaType;
 
   if (PartHandle == NULL) {
     return EFI_INVALID_PARAMETER;
@@ -531,6 +532,7 @@ FindPartitions (
   if (EFI_ERROR (Status)) {
     return Status;
   }
+
 
   PartBlockDev = (PART_BLOCK_DEVICE *) AllocateZeroPool (sizeof (PART_BLOCK_DEVICE));
   if (PartBlockDev == NULL) {
@@ -542,7 +544,9 @@ FindPartitions (
 
   CopyMem (&PartBlockDev->BlockInfo, &DevBlockInfo, sizeof (DEVICE_BLOCK_INFO));
 
-  if (DevBlockInfo.BlockSize == 1) {
+  CurrentMediaType = MediaGetInterfaceType();
+
+  if (CurrentMediaType == OsBootDeviceMemory || CurrentMediaType == OsBootDeviceSpi) {
     // It is memory mapped SPI block device.
     Status = FindSpiPartitions (PartBlockDev);
     if (!EFI_ERROR (Status)) {


BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -542,7 +540,7 @@ FindPartitions (
 
   CopyMem (&PartBlockDev->BlockInfo, &DevBlockInfo, sizeof (DEVICE_BLOCK_INFO));
 
-  if (DevBlockInfo.BlockSize == 1) {
+  if (MediaGetInterfaceType() == OsBootDeviceMemory) {
```
mauricema, line 543: For both SPI and memory device.

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -551,36 +549,18 @@ FindPartitions (
...
-      // Could not find any partition, so assume no partitions
-      BlockDev              = & (PartBlockDev->BlockDevice[0]);
-      BlockDev->StartBlock  = 0;
-      BlockDev->LastBlock   = DevBlockInfo.BlockNum - 1;
-      PartBlockDev->PartitionType    = EnumPartTypeUnknown;
-      PartBlockDev->PartitionChecked = TRUE;
-      PartBlockDev->BlockDeviceCount = 1;
-      Status = EFI_SUCCESS;
-    }
+  // Could not find any partition, so assume no partitions
```
mauricema, line 552: This is not correct.  Why did you remove all the partition check ?
It is still required for other devices, such as USB, SATA, UFS, etc.  It is a common function for all.  We just need to handle SPI/Mem device specially.  Remaining logic should be kept untouched.

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -551,36 +549,18 @@ FindPartitions (
...
 
-    DEBUG ((DEBUG_INFO, "Partition type: %a  (%d logical partitions)\n", \
-            GetPartitionTypeName (PartBlockDev->PartitionType), \
-            PartBlockDev->BlockDeviceCount));
-  } else {
-    Status = EFI_DEVICE_ERROR;
-  }
+  DEBUG ((DEBUG_INFO, "Partition type: %a  (%d logical partitions)\n", \
+          GetPartitionTypeName (PartBlockDev->PartitionType), \
+          PartBlockDev->BlockDeviceCount));
```
myguitar, line 563: Agree with Maurice. This common logic should be kept.

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
naveenm18: Ok. I considered only SPI/MEM. I will revert this change because if SPI/MEM it is handled earlier.

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -522,6 +522,7 @@ FindPartitions (
   LOGICAL_BLOCK_DEVICE        *BlockDev;
   EFI_PARTITION_TABLE_HEADER  *Gpt;
   PART_BLOCK_DEVICE           *PartBlockDev;
+  OS_BOOT_MEDIUM_TYPE          mCurrentMediaType;
```
myguitar, line 525: Can you remove 'm' prefix?

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -532,6 +533,8 @@ FindPartitions (
     return Status;
   }
 
+  mCurrentMediaType = MediaGetInterfaceType();
```
myguitar, line 536: Can you move this line to right before using the variable?

BootloaderCommonPkg/Library/PartitionLib/PartitionLib.c
```diff
@@ -542,7 +545,7 @@ FindPartitions (
 
   CopyMem (&PartBlockDev->BlockInfo, &DevBlockInfo, sizeof (DEVICE_BLOCK_INFO));
 
-  if (DevBlockInfo.BlockSize == 1) {
+  if (mCurrentMediaType == OsBootDeviceMemory || mCurrentMediaType == OsBootDeviceMemory) {
```
myguitar, line 548: I guess you meant OsBootDeviceSpi.
```
if (mCurrentMediaType == OsBootDeviceMemory || mCurrentMediaType == OsBootDeviceMemory)
->
if ((CurrentMediaType == OsBootDeviceSpi) || (CurrentMediaType == OsBootDeviceMemory))
```

Quick question. Did the original code 'if (DevBlockInfo.BlockSize == 1)' cause a problem? Do we really need this change?

mauricema: Using BlockSize==1 to check SPI/Mem device will work as long as you set block size to 1. But for SPI, we originally have bigger block size.  I think it is always good to use the actual media type for check.

gdong1: OsBootDeviceSpi is mainly used for PDR region. For Memory mapped SPI BIOS region, OsBootDeviceMemory should be used for better performance. What issue for this patch is to solve?

naveenm18: OsBootDeviceSpi is supported for BIOS region also now. This patch if for optimization and improving the code checks.

myguitar: Understood. So, this is code improvement to check actual media type. Please address review comments.

PR #750, opened 2020-06-15
Handling CFGDATA_KEY_ID in CfgDataStitch

lsubash:
Do not check for key file when KEY ID is
used.

### DIFF
--- a/BootloaderCorePkg/Tools/CfgDataStitch.py
+++ b/BootloaderCorePkg/Tools/CfgDataStitch.py
@@ -84,10 +84,12 @@ def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_file, script_dir, tool
 
     # ensure all required files exist
     chk_files = [
-        (script_dir, ['GenCfgData.py', 'CfgDataTool.py']),
-        (os.path.dirname(key_file), [os.path.basename(key_file)]),
-        (cfg_dir, ['CfgDataDef.dsc']),
-    ]
+            (script_dir, ['GenCfgData.py', 'CfgDataTool.py']),
+            (cfg_dir, ['CfgDataDef.dsc']),
+        ]
+    # Check for KEY_ID at end of string
+    if not key_file.endswith('KEY_ID'):
+         chk_files.extend([(os.path.dirname(key_file), [os.path.basename(key_file)])])
 
     result = check_file_exist (chk_files)
     if result:


BootloaderCorePkg/Tools/CfgDataStitch.py
```diff
@@ -82,12 +82,20 @@ def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_file, script_dir, tool
...
+    # Check for KEY_ID at end of string
+    if (key_file.endswith('KEY_ID')):
+        # ensure all required files exist
+        chk_files = [
+            (script_dir, ['GenCfgData.py', 'CfgDataTool.py']),
+            (cfg_dir, ['CfgDataDef.dsc']),
+        ]
+    else:
+        # ensure all required files exist
+        chk_files = [
```
sagraw2, line 94: If a HSM is being used to perform signing, won't this check fail ?
Is it ok if we let the execution continue here and let the failure happen in signing script ?

lsubash: Yes for key_file, it can be handled in single signing script.
I would remove key file check from here for real path/key id.

BootloaderCorePkg/Tools/CfgDataStitch.py
```diff
@@ -82,12 +82,20 @@ def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_file, script_dir, tool
...
+            (script_dir, ['GenCfgData.py', 'CfgDataTool.py']),
+            (os.path.dirname(key_file), [os.path.basename(key_file)]),
+            (cfg_dir, ['CfgDataDef.dsc']),
+        ]
```
myguitar, line 98:
```
    chk_files = [
        (script_dir, ['GenCfgData.py', 'CfgDataTool.py']),
        (cfg_dir, ['CfgDataDef.dsc']),
    ]
    if not key_file.endswith('KEY_ID'):
        chk_files.extend([(os.path.dirname(key_file), [os.path.basename(key_file)])])
```

lsubash: Thanks Aiden

BootloaderCorePkg/Tools/CfgDataStitch.py
sagraw2: @Subash Ignore my original comment.

BootloaderCorePkg/Tools/CfgDataStitch.py
sagraw2: Looks good to me.

BootloaderCorePkg/Tools/CfgDataStitch.py
Subash: > @Subash Ignore my original comment.

Sorry I can't. Github sent me a notification to make sure I look at it.

PR #752, opened 2020-06-16
Make Gpio convert tool common

stalamudupula:
Gpio convert tool can be common and accepts a
platform specific config file with Group Info
and other settings as input.

Example for CFL/WHL:

  Convert gpio.csv into .dsc format:

  python
  ($SBL_ROOT)\SblOpen\Platform\CommonBoardPkg\Tools\GpioDataConvert.py
  -if gpio.csv
  -of dsc
  -cf ($SBL_ROOT)\SblOpen\Platform\CoffeelakeBoardPkg\Script\GpioDataConfig.py

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Script/GpioDataConfig.py
+++ b/Platform/CoffeelakeBoardPkg/Script/GpioDataConfig.py
@@ -1,0 +1,36 @@
+## @ GpioDataConfig.py
+#  This is a Gpio config script for Slim Bootloader
+#
+# Copyright (c) 2020, Intel Corporation. All rights reserved. <BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+
+#
+# The index for a group has to match implementation
+# in a platform specific Gpio library.
+#
+
+def get_grp_info():
+    grp_info ={
+      # Grp     Index
+      'GPP_A' : [ 0x0],
+      'GPP_B' : [ 0x1],
+      'GPP_C' : [ 0x2],
+      'GPP_D' : [ 0x3],
+      'GPP_E' : [ 0x4],
+      'GPP_F' : [ 0x5],
+      'GPP_G' : [ 0x6],
+      'GPP_H' : [ 0x7],
+      'GPP_I' : [ 0x8],
+      'GPP_J' : [ 0x9],
+      'GPP_K' : [ 0xA],
+    }
+    return grp_info
+
+def rxraw_override_cfg():
+    return False
+
+def plat_name():
+    return 'cfl'
--- a/Platform/CommonBoardPkg/Tools/GpioDataConvert.py
+++ b/Platform/CommonBoardPkg/Tools/GpioDataConvert.py
@@ -2,7 +2,7 @@
 #  This is a python script for generating required output format
 #  of Gpio pin information from a given input format
 #
-# Copyright (c) 2019, Intel Corporation. All rights reserved. <BR>
+# Copyright (c) 2019 - 2020, Intel Corporation. All rights reserved. <BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 ##
@@ -11,16 +11,17 @@ import os
 import sys
 import argparse
 import re
+import imp
 from   ctypes import *
 
 DSC_LINE_HDR = "# !BSF SUBT:{GPIO_TMPL:"
 DLT_CFGx_HDR = "GPIO_CFG_DATA.GpioPinConfig"
 DLT_CFG0_HDR = DLT_CFGx_HDR + "0_"
 DLT_CFG1_HDR = DLT_CFGx_HDR + "1_"
-H_STRUCT_HDR = "static GPIO_INIT_CONFIG mGpioTableCnlXXXX[] = {\n"
+H_STRUCT_HDR = "static GPIO_INIT_CONFIG mGpioTable[] = {\n"
 H_STRUCT_FTR = "\n};"
-CSV_FILE_HDR = "GpioPad, PadMode, HostSoftPadOwn, Direction, OutputState, InterruptConfig, PowerConfig, ElectricalConfig, LockConfig\n"
-PAD_NAME_HDR = "GPIO_CNL_XX_"
+CSV_FILE_HDR = "GpioPad, PadMode, HostSoftPadOwn, Direction, OutputState, InterruptConfig, PowerConfig, ElectricalConfig, LockConfig, OtherSettings\n"
+PAD_NAME_HDR = "GPIO_XXX_XX_"
 
 
 # This tool's input/output can be one of the following formats:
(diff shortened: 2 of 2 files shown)


Platform/CommonBoardPkg/Tools/GpioDataConvert.py
```diff
@@ -107,13 +109,21 @@
...
     'GpioIntEdge|GpioIntSci'    : 0x69,
+    'GpioIntSci|GpioIntEdge'    : 0x69,
     'GpioIntEdge|GpioIntApic'   : 0x71,
+    'GpioIntApic|GpioIntEdge'   : 0x71,
     'GpioIntEdge|GpioIntSmi'    : 0x65,
+    'GpioIntSmi|GpioIntEdge'    : 0x65,
     'GpioIntEdge|GpioIntDefault': 0x60,
+    'GpioIntDefault|GpioIntEdge': 0x60,
     'GpioIntLvlEdgDis|GpioIntDis':0xA1,
+    'GpioIntDis|GpioIntLvlEdgDis':0xA1,
```
myguitar, line 126: Better to put each key and value separately in this dictionary. A caller needs to do OR according to input.
ex)
```
GPIO_INIT_CONFIG = {
GpioIntDefault : 0x00,
GpioIntDis : 0x01,
...
GpioIntLvlEdgDis : 0xA0,
...
}
```

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
```diff
@@ -2,7 +2,7 @@
 #  This is a python script for generating required output format
 #  of Gpio pin information from a given input format
 #
-# Copyright (c) 2019, Intel Corporation. All rights reserved. <BR>
+# Copyright (c) 2019 - 2020, Intel Corporation. All rights reserved. <BR>
```
myguitar, line 5: It looks this file moves under CommonBoardPkg/Tools. This is new directory, right?
In my opinion, if this is a common tool, 'BootloaderCorePkg/Tools' would be right place instead of having another location. @mauricema Any comment about the tool/script location?

Platform/CoffeelakeBoardPkg/Script/GpioDataConfig.py
```diff
@@ -0,0 +1,34 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+
+#
+# The index for a group has to match implementation
+# in a platform specific Gpio library.
+#
+
```
myguitar, line 14: Having platform name as a mandatory input would be helpful in common script.

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
mauricema: For GPIO related tool , I guess it is OK to put under Platform folder currently.  
If we do have enough confidence for this script to work for all platforms going forward, I am also fine to promote it to BootloaderCorePkg/Tools.

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
mauricema, line 126: If we keep these ,  I would think better to put script under  Platform/CommonBoardPkg.
GPIO_MIXED_INT_CONFIG is project specific,  and not that common to be in  BootloaderCorePkg.

stalamudupula: GPIO_INT_CONFIG is already provided. But some csv/h files may use this mixed config ,so added this new dict so that I can directly use the ORed config key to get its value.

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
stalamudupula: Will add it in Platform folder for now.

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
stalamudupula: As I said in other comment, I added this only for convenience. Ideally, if there is a mixed config provided in csv/h file, then we need to get the value (from GPIO_INT_CONFIG) for each operand in OR operation and OR the values. This way, we can eliminate GPIO_MIXED_INT_CONFIG. let me add that.

Platform/CoffeelakeBoardPkg/Script/GpioDataConfig.py
stalamudupula: OK. But how is this input going to be used in the common script?

myguitar: if 'apl', it may return 'unsupported'.

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
myguitar: I mean we can use GPIO_INT_CONFIG for mixed one as well.
```
    if '|' not in parts[5]:
        sbl_dw0.Dw0Tmpl.InterruptConfig = GPIO_INT_CONFIG[parts[5]]
    else:
        sbl_dw0.Dw0Tmpl.InterruptConfig = GPIO_MIXED_INT_CONFIG[parts[5]]
```
->
```
for key in parts[5].split(':'):
  sbl_dw0.Dw0Tmpl.InterruptConfig |= GPIO_INT_CONFIG[key]
```

myguitar: typo. split('|')

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
```diff
@@ -585,22 +589,29 @@ def get_sbl_dws (inp_fmt, parts):
...
 
 # Get .h, .csv line from SBL Config DWs
-def get_h_csv_from_sbl_dws (sbl_dw0, sbl_dw1):
+def get_h_csv_from_sbl_dws (sbl_dw0, sbl_dw1, gpio_cfg_file):
     PadMode         = list(GPIO_PAD_MODE.keys())[list(GPIO_PAD_MODE.values()).index(sbl_dw0.Dw0Tmpl.PadMode)]
     HostSoftPadOwn  = list(GPIO_HOSTSW_OWN.keys())[list(GPIO_HOSTSW_OWN.values()).index(sbl_dw0.Dw0Tmpl.HostSoftPadOwn)]
     Direction       = list(GPIO_DIRECTION.keys())[list(GPIO_DIRECTION.values()).index(sbl_dw0.Dw0Tmpl.Direction)]
     OutputState     = list(GPIO_OUTPUT_STATE.keys())[list(GPIO_OUTPUT_STATE.values()).index(sbl_dw0.Dw0Tmpl.OutputState)]
-    if sbl_dw0.Dw0Tmpl.InterruptConfig in list(GPIO_INT_CONFIG.values()):
+    if sbl_dw0.Dw0Tmpl.InterruptConfig <= 0x1F:
```
myguitar, line 597: IMHO, a reversed dictionary would make easy to read code in this function.
```
ex)
reversed_dict = {v:k for k, v in GPIO_INT_CONFIG.items()}
```

PR #753, opened 2020-06-16
Use an variable for defining key type for signing interface

lsubash:
Signing interface accepts KEY ID. Currently default
keys accessed are RSA2048. Based on crypto recommendation
some platfom would need RSA3072 as default.

Use KEY_SIZE_TYPE environment variable to define key type.
Default env setting would be RSA2048 in signing interface.

### DIFF
--- a/BootloaderCorePkg/Tools/SingleSign.py
+++ b/BootloaderCorePkg/Tools/SingleSign.py
@@ -22,7 +22,6 @@ import string
 # KEY_SIZE_TYPE defines the key sizes to be used for signing
 # KEY_SIZE_TYPE = RSA2048 uses RSA 2K size keys
 # KEY_SIZE_TYPE = RSA3072 uses RSA 3K size keys
-KEY_SIZE_TYPE = 'RSA2048'
 
 SIGNING_KEY = {
     # Key Id                    | Key File Name start |
@@ -94,6 +93,16 @@ def run_process (arg_list, print_cmd = False, capture_out = False):
 
     return output
 
+def get_key_sz_type ():
+
+    if 'KEY_SIZE_TYPE' not in os.environ:
+        # default Key size "RSA2048'
+        key_sz_type = 'RSA2048'
+    else:
+        key_sz_type = os.environ.get('KEY_SIZE_TYPE')
+
+    return key_sz_type
+
 def check_file_pem_format (priv_key):
     # Check for file .pem format
     key_name = os.path.basename(priv_key)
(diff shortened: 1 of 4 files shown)


mauricema:
Are we using too many env variables ?    Why can't this be some parameter  ?

Platform/QemuBoardPkg/Script/TestCases/firmware_update.py
```diff
@@ -257,8 +257,9 @@ def main():
     cmd = [ sys.executable,
             'BootloaderCorePkg/Tools/GenCapsuleFirmware.py',
             '-p',  'BIOS', bios_img,
-            '-k',  '../SblKeys/FirmwareUpdateTestKey_Priv_RSA2048.pem',
-            '-o',  '%s/FwuImage.bin' % fwu_dir
+            '-k',  '../SblKeys/FirmwareUpdateTestKey_Priv_RSA3072.pem',
+            '-o',  '%s/FwuImage.bin' % fwu_dir,
+            '-a',  'SHA2_384'
```
mauricema, line 262: can;'t this be determined by the key type ?
EX: 3K => SHA384  2K=>SHA256

BootloaderCorePkg/Tools/SingleSign.py
```diff
@@ -94,6 +93,16 @@ def run_process (arg_list, print_cmd = False, capture_out = False):
 
     return output
 
+def get_key_sz_type ():
```
mauricema, line 96: Do we really need this ?   
Can't this be determined from the  key itself ?

lsubash: We have defined generic key id's as MASTER_KEY_ID. Ideally, signing interface (single singing) should decide key type to used for signing. Currently we support all platforms with same interface script. Legacy platforms use RSA2K and some platforms need RSA3K.
We would need to define a method to intimate signing interface script which key type to be used as RSA2048, RSA3072, ECC or so. Also, we do not wann to maintain different key id's for various key types.

Understand adding a new Env variable increase number of of variable settings. Looking for recommendations here.

Platform/QemuBoardPkg/Script/TestCases/firmware_update.py
lsubash: Our signing interface is generic to support multiple key types and signing hash alg  combinations.
If we restrict, 2k --> SHA256 and 3K--> SHA384, we could adjust this in respective tools.
This adjust is getting done in container due to auth types

BootloaderCorePkg/Tools/SingleSign.py
lsubash: Yes, if we use actual key file this can be handled currently.

lsubash:
As per discussions, planning to add key id's as MASTER_KEY_ID_RSA2048/MASTER_KEY_ID_RSA3072.
Will create a new PR. Closing this request.
Thanks for review comments.

PR #755, opened 2020-06-17
Fix smbiosview issue on Uefi payload

stalamudupula:
Smbios spec advises to use 'Handle' field in the
Type header to get the type information. This patch
updates the Handle field with the 'Type' value to
be unique. Also, update the Entry Point struct to
report the number of Types implemented currently.


BootloaderCorePkg/Library/SmbiosInitLib/SmbiosInitLib.c
```diff
@@ -167,6 +167,7 @@ FinalizeSmbios (
   Type127 = (SMBIOS_TABLE_TYPE127 *)(UINTN)(SmbiosEntry->TableAddress + SmbiosEntry->TableLength);
   Type127->Hdr.Type = SMBIOS_TYPE_END_OF_TABLE;
   Type127->Hdr.Length = sizeof (SMBIOS_STRUCTURE);
+  Type127->Hdr.Handle = Type127->Hdr.Type;
```
mauricema, line 170: can we use NumberOfSmbiosStructures for Handle ?
Handle needs to be unique.

PR #756, opened 2020-06-18
Extend Key Ids to include sign and size types.

lsubash:
KEY IDs are extended to include key type and sizes.
Platforms can configure corresponding RSA2048 and
RSA3072 KEY IDs. Updated tools to adjust hash type
based on key size.

### DIFF
--- a/BootloaderCorePkg/Tools/CfgDataStitch.py
+++ b/BootloaderCorePkg/Tools/CfgDataStitch.py
@@ -87,8 +87,8 @@ def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_file, script_dir, tool
             (script_dir, ['GenCfgData.py', 'CfgDataTool.py']),
             (cfg_dir, ['CfgDataDef.dsc']),
         ]
-    # Check for KEY_ID at end of string
-    if not key_file.endswith('KEY_ID'):
+    # Check for KEY_ID in key file string
+    if not key_file.startswith('KEY_ID'):
          chk_files.extend([(os.path.dirname(key_file), [os.path.basename(key_file)])])
 
     result = check_file_exist (chk_files)
--- a/BootloaderCorePkg/Tools/CfgDataTool.py
+++ b/BootloaderCorePkg/Tools/CfgDataTool.py
@@ -763,6 +763,9 @@ def CmdSign(Args):
     Fd.write (FileData)
     Fd.close ()
 
+    if Args.hash_alg is 'AUTO':
+        Args.hash_alg = adjust_hash_type(Args.cfg_pri_key)
+
     rsa_sign_file (Args.cfg_pri_key, None, Args.hash_alg, Args.sign_scheme, TmpFile, Args.cfg_out_file, True, True)
     if os.path.exists(TmpFile):
       os.remove(TmpFile)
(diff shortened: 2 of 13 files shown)


BootloaderCorePkg/Tools/CfgDataStitch.py
```diff
@@ -87,8 +87,8 @@ def cfgdata_stitch(ifwi_file, ifwi_out_file, cfg_dir, key_file, script_dir, tool
             (script_dir, ['GenCfgData.py', 'CfgDataTool.py']),
             (cfg_dir, ['CfgDataDef.dsc']),
         ]
-    # Check for KEY_ID at end of string
-    if not key_file.endswith('KEY_ID'):
+    # Check for KEY_ID in key file string
+    if not 'KEY_ID' in key_file:
```
mauricema, line 91: do you have specific rules for the KEY_ID name conversion ?  
will '_KEY_ID_' be better ?

BootloaderCorePkg/Tools/CfgDataTool.py
```diff
@@ -921,7 +924,7 @@ def Main():
                             help='Configuration binary file')
     SignParser.add_argument('-o', dest='cfg_out_file', type=str, help='Signed configuration output binary file name to be generated', required=True)
     SignParser.add_argument('-k', dest='cfg_pri_key', type=str, help='Key Id or Private key file (PEM format) used to sign configuration data', required=True)
-    SignParser.add_argument('-a', dest='hash_alg', type=str, choices=['SHA2_256', 'SHA2_384'], help='Hash Type for signing',  default = 'SHA2_256')
+    SignParser.add_argument('-a', dest='hash_alg', type=str, choices=['SHA2_256', 'SHA2_384', None], help='Hash Type for signing',  default = None)
```
mauricema, line 927: Change "None" to "AUTO".   Auto will select hash automatically according to the key used.

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -211,6 +211,17 @@ def run_process (arg_list, print_cmd = False, capture_out = False):
...
 
+# Adjust hash type algorithm based on Public key file
+def adjust_hash_type (pub_key_file):
+    key_type =  get_key_type (pub_key_file)
+    if key_type ==  'RSA2048':
+        hash_type = 'SHA2_256'
+    elif key_type ==  'RSA3072':
+        hash_type = 'SHA2_384'
+    else:
+        hash_type = None
```
mauricema, line 222: Should be an error condition here if  hash_type cannot be recognized ?

BootloaderCorePkg/Tools/CfgDataStitch.py
lsubash: Now, we have defined as **COMPONENT**_KEY_ID_**SIGNTYPESIZE**. Prior it this patch it use end with KEY_ID.
Yes, KEY_ID_**COMPONENT**_**SIGNTYPESIZE** would be better, Will update these namings.

BootloaderCorePkg/Tools/CfgDataTool.py
lsubash: Sure.

BootloaderCorePkg/Tools/CommonUtility.py
lsubash: It would be better if calling methods handle hash_type to None condition. Will update error check at caller end.

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -482,7 +484,7 @@ def ValidateRegistryFormatGuid(Argument):
 
     parser.add_argument('-p',  '--payload', nargs=2, action='append', type=str, required=True, help='Specify payload information including GUID, FileName')
     parser.add_argument('-k',  '--priv_key', dest='PrivKey', type=str, required=True, help='Key Id or Private RSA 2048/RSA3072 key in PEM format to sign image')
-    parser.add_argument('-a',  '--alg_hash', dest='HashType', type=str, choices=['SHA2_256', 'SHA2_384'], default='SHA2_256', help='Hash type for signing')
+    parser.add_argument('-a',  '--alg_hash', dest='HashType', type=str, choices=['SHA2_256', 'SHA2_384', None], default=None, help='Hash type for signing')
```
sagraw2, line 487: Can we add a small description for the meaning of 'None'?
i.e. When None is chosen, the hashing mechanism will depend upon the key length chosen.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -578,7 +578,7 @@ def extract (self, name = '', file_path = ''):
 
             # create header entry
             auth_type_str = self.get_auth_type_str (self.header.auth_type)
-            key_file = 'CONTAINER_KEY_ID' if auth_type_str.startswith('RSA') else ''
+            key_file = 'KEY_ID_CONTAINER_RSA2048' if auth_type_str.startswith('RSA') else ''
```
lsubash, line 581: This Id need to be generated based on auth type while doing container extraction. I would improve this as separate patch,

lsubash:
Update PR as per the review feedback.

PR #759, opened 2020-06-24
Default key generation option with build

lsubash:
Add option -k with SBL build for key generation.
This is to enable user who do not generate keys
for signing as pre-build step.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -96,7 +96,6 @@ def prep_env ():
     check_for_openssl()
     check_for_nasm()
     check_for_git()
-    check_for_slimbootkeydir()
 
     # Update Environment vars
     os.environ['SBL_SOURCE']     = sblsource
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -1020,6 +1021,18 @@ def pre_build(self):
...
             os.environ['PACKAGES_PATH'] += os.pathsep + sbl_dir
 
+        # Generate SblKeys
+        if self._board.KEY_GEN and not os.path.exists(os.environ.get('SBL_KEY_DIR')):
+            print ("Generating default Sbl Keys..")
+            key_gen_tool = os.path.join(sbl_dir, 'BootloaderCorePkg', 'Tools', 'GenerateKeys.py')
+            args_list = ['python', key_gen_tool, '-k', os.environ['SBL_KEY_DIR']]
+            ret = subprocess.call(args_list)
+            if ret:
+                raise Exception  ('Failed to generate keys !')
```
mauricema, line 1031: Added warning when KEY_GEN is set , but the SBL_KEY_DIR exists.

BuildLoader.py
```diff
@@ -1020,6 +1021,18 @@ def pre_build(self):
         if plt_dir != sbl_dir:
             os.environ['PACKAGES_PATH'] += os.pathsep + sbl_dir
 
+        # Generate SblKeys
+        if self._board.KEY_GEN and not os.path.exists(os.environ.get('SBL_KEY_DIR')):
+            print ("Generating default Sbl Keys..")
```
mauricema, line 1026: Print the key path as well. 
print ("Generating default Sbl Keys in directory %s .." %  sbl_key_dir)

lsubash:
Updated PR as per Maurice review comments

PR #765, opened 2020-06-29
Added firmware update trigger script for yocto

raghavag:
Added firmware update trigger script for Linux

This script is intended to be called by firmware update
application to trigger firmware update after receiving
update capsule in Linux. This script is provided as a
reference implementation and does the following:

1. Copies the capsule to a known location where SBL
   will look for (/boot/efi/FwuImage.bin)

2. Signals FW update to SBL using the WMI interface
   provided by SBL. The WMI interface is provided
   by the ASL (https://github.com/slimbootloader/slimbootloader
/blob/master/Platform/CommonBoardPkg/AcpiTables/Dsdt/FwuWmi.asl)
   and linux kernel driver (https://lkml.org/lkml/2020/4/27/1289)

Reboot command was removed from the script, need to be
included if required.

### DIFF
--- a/BootloaderCorePkg/Tools/SblFwUpdate.py
+++ b/BootloaderCorePkg/Tools/SblFwUpdate.py
@@ -1,0 +1,45 @@
+## @ SblFwUpdate.py
+#
+# Copyright (c) 2017-2020, Intel Corporation. All rights reserved.<BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+import os
+import re
+import sys
+import shutil
+import subprocess
+
+def Usage():
+    print ("SblFwUpdate")
+    print ("Usage:")
+    print ("    SblFwUpdate  <Input Capsule File>")
+
+def Main():
+    #
+    # Parse the options and args
+    #
+    argc = len(sys.argv)
+    if argc != 2:
+        Usage()
+        return 1
+
+    if not os.path.exists(sys.argv[1]) :
+        raise Exception ('%s file not found !' % sys.argv[1])
+
+    if not os.path.exists('/boot/efi/'):
+        raise Exception ('/boot/efi/ path not found !')
+
+    shutil.copy(sys.argv[1], '/boot/efi/FwuImage.bin')
+    fwucmd = "/sys/bus/wmi/devices/44FADEB1-B204-40F2-8581-394BBDC1B651/firmware_update_request"
+    ofile = open(fwucmd, "wb")
+    change=subprocess.Popen(["echo","1"], stdout=ofile)
+    ofile.close()
+    output = subprocess.check_output(['cat',fwucmd]).decode()
+    if output[0] is not '1':
+        raise Exception ('Triggering firmware update failed !')
+
+    return 0
+
+if __name__ == '__main__':
+    sys.exit(Main())


BootloaderCorePkg/Tools/SblFwUpdate.py
```diff
@@ -0,0 +1,32 @@
+## @ SblFwUpdate.py
+#
+# Copyright (c) 2017-2020, Intel Corporation. All rights reserved.<BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
```
rprangar, line 6: May need a couple of lines of documentation about the intended usage and where it may need customization.

BootloaderCorePkg/Tools/SblFwUpdate.py
```diff
@@ -0,0 +1,32 @@
...
+import os
+import re
+import sys
+import shutil
+
+def Usage():
+    print "SblFwUpdate"
+    print "Usage:"
+    print "    SblFwUpdate  <Input Capsule File>"
```
mauricema, line 15: we should use py3 style print.  Using print ("...") instead.

BootloaderCorePkg/Tools/SblFwUpdate.py
```diff
@@ -0,0 +1,32 @@
...
+
+def Main():
+    #
+    # Parse the options and args
+    #
+    argc = len(sys.argv)
+    if argc == 2:
+        shutil.copy(sys.argv[1], '/boot/efi/FwuImage.bin')
+        os.system('echo 1 > /sys/bus/wmi/devices/44FADEB1-B204-40F2-8581-394BBDC1B651/firmware_update_request')
```
mauricema, line 24: you might want to check if this file exists before you do "echo 1".
Also better to read it back to ensure the value has actually been changed.  Provide status to end user if this trigger request is successful or not.

BootloaderCorePkg/Tools/SblFwUpdate.py
raghavag: done

BootloaderCorePkg/Tools/SblFwUpdate.py
```diff
@@ -0,0 +1,32 @@
...
+    print "Usage:"
+    print "    SblFwUpdate  <Input Capsule File>"
+
+def Main():
+    #
+    # Parse the options and args
+    #
+    argc = len(sys.argv)
+    if argc == 2:
+        shutil.copy(sys.argv[1], '/boot/efi/FwuImage.bin')
```
gdong1, line 23: Better copy it to "/boot/FwuImage.bin" since SBL firmware update is not related with EFI?

BootloaderCorePkg/Tools/SblFwUpdate.py
raghavag: Done

BootloaderCorePkg/Tools/SblFwUpdate.py
raghavag: Done

BootloaderCorePkg/Tools/SblFwUpdate.py
raghavag: /boot/efi path is required because this way, no changes are needed to SBL code

BootloaderCorePkg/Tools/SblFwUpdate.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if not os.path.exists(sys.argv[1]) :
+        raise Exception ('%s file not found !' % sys.argv[1])
+
+    if not os.path.exists('/boot/efi/'):
+        raise Exception ('/boot/efi/ path not found !')
+
+    shutil.copy(sys.argv[1], '/boot/efi/FwuImage.bin')
+    os.system('echo 1 > /sys/bus/wmi/devices/44FADEB1-B204-40F2-8581-394BBDC1B651/firmware_update_request')
+    output = subprocess.check_output(['cat','/sys/bus/wmi/devices/44FADEB1-B204-40F2-8581-394BBDC1B651/firmware_update_request']).decode()
```
myguitar, line 35: Let's use subprocess instead of os.system which is blocking call.

```
fwu_sysfs = '/sys/bus/wmi/devices/44FADEB1-B204-40F2-8581-394BBDC1B651/firmware_update_request'
subprocess.call(['echo', '1', '>', fwu_sysfs]);
subprocess.check_output(['cat', fwu_sysfs]).decode()
```

BootloaderCorePkg/Tools/SblFwUpdate.py
gdong1: do you mean we hard-coded the path "/boot/efi" in SBL or "/boot/efi" is in the default search path when we only provide capsule name "FwuImage.bin"?

jgutbub: In Yocto WIC images the EFI partition is being mapped to '/boot/efi' directory. The default FWU capsule location is SW partition 0 on a FAT32 FS. The script is just copying the FWU capsule file to the EFI partition which happens to align with the default capsule location for all/most of the platforms. The capsule media is set to USB so user would just have to change that to whatever media they are booting Yocto from and this script will work by default.

gdong1: From @jgutbub , it looks this script is specific for the yocto OS that has mapped "/boot/efi" to EFI partition. If other OS doesn't have this mapping, this script might not work. But it is ok for now to make it works for Yocto at least.

BootloaderCorePkg/Tools/SblFwUpdate.py
raghavag: I could not use call but I used subprocess.Popen

BootloaderCorePkg/Tools/SblFwUpdate.py
raghavag: As mentioned in the commit description, this works only with Linux.

BootloaderCorePkg/Tools/SblFwUpdate.py
raghavag: I could not use call but I used subprocess.Popen, fixed this in new patch

BootloaderCorePkg/Tools/SblFwUpdate.py
gdong1: In the commit msg, "Copies the capsule to a known location where SBL will look for (/boot/efi/FwuImage.bin)." and here it copied to /boot/efi/FwuImage.bin.
From the SBL code, the SBL didn't really look for "/boot/efi/FwuImage.bin", the capsule file from the config data is only "FwuImage.bin". that's why it is very confusing. 
Anyway, I knows the bkm from James on this since there is a mapping in Yocto.

PR #766, opened 2020-07-03
Configdata firmware update svn check

lsubash:
Add support for security version check for config data blob update.
Added an new data struct CDATA_BLOB_ON_FLASH to reference flash data.
Update cfgdatatool.py for svn support.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
@@ -17,6 +17,7 @@
 #include <Library/FirmwareUpdateLib.h>
 #include <Library/ContainerLib.h>
 #include <Library/DecompressLib.h>
+#include <Library/ConfigDataLib.h>
 #include "FirmwareUpdateHelper.h"
 
 /**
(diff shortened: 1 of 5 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
```diff
@@ -584,6 +585,56 @@ IsUpdateComponentForContainer (
...
+  )
+{
+  CDATA_BLOB_ON_FLASH       *CfgBlobCapAddr;
+  CDATA_BLOB_ON_FLASH        CfgBlobFlashAddr;
+  UINT32                     CfgBlobSize;
+  EFI_STATUS                 Status;
+
+  ZeroMem (&CfgBlobFlashAddr, sizeof(CfgBlobFlashAddr));
+  Status = GetComponentInfo (FLASH_MAP_SIG_CFGDATA, (UINT32 *)&CfgBlobFlashAddr, &CfgBlobSize);
+  if (EFI_ERROR (Status)) {
```
gdong1, line 613: SBL CFG data has two copies in BP0 and BP1. GetComponentInfo () will read current CFG data copy. 
I think here it should read the other CFG data copy since the firmware update is updating the CFG data of the other copy.

lsubash: Are these copies of CFGDATA in Redundant A and B are same or different?  I expect it to be same.
The flow looks as, first its checks SVN for CFGDATA in Redundant A and it updates Redundant B CFGDATA.
Next time its check CFGDATA in  Redundant B. As this block is already updated, SVN check passes and updated Redundant A.

gdong1: When updating CFG data B, the SVN from current boot CFG data A is correct since both A and B has same value (e.g. 2), after CFG B SVN is updated to new value (e.g. 3), and boot with CFG data B, the SVN from current boot CFG data would be new SVN (3.)in our logic. 
Ideally, regardless booting from A or B, both should get same SVN number (old one) and has exact same behavior.

lsubash:
I will push updated patch for handling redundant region updates.
Closing this patch.

PR #768, opened 2020-07-07
Fix MultiBoot image loading issue

mauricema:
The current MulitBoot loading code in SBL did not follow the specifiction.
Spec stated "The offset in the OS image file at which to start loading
is defined by the offset at which the header was found, minus
(header_addr - load_addr)". However, the current code always copies from
offset 0 of the image file. It caused exception when loading some valid
mulitboot image.


jgutbub:
I will test this out locally.

PR #776, opened 2020-07-14
Add Multiboot image support to PreOsChecker flow

jgutbub:
The pre-OS checker/payload flow can support more than
just Linux image type launching; there are use cases
for adding multiboot image support to this flow and
there may be others in the future.

### DIFF
--- a/PayloadPkg/OsLoader/OsLoader.c
+++ b/PayloadPkg/OsLoader/OsLoader.c
@@ -611,7 +611,7 @@ StartBooting (
   if ((LoadedImage->Flags & LOADED_IMAGE_LINUX) != 0) {
     if (FeaturePcdGet (PcdPreOsCheckerEnabled) && IsPreOsCheckerLoaded ()) {
       BeforeOSJump ("Starting Pre-OS Checker ...");
-      StartPreOsChecker (GetLinuxBootParams ());
+      StartPreOsChecker ((VOID *)GetLinuxBootParams (), LoadedImage->Flags);
     } else {
       BeforeOSJump ("Starting Kernel ...");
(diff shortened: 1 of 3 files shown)


PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -624,8 +624,13 @@ StartBooting (
       DEBUG ((DEBUG_ERROR, "EntryPoint is not found\n"));
       return RETURN_INVALID_PARAMETER;
     }
-    BeforeOSJump ("Starting MB Kernel ...");
-    JumpToMultibootOs((IA32_BOOT_STATE*)&MultiBoot->BootState);
+    if (FeaturePcdGet (PcdPreOsCheckerEnabled) && IsPreOsCheckerLoaded ()) {
```
gdong1, line 627: For PreOsChecker, will it boot into LOADED_IMAGE_LINUX or LOADED_IMAGE_MULTIBOOT?
I saw it was checked in both path.

jgutbub: The pre-OS checker/payload being used would need to support it; someone else was able to validate their pre-OS checker binary is working as expected with this multiboot flow.

rprangar: Request was to enable PreOsChecker even for MB ELF images to support for example ACRN boot.

gdong1: Got it.

mauricema:
It looks like it can only handle x86 case, not x64 entry point,  So as an enhancement,  maybe we should catch the error condition.  EX:  If SBL is already in x64 mode, but the multi-boot is 32 bit,  then we can either error out saying not supported,  or we can switch back to 32 bit mode before calling into PreOsChecker.   It can be done separately.

PR #781, opened 2020-07-17
Configdata firmware update svn check

lsubash:
Add support for security version check for
config data blob update. SVN is checked
for redundant region which would be updated.

Fixed python warnings in CfgDataTool and GenCapsuleFirmware.

### DIFF
--- a/BootloaderCommonPkg/Include/Library/ConfigDataLib.h
+++ b/BootloaderCommonPkg/Include/Library/ConfigDataLib.h
@@ -59,11 +59,18 @@ typedef struct {
   //
   UINT8   HeaderLength;
   UINT8   Attribute;
-  //
-  // Internal configuration data offset in DWORD from the start of data blob.
-  // This value is only valid in runtime.
-  //
-  UINT16  InternalDataOffset;
+  union {
+    //
+    // Internal configuration data offset in DWORD from the start of data blob.
+    // This value is only valid in runtime.
+    //
+    UINT16  InternalDataOffset;
+    //
+    // Security version number
+    // This is available only in flash. It would be overwritten by CDATA_BLOB.InternalDataOffset in runtime
+    //
+    UINT8   Svn;
+  } ExtraInfo;
   //
   // The total valid configuration data length including this header.
   //
--- a/BootloaderCorePkg/Tools/CfgDataTool.py
+++ b/BootloaderCorePkg/Tools/CfgDataTool.py
@@ -13,19 +13,6 @@ from   CommonUtility import *
 
 CFGDATA_INT_GUID = b'\xD0\x6C\x6E\x01\x34\x48\x7E\x4C\xBC\xFE\x41\xDF\xB8\x8A\x6A\x6D'
 
-class CDATA_BLOB_HEADER(Structure):
-    ATTR_EXTERN = 1 << 0
-    ATTR_MERGED = 1 << 7
-    _pack_ = 1
-    _fields_ = [
-        ('Signature', ARRAY(c_char, 4)),
-        ('HeaderLength', c_uint8),
-        ('Attribute', c_uint8),
-        ('Reserved', ARRAY(c_char, 2)),
-        ('UsedLength', c_uint32),
-        ('TotalLength', c_uint32),
-    ]
-
 class CCfgData:
 
     DEBUG_FLAG_PARSE  = (1 << 0)
@@ -42,7 +29,8 @@ class CCfgData:
             ('Signature', ARRAY(c_char, 4)),
             ('HeaderLength', c_uint8),
             ('Attribute', c_uint8),
-            ('Reserved', ARRAY(c_char, 2)),
+            ('Svn', c_uint8),
+            ('Reserved', ARRAY(c_char, 1)),
             ('UsedLength', c_uint32),
             ('TotalLength', c_uint32),
         ]
@@ -758,12 +746,14 @@ def CmdSign(Args):
         raise Exception("Invalid config binary file '%s' !" % CfgDataFile)
     CfgBlobHeader.Attribute |= CCfgData.CDATA_BLOB_HEADER.ATTR_SIGNED
 
+    CfgBlobHeader.Svn = Args.svn
+
     TmpFile = Args.cfg_in_file + '.tmp'
     Fd      = open (TmpFile, 'wb')
     Fd.write (FileData)
     Fd.close ()
 
-    if Args.hash_alg is 'AUTO':
+    if Args.hash_alg == 'AUTO':
         Args.hash_alg = adjust_hash_type(Args.cfg_pri_key)
 
     rsa_sign_file (Args.cfg_pri_key, None, Args.hash_alg, Args.sign_scheme, TmpFile, Args.cfg_out_file, True, True)
@@ -817,11 +807,11 @@ def CmdReplace(Args):
     CfgBins = bytearray(Fh.read())
     Fh.close()
 
-    CfgHdr = CDATA_BLOB_HEADER.from_buffer(CfgBins)
+    CfgHdr = CCfgData.CDATA_BLOB_HEADER.from_buffer(CfgBins)
     if CfgHdr.Signature != b'CFGD':
         raise Exception("Invalid CFGDATA image file '%s'" % CfgFile)
 
-    if not CfgHdr.Attribute & CDATA_BLOB_HEADER.ATTR_MERGED:
+    if not CfgHdr.Attribute & CCfgData.CDATA_BLOB_HEADER.ATTR_MERGED:
         raise Exception("CFGDATA image file '%s' is not merged yet!" % CfgFile)
 
     # Get flash image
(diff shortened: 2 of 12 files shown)


BootloaderCommonPkg/Include/Library/ConfigDataLib.h
```diff
@@ -74,6 +74,35 @@ typedef struct {
...
+  //
+  // This header Length
+  //
+  UINT8   HeaderLength;
+  UINT8   Attribute;
+  //
+  // Security version number
+  // This is available only in flash. It would be overwritten by CDATA_BLOB.InternalDataOffset in runtime
+  //
+  UINT8   Svn;
```
mauricema, line 92: Please reuse structure CDATA_BLOB
Define a "union" for  "InternalOffset" and "Svn"

PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
```diff
@@ -590,6 +591,67 @@ IsUpdateComponentForContainer (
...
+  // Locate config data blob header info from capsule image
+  CfgBlobCapAddr = (CDATA_BLOB_ON_FLASH *)((UINTN)ImageHdr + sizeof(EFI_FW_MGMT_CAP_IMAGE_HEADER));
+  if (CfgBlobCapAddr->Signature != CFG_DATA_SIGNATURE){
+    DEBUG((DEBUG_INFO, "Config data signature in capsule image does not match\n"));
+    return EFI_NOT_FOUND;
+  }
+
+  //Check SVN
+  if (CfgBlobCapAddr->Svn >= CfgBlobFlashDataPtr->Svn){
+    *SvnStatus = 1;
```
mauricema, line 647: Add checking for SvnStatus is NULL or not.

BootloaderCorePkg/Tools/CfgDataTool.py
```diff
@@ -21,7 +21,8 @@ class CDATA_BLOB_HEADER(Structure):
         ('Signature', ARRAY(c_char, 4)),
         ('HeaderLength', c_uint8),
         ('Attribute', c_uint8),
-        ('Reserved', ARRAY(c_char, 2)),
+        ('Svn', c_uint8),
```
mauricema, line 24: Let us remove the full class CDATA_BLOB_HEADER to avoid duplicated definitions.
We can always use CCfgData.CDATA_BLOB_HEADER instead.

lsubash:
Update PR as per review comments fixed.
Also, updated with fix for container SVN check where Capsule components are retrieved incorrectly. Currently, updating container region signed with different RSA key size would fail.

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
```diff
@@ -59,11 +59,42 @@ typedef struct {
...
+    // This value is only valid in runtime.
+    //
+    UINT16  InternalDataOffset;
+    //
+    // Security version number
+    // This is available only in flash. It would be overwritten by CDATA_BLOB.InternalDataOffset in runtime
+    //
+    UINT8   Svn;
+    // Reserved
+    UINT8   Rsvd;
```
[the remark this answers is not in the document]
lsubash: Thanks Guo. Yes, SVN and RSVD would point to same location.
Specifying bits would also point to same value.  We can use struct so that rsvd can be used for future.

struct{
   UINT8   Svn;
    UINT8   Rsvd;
}

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
```diff
@@ -59,11 +59,42 @@ typedef struct {
...
+  } Cdata_Uni;
```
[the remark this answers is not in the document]
lsubash: Uni means Union. Will update.

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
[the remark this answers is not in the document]
lsubash: Done

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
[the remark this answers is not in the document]
lsubash: Done

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
```diff
@@ -59,11 +59,44 @@ typedef struct {
...
+    UINT16  InternalDataOffset;
+    struct {
+      //
+      // Security version number
+      // This is available only in flash. It would be overwritten by CDATA_BLOB.InternalDataOffset in runtime
+      //
+      UINT8   Svn;
+      // Reserved
+      UINT8   Rsvd;
+    } Addi_Info;
```
mauricema, line 76: You can simplify it using:
union {
    //
    // Internal configuration data offset in DWORD from the start of data blob.
    // This value is only valid in runtime.
    //
    UINT16  InternalDataOffset;
    UINT8    Svn;
}  ExtraInfo;

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
```diff
@@ -72,7 +105,7 @@ typedef struct {
   // The total space for configration data
   //
   UINT32  TotalLength;
-} CDATA_BLOB;
+} CDATA_BLOB_ON_FLASH;
```
mauricema, line 108: We can use a single structure CDATA_BLOB now since it has union.

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
lsubash: Sure. There would be an Reserved byte available which we can add in future.

PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
```diff
@@ -590,6 +591,72 @@ IsUpdateComponentForContainer (
...
+
+  //
+  // Get base address of CFGDATA from  firmware block to update
+  //
+
+  if (FwPolicy.Fields.UpdatePartitionB == 0x1) {
+    Status = GetComponentInfoByPartition(FLASH_MAP_SIG_CFGDATA, TRUE, (UINT32 *) &CfgBlobFlashAddr, &CfgBlobSize);
+  } else if (FwPolicy.Fields.UpdatePartitionA == 0x1) {
+    Status = GetComponentInfoByPartition(FLASH_MAP_SIG_CFGDATA, FALSE, (UINT32 *) &CfgBlobFlashAddr, &CfgBlobSize);
+  }
```
gdong1, line 634: This logic is different with the one to update the whole bios region on reading partition A/B. should we make them aligned?

lsubash: Yes, It looks BIOS region update is getting FV version from current firmware.  It need to be checked for region capsule update would happen.

PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
lsubash: Added SvnStatus check at start of function in Ln: 622.

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
lsubash: Done.

BootloaderCorePkg/Tools/CfgDataTool.py
lsubash: Done.

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
lsubash: Done. Defined union.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -95,79 +95,98 @@ VerifyFwVersion (
...
+    //
+    Status = UpdateStatus(ImageHdr->UpdateHardwareInstance, \
+                          (CapsuleBlVersion->ImageVersion.ProjMajorVersion << 8) | CapsuleBlVersion->ImageVersion.ProjMinorVersion, \
+                          0xFFFFFFFF);
+    if (EFI_ERROR (Status)) {
+      DEBUG((DEBUG_ERROR, "Updating status to reserved region failed: %r\n", Status));
+      return Status;
+    }
+
+    if (CapsuleBlVersion->ImageVersion.SecureVerNum >= CurrentBlVersion->ImageVersion.SecureVerNum) {
```
sagraw2, line 176: How is this SVN different from the SVN stored in BtG BPM ?

lsubash: @sagraw2 This **SecureVerNum** is part of Stage1A flash region which gets included at build image generation as part of ImageVer details. This is not fused and it would be overwritten with flash region update.
**SecureVerNum** happens at capsule firmware update where as Btg BPM SVN will be fused and it would be an boot time check. Ideally **SecureVerNum** and BtgSVN should be in sync.

PR #782, opened 2020-07-20
Tool for External Key Hash store generation

lsubash:
This patch adds initial support for external public
key hash store generation in container format.
It uses layout file as input with usage and key file or id.
Common functionality available is resued for KEYH
component generation. Also componet is aligned to 4K size
to use for firmware component update.

### DIFF
--- a/BootloaderCorePkg/Tools/GenExtKeyHashStore.py
+++ b/BootloaderCorePkg/Tools/GenExtKeyHashStore.py
@@ -1,0 +1,77 @@
+## @ GenExtKeyHashStore.py
+# Tool to generate an public key hash store.
+# Generated binary can be updated using capsule firmware update
+#
+#
+# Copyright (c) 2020, Intel Corporation. All rights reserved.<BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+import argparse
+
+from CommonUtility import *
+from BuildUtility import gen_pub_key_hash_store, HASH_USAGE
+
+
+#
+#  HASH USAGE definitions supported
+#
+#  'PUBKEY_CFG_DATA'     : (1<<9)
+#  'PUBKEY_FWU'          : (1<<10)
+#  'PUBKEY_OS'           : (1<<11)
+#  'PUBKEY_CONT_DEF'     : (1<<12)
+#  'PUBKEY_OEM_0'        : (1<<24)
+#  'PUBKEY_OEM_1'        : (1<<25)
+#  'PUBKEY_OEM_2'        : (1<<26)
+#  'PUBKEY_OEM_3'        : (1<<27)
+#  'PUBKEY_OEM_4'        : (1<<28)
+#  'PUBKEY_OEM_5'        : (1<<29)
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/GenExtKeyHashStore.py
```diff
@@ -0,0 +1,91 @@
...
+##
+
+import os
+import argparse
+import subprocess
+import sys
+import struct
+import binascii
+import string
+from ctypes import *
```
mauricema, line 18: Please remove unnecessary imports, binascii, ctypes, etc

BootloaderCorePkg/Tools/GenExtKeyHashStore.py
```diff
@@ -0,0 +1,91 @@
...
+    key_list = eval ('[%s]' % layout.replace('\\', '/'))
+
+    if args.HashType == 'AUTO':
+        HashType = adjust_hash_type(args.PrivKey)
+
+    gen_pub_key_hash_store (args.PrivKey, key_list, HashType, args.SignScheme, args.svn, args.input_key_dir, args.out_bin)
+
+    # Firmware update need 4K  align files
+    # Align ouput file for 4K size
+    keyh_file_data = get_file_data(args.out_bin)
```
mauricema, line 83: Do you really need to align to 4KB ?  I believe the FWU will still update it correctly as long as the binary is smaller than the space available.

lsubash: Currently we have 4K block aligned check's while getting firmware update info.

@raghavag, 4K size alignment for capsule binary is mandatory?

raghavag: Does not necessarily need to be 4k aligned, please try with the exact size.

lsubash: We have the below check in GetFirmwareUpdateInfo(),

    if (ImageHdr->UpdateImageSize & 0xFFF) {
      DEBUG ((DEBUG_INFO, "capsule payload size is not block aligned!"));
      return EFI_UNSUPPORTED;
    }

raghavag: I just saw the code, if we are updating a component from the flash map, we assume that they are block aligned since all the components are usually 4k aligned. But it is possible to remove this check and update for the size as is.

BootloaderCorePkg/Tools/GenExtKeyHashStore.py
lsubash: Done. Removed imports which are not required.

BootloaderCorePkg/Tools/GenExtKeyHashStore.py
lsubash: Done. Removed  4K align keyhash binary generation.
Also, removed 4K alignment in Fwupdate with PR #798

lsubash:
Updated PR with review feedback

PR #810, opened 2020-08-11
Extend SBL build process to build individual component

mauricema:
Current SBL build script can only be used to build SBL image.
It would be helpful if it can be extended to support build individual
DSC component, such as some other standalone payload. This patch
introduced a new subcommand "build_dsc" to enable this capability.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -45,6 +45,18 @@ def rebuild_basetools ():
         print ("Build BaseTools failed, please check required build environment and utilities !")
         sys.exit(1)
 
+def create_conf (workspace, sbl_source):
+    # create conf and build folder if not exist
+    workspace = os.environ['WORKSPACE']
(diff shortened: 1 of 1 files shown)


myguitar:
Do we have a .dsc we can test on the latest code base?

mauricema:
We don't have any existing payload DSC in the repo.  But I think we can add a helloworld DSC to show how to build a dummy payload using this method.

mauricema:
Added the following PR to show an example on how to use it to build a payload binary.
https://github.com/slimbootloader/slimbootloader/pull/813

myguitar:
The rebuild_basetools() needs to be added in build_dsc while testing #813 .

mauricema:
ok, let me take a look.

mauricema:
Yes, you are right.  Added the missing base tool build in new commit.

PR #814, opened 2020-08-11
Check PPB decode space and assign BAR accordingly

myguitar:
All child devices under a PPB must be in scope of its PPB's decode space.
Therefore, all PPB checks the decode capability and downgrades its child
devices' resources accordingly.

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciIov.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciIov.c
@@ -32,11 +32,10 @@ PciIovParseVfBar (
   IN UINTN          BarIndex
   )
 {
-  UINT32      Value;
-  UINT32      OriginalValue;
-  UINT32      Mask;
-  EFI_STATUS  Status;
-  PCI_ENUM_POLICY_INFO  *EnumPolicy;
+  UINT32                Value;
+  UINT32                OriginalValue;
+  UINT32                Mask;
+  EFI_STATUS            Status;
 
   //
   // Ensure it is called properly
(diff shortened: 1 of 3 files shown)


BootloaderCorePkg/Library/PciEnumerationLib/PciIov.c
```diff
@@ -68,6 +70,16 @@ PciIovParseVfBar (
     return Offset + 4;
   }
 
+  //
+  // All child devices under PPB must be in scope of PPB's decode space.
+  //
+  Parent = PciIoDevice->Parent;
+  if ((Parent != NULL) && (IS_PCI_BRIDGE (&Parent->Pci))) {
+    DecodeMask = Parent->Decodes;
```
mauricema, line 78: Do you think it might be easier to just use DecodeMask for the control? 
EnumPolicy can be used to initialize the top level root bridge DecodeMask.
And then everyone needs to AND with its parent all the way to the root bridge to get its decodes

mauricema: Consider  Bridge1 (Decoding PMEM64/PMEM32) behind Bridge2 (Decoding PMEM32 only),  the device decode should be AND for all its parents, not just its direct parent.

myguitar: > Do you think it might be easier to just use DecodeMask for the control?
> EnumPolicy can be used to initialize the top level root bridge DecodeMask.
> And then everyone needs to AND with its parent all the way to the root bridge to get its decodes

That's good idea! Let me use EnumPolicy for root bridge DecodeMask and simplify downgrade control.

myguitar: > Consider Bridge1 (Decoding PMEM64/PMEM32) behind Bridge2 (Decoding PMEM32 only), the device decode should be AND for all its parents, not just its direct parent.

You are right. Let me fix this. Thanks!

myguitar:
PU has been updated to address review comment.

PR #827, opened 2020-08-27
[CFL] Add VBT file for Coffeelake

raghavag:
This patch added vbt.bsf and vbt.bat for coffeelake. Opensourced
vbt from github does not support DP port, newly added
vbt supports DP.


mauricema:
Looks good to me.

PR #833, opened 2020-09-01
Add SBL setup support for QEMU

mauricema:
This patch set added basic infrastructure to support SBL setup through MicroPython payload module.

### DIFF
--- a/PayloadPkg/OsLoader/OsLoader.c
+++ b/PayloadPkg/OsLoader/OsLoader.c
@@ -632,6 +632,11 @@ StartBooting (
 {
   MULTIBOOT_IMAGE            *MultiBoot;
   EFI_STATUS                 Status;
+  PLD_MOD_PARAM              PldModParam;
+  OS_BOOT_OPTION_LIST       *OsBootOptionList;
+  UINT32                     CurrIdx;
+  CHAR8                     *PathPtr;
+  CHAR8                      ModCmdLineBuf[16];
 
   DEBUG_CODE_BEGIN();
   PrintStackHeapInfo ();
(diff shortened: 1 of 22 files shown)


PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -668,13 +673,36 @@ StartBooting (
...
+          // So build a path like !SETP/???? for command line.
+          CopyMem (ModCmdLineBuf, PathPtr, 6);
+          CopyMem (ModCmdLineBuf + 6, PathPtr + 11, 4);
+          ModCmdLineBuf[15] = 0;
+        } else {
+          ModCmdLineBuf[0] = 0;
+        }
+        PldModParam.CmdLineBuf = ModCmdLineBuf;
+      }
+    }
```
gdong1, line 697: Had better these code is built out by FeaturePcdGet (PcdEnableSetup) for size.

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -1219,11 +1249,15 @@ PayloadMain (
...
-      BootShell    = TRUE;
+    if (OsBootOptionList->RestrictedBoot != 0) {
+      BootShell    = FALSE;
+    } else {
+      if (OsBootOptionList->BootToShell != 0) {
+        BootShell    = TRUE;
+      } else if (DebugCodeEnabled ()) {
+        ShellTimeout = (UINTN)PcdGet16 (PcdPlatformBootTimeOut);
+        BootShell    = TRUE;
+      }
```
gdong1, line 1260: If we boot to shell firstly, then boot to setup, is there any impact?
If it works, the setup boot option is same with other boot options, maybe we don't need add this limitation. and this behavior is also aligned with UEFI bios. (they could boot to shell firstly, then entering setup menu).

PayloadPkg/OsLoader/OsLoader.c
mauricema: We might need another PCD to control the payload module support.   This is different from the Setup support.   
Payload module will enable launching from OsLoader.   Setup support will control if we can edit CFGDATA. 
I can work on it.

PayloadPkg/OsLoader/OsLoader.c
mauricema: I think we can keep this capability.  But it is platform's choice to determine if it needs to set this flag.  
For production that needs this,  I would think the restriction is required for security reason.

mauricema:
submitted new commit to address review comments.

myguitar (approved):
Looks good to me

gdong1 (approved):
Thanks for adding this. Let's start with this firstly :)

PR #835, opened 2020-09-03
Sync IntelFsp2Pkg to support FSP 2.2

jgutbub:
There are several definitions that have
been added to the IntelFsp2Pkg in order
to support FSP v2.2. These changes are
backwards compatible with FSP 2.0 and 2.1.

For more info please see the Intel
FSP External Architecture Specification
v2.2 at the site below:

https://www.intel.com/content/www/us/en/intelligent-systems/intel-firmware-support-package/intel-fsp-overview.html

### DIFF
--- a/IntelFsp2Pkg/Include/FspEas/FspApi.h
+++ b/IntelFsp2Pkg/Include/FspEas/FspApi.h
@@ -1,14 +1,16 @@
 /** @file
   Intel FSP API definition from Intel Firmware Support Package External
-  Architecture Specification v2.0.
+  Architecture Specification v2.0 - v2.2
 
-  Copyright (c) 2014 - 2019, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2014 - 2020, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
 #ifndef _FSP_API_H_
 #define _FSP_API_H_
+
+#include <Pi/PiStatusCode.h>
 
 ///
 /// FSP Reset Status code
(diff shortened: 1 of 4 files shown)


gdong1 (approved):

IntelFsp2Pkg/Include/FspEas/FspApi.h
```diff
@@ -57,12 +141,12 @@ typedef struct {
   /// Pointer to the non-volatile storage (NVS) data buffer.
   /// If it is NULL it indicates the NVS data is not available.
   ///
-  UINT32                      NvsBufferPtr;
+  VOID                        *NvsBufferPtr;
```
mauricema, line 144: This might break the current x64 boot.

jgutbub: Do you want to change all of the VOID* to UINT32? Or if there is a way to use an override package for x64 build? Or some other approach?

mauricema: At this point, let us change all  VOID* and function pointer such as FspEventHandler into UINT32 so as to allow x64 SBL to be able to call 32bit FSP binary properly.

jgutbub: I think I got them all, all of the pointers inside of the structures but not in the function pointer typedef's, is that right?

IntelFsp2Pkg/Include/FspEas/FspApi.h
```diff
@@ -77,9 +161,40 @@ typedef struct {
...
+  UINT8                      Reserved[3];
+  ///
+  /// Length Length of the structure in bytes. The current value for this field is 32.
+  ///
+  UINT32                      Length;
+  ///
+  /// FspEventHandler Optional event handler for the bootloader to be informed of events
+  /// occurring during FSP execution.
+  ///
+  FSP_EVENT_HANDLER           FspEventHandler;
```
mauricema, line 186: This needs to be UINT32 too.

jgutbub: I was not sure, in the EDK2 file it is not a pointer either, does it need to change to UINT32 still?

https://github.com/tianocore/edk2/blob/master/IntelFsp2Pkg/Include/FspEas/FspApi.h#L186

I did change the one on line 168 to UINT32 because it was a pointer.

jgutbub: I updated it now, I see now since the compiler will also treat function pointers the same as other pointers it could cause issue to allocate 32 or 64 bits.

IntelFsp2Pkg/Include/FspEas/FspApi.h
```diff
@@ -38,12 +99,35 @@ typedef struct {
...
+  UINT8                       Reserved[3];
+  ///
+  /// Length Length of the structure in bytes. The current value for this field is 32.
+  ///
+  UINT32                      Length;
+  ///
+  /// FspDebugHandler Optional debug handler for the bootloader to receive debug messages
+  /// occurring during FSP execution.
+  ///
+  UINT32                      FspDebugHandler;
```
jgutbub, line 127: Updated this one as well.

mauricema (approved):
Looks good to me.

PR #837, opened 2020-09-03
[CFL] Fix klockwork issue

Praveenhp7:
This patch adds code to check FitEntry for NULL to avoid null pointer
dereference.

Signed-off-by: Praveen <praveen.hodagatta.pranesh@intel.com>


mauricema (approved):
Looks good.

PR #839, opened 2020-09-04
Fix GCC x64 Crypto boot issue

mauricema:
This patch fixed GCC x64 Crypto boot issue. The issue was related
to the calling convention. "EFIAPI" was missing for several ASM
provided functions.


myguitar (approved):
Looks good to me

lsubash:
Verified for SHA256(Ni) and SHA384(M7) for TGL x64 uising Linux builds, hash operations are going fine.

lsubash (approved):

PR #842, opened 2020-09-09
Fix klockwork scanning issue

Praveenhp7:
This patch add code to check OsBootOptionList for NULL to avoid
NULL pointer dereference.


mauricema (approved):
Looks good.

PR #845, opened 2020-09-11
Add EndOfStage callback in S3 flow

gdong1:
EndOfStage callback is missing in S3 boot flow.
so just add it in case platform need do something.


jgutbub (approved):
Looks good to me.

PR #852, opened 2020-09-18
[CFL] Added support for mutiple VBT

raghavag:
This patch did the following

1) Added common routines LocateVbtByImageId to look for VBT image using
   ImageId provided by configuration data and GetVbtAddress.
2) GetVbtAddress routine will provide abstaction for all platforms
   irrespective of multiple VBT or single VBT used by the platform.
3) LocateVbtByImageId routine is moved from platform local function to
   common package.
4) VbtImageId configuration option defined in QEMU platform config is
   moved to common configuration in CfgData_Common.yaml
5) ApolloLake VBT ID selection is now done using configuration data.
6) Added latest VBT binary for CFL, WHL is using existing VBT.


mauricema (approved):
Looks good.

PR #854, opened 2020-09-18
Clarify SATA and USB HwPart usage

jgutbub:
The SATA and USB HwPart usage for BootOption
config data varies from the other block device
types. This patch adds clarification to explain
the meaning of HwPart for these 2 media types
respectively.


gdong1 (approved):

mauricema (approved):
Looks good.

PR #855, opened 2020-09-18
Add SetDeviceAddr() for device table update

gdong1:
Currently SBL library has GetDeviceAddr() to get the device
address based on device type and instance. This patch adds
SetDeviceAddr() to update a given device type and address so
that platform could update the device table dynamically.


mauricema (approved):
Looks good.

PR #858, opened 2020-09-28
Enable extra including path for build

mauricema:
This patch added support to allow platform to add specific include
folders for the build. All include paths will be relative to the
SBL $(WORKSPACE).

To eanable this feature, please add similar definitions as below
into BoardConfig.py. For example,
  self._EXTRA_INC_PATH = ['Silicon/QemuSocPkg/Include/Fsp']


gdong1 (approved):

myguitar (approved):
Looks good to me

PR #859, opened 2020-09-29
Enhance FSP build path support

gdong1:
Currently the build tool will always find FSP binaries from
Silicon'\self._board.SILICON_PKG_NAME\FspBin folder.
This patch enhance the build tool to support to get FSP from
_FSP_PATH_NAME. If _FSP_PATH_NAME is not specified, the default
behavior is same with current build.

If user wants to use a different one, they could override it
in BoardConfig.py as below:
self._FSP_PATH_NAME  = 'Silicon/Fsp'

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -1089,10 +1089,15 @@ class Build(object):
             raise Exception  ('IPP_HASH_LIB_SUPPORTED_MASK is not set correctly!!')
 
         # check if FSP binary exists
-        fsp_dir  = os.path.join(plt_dir, 'Silicon', self._board.SILICON_PKG_NAME, "FspBin", self._board._FSP_PATH_NAME)
+        if self._board._FSP_PATH_NAME != '':
+            fsp_path_name = self._board._FSP_PATH_NAME
+        else:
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -1089,10 +1089,15 @@ def pre_build(self):
             raise Exception  ('IPP_HASH_LIB_SUPPORTED_MASK is not set correctly!!')
 
         # check if FSP binary exists
-        fsp_dir  = os.path.join(plt_dir, 'Silicon', self._board.SILICON_PKG_NAME, "FspBin", self._board._FSP_PATH_NAME)
+        if self._board._FSP_PATH_NAME != '':
+            fsp_path_name = self._board._FSP_PATH_NAME
+        else:
+            fsp_path_name = os.path.join('Silicon', self._board.SILICON_PKG_NAME, "FspBin", self._board._FSP_PATH_NAME)
```
mauricema, line 1095: line 1095, you don't need to add self._board._FSP_PATH_NAME part since it is empty already.

gdong1: Agree. Let me remove it.

mauricema (approved):
Looks good.

myguitar (approved):
Looks good to me

PR #862, opened 2020-10-07
Expand HECI service to add more API's

raghavag:
This patch expanded HECI service to include send, receive and
reset interface functions. This helps in making firmwareupdatelib.c
and PSDlib common across platforms.

### DIFF
--- a/Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
+++ b/Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
@@ -17,6 +17,7 @@
 #include <Library/MemoryAllocationLib.h>
 #include <Library/TimerLib.h>
 #include <Service/SpiFlashService.h>
+#include <Service/HeciService.h>
 #include <Library/FirmwareUpdateLib.h>
 #include <Library/BootloaderCommonLib.h>
 #include <Library/PchSbiAccessLib.h>
@@ -130,6 +131,10 @@ InitCsmeUpdInputData (
   )
 {
   CSME_UPDATE_DRIVER_INPUT    *CsmeUpdDriverInput;
+  HECI_SERVICE                *HeciService;
+
+  HeciService = NULL;
+  HeciService = (HECI_SERVICE *)GetServiceBySignature (HECI_SERVICE_SIGNATURE);
 
   CsmeUpdDriverInput = (CSME_UPDATE_DRIVER_INPUT *)AllocateZeroPool (sizeof(CSME_UPDATE_DRIVER_INPUT));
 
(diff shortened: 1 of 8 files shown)


raghavag:
Verified CSME firmware update is working in WHL

Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
```diff
@@ -131,6 +133,8 @@ InitCsmeUpdInputData (
 {
   CSME_UPDATE_DRIVER_INPUT    *CsmeUpdDriverInput;
 
+  mHeciService = (HECI_SERVICE *)GetServiceBySignature (HECI_SERVICE_SIGNATURE);
```
myguitar, line 136: This can be a local variable.

raghavag: Modified to local variable.

myguitar: Can you remove 'm' prefix for the local variable?

myguitar (approved):
Other than 'm' local variable prefix, it looks good to me.

raghavag:
> Other than 'm' local variable prefix, it looks good to me.

Done

Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
raghavag: Done

Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
```diff
@@ -142,9 +147,11 @@ InitCsmeUpdInputData (
     CsmeUpdDriverInput->CompareMem       = (VOID *)((UINTN)CompareMem);
     CsmeUpdDriverInput->Stall            = (VOID *)((UINTN)MicroSecondDelay);
     CsmeUpdDriverInput->PciRead          = (VOID *)((UINTN)PciReadBuffer);
-    CsmeUpdDriverInput->HeciReadMessage  = (VOID *)((UINTN)HeciReceive);
-    CsmeUpdDriverInput->HeciSendMessage  = (VOID *)((UINTN)HeciSend);
-    CsmeUpdDriverInput->HeciReset        = (VOID *)((UINTN)HeciResetInterface);
+    if (HeciService != NULL) {
+      CsmeUpdDriverInput->HeciReadMessage  = (VOID *)((UINTN)HeciService->HeciReceive);
+      CsmeUpdDriverInput->HeciSendMessage  = (VOID *)((UINTN)HeciService->HeciSend);
```
gdong1, line 152: Since FirmwareUpdate doesn't link with HeciLib, CsmeUpdDriverInput->HeciReadMessage might be NULL if the service is not available. do you have that check before calling HeciReadMessage ()?

raghavag: These functions would be called from Csme Update library, we have no control over that code. 

I updated new patch to check if these are NULL and return NULL. In the calling function, it would return Unsupported when this function returns NULL

jgutbub (approved):
Looks good to me.

myguitar (approved):

Silicon/CoffeelakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
gdong1: thanks for the update.

gdong1 (approved):

PR #863, opened 2020-10-07
Add MultiBoot support for X64 build

mauricema:
During X64 enabling, there was a pending task to enable 32bit
MultiBoot support. It is not implemented.  This patch added the
support to allow X64 SBL to boot a 32bit MB image through thunking.
As part of this patch, the ThunkLib is separated from the FspApiLib
so that it can be shared by other component.


myguitar (approved):
Looks good to me

PR #870, opened 2020-10-09
Enable mixed payload mode calling

mauricema:
Since SBL could be built into either x86 or x64 mode, and the payload
can also be x86 or x64 mode. When mixed modes are used, it is required
to switch to proper mode first before calling into payload entrypoint.
This patch added this check to switch to required mode before calling
into payload entry point.


myguitar (approved):
Looks good to me

PR #873, opened 2020-10-12
Resync BaseTools with latest EDKII stable tag

mauricema:
This patch synced BaseTools with EDKII stable tag:
  edk2-stable202008


jgutbub (approved):
Looks good to me.

Confirmed that the build worked for some platforms using GCC and confirmed able to stitch and boot a new IFWI based on this BaseTools update (any of the updated build utilities doesn't seem to have caused any issues so far).

gdong1 (approved):

PR #874, opened 2020-10-13
Fix permission check during Spi access

stalamudupula:
Check for Write access permission during
Write/Erase Flash cycles for BIOS/FLAHD regions.


mauricema (approved):
Looks good to me.

PR #876, opened 2020-10-14
[APL] Fix compiler intrinsics link error for NOOPT build

mauricema:
This patch fixed link error for APL NOOPT build due to compiler
intrinsics functions. However, due to APL hardware requirements,
it is not feasible to fit NOOPT build into real flash. This
patch will not fix the NOOPT build error caused by code size issue.

For example, the following error might still occur for APL NOOPT
build:
  Invalid the required fv image size 0xe3b0 exceeds the set fv image
  size 0x6000
The APL SOC requires Stage1A to fit into 32KB. Since FSP-T will take
8KB, it only gives 24KB for SBL Stage1A code. NOOPT build will create
about 56KB for Stage1A, and it is impossible to fit into the layout.

### DIFF
--- a/Silicon/ApollolakePkg/Library/SocInfoLib/SocInfoLib.c
+++ b/Silicon/ApollolakePkg/Library/SocInfoLib/SocInfoLib.c
@@ -60,7 +60,7 @@ GetCpuMaxNbFrequency (
 {
   UINT64 MsrValue;
   MsrValue = AsmReadMsr64 (MSR_PLATFORM_INFO);
-  return (100 * ((MsrValue >> 8) & 0xff));
+  return (100 * (((UINT32)MsrValue >> 8) & 0xff));
 }
 
 /**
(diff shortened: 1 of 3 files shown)


Silicon/ApollolakePkg/Library/SocInfoLib/SocInfoLib.c
```diff
@@ -60,7 +60,7 @@ GetCpuMaxNbFrequency (
 {
   UINT64 MsrValue;
   MsrValue = AsmReadMsr64 (MSR_PLATFORM_INFO);
-  return (100 * ((MsrValue >> 8) & 0xff));
+  return (100 * (((UINT32)MsrValue >> 8) & 0xff));
```
rprangar, line 63: this is still using >>?

mauricema: UINT32  shift  is fine.  Only  UINT64 will cause issue.

rprangar (approved):

PR #879, opened 2020-10-14
Fix default value change issue in ConfigEditor

### LINKED ISSUE #878: Combo option will change in ConfigEditor while scrolling
In ConfigEditor, while scrolling page using mouse middle wheel,  the Combo configuration items will change its default value.    It does not occur if scrolling by dragging scroll bar.

mauricema:
In ConfigEditor, while scrolling page using mouse middle wheel,
the Combo configuration items will change its default value. It
is because Combo control will bind MouseWheel event by default.
To address it, added code to unbind it explicitly.

It fixed #878.


sindhuragrandhi:
Looks good to me

myguitar (approved):

PR #883, opened 2020-10-16
Fix save full delta file issue in ConfigEditor

### LINKED ISSUE #882: Same results produced from both Save full cfg data vs save cfg changes to dlt file
In ConfigEditor.py, the same output file is generated with these 2 options:

1. Save full cfg data to delta file.
2. Save cfg changes to delta file.

It should result in 2 different output files.

mauricema:
Current ConfigEditor produces the same result for
"Save full cfg data" vs "save cfg changes".  This patch provided
a fix for it.  It fixed #882.


sindhuragrandhi (commented):
Looks good to me

myguitar (approved):

PR #884, opened 2020-10-17
Fix duplicated PlatformId in DLT file

mauricema:
Using ConfigEditor to save back DLT changes, it will produce two
PlatformId lines in DLT file. This patch removed the extra line.


myguitar (approved):
Looks good to me

PR #892, opened 2020-10-31
Enable build without FSP binary

mauricema:
Current SBL build needs the actual FSP binary to pass the build.
This patch enables a mode to test the build without the actual FSP
bianry. It is useful for test before the FSP binary is available.
It is controlled by HAVE_FSP_BIN option. It can be overriden in
BoardConfig.py file.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -916,7 +916,7 @@ class Build(object):
         # Patch flashmap to indicate boot parititon
         fo = open(os.path.join(self._fv_dir, 'STAGE1A_B.fd'), 'r+b')
         bins = bytearray(fo.read())
-        fmapoff = (bytes_to_value(bins[-8:-4]) + len(bins)) & 0xFFFFFFFF
+        fmapoff = bytes_to_value(bins[-8:-4]) - bytes_to_value(bins[-4:]) + self._board.STAGE1A_FV_OFFSET
         fmaphdr = FLASH_MAP.from_buffer (bins, fmapoff)
         if fmaphdr.sig != FLASH_MAP.FLASH_MAP_SIGNATURE:
             raise Exception ('Failed to locate flash map in STAGE1A_B.fd !')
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -1119,16 +1120,17 @@ def pre_build(self):
...
-                    raise Exception ('Required minimum FSP revision is 0x%08X, but current revision is 0x%08X !' %
-                                (self._board.MIN_FSP_REVISION, revision))
+                        raise Exception ('Expecting FSP ImageId: %s, but got %s !' % (self._board.FSP_IMAGE_ID, imageid))
+                revision = get_fsp_revision (fsp_bin)
+                if revision < self._board.MIN_FSP_REVISION:
+                        raise Exception ('Required minimum FSP revision is 0x%08X, but current revision is 0x%08X !' %
+                                    (self._board.MIN_FSP_REVISION, revision))
             setattr(self._board, '%s_SIZE' % each, get_fsp_size(fsp_bin) if self._board.HAVE_FSP_BIN else 0)
-            setattr(self._board, '%s_UPD_SIZE' % each, get_fsp_upd_size(fsp_bin) if self._board.HAVE_FSP_BIN else 0)
+            setattr(self._board, '%s_UPD_SIZE' % each, get_fsp_upd_size(fsp_bin) if self._board.HAVE_FSP_BIN else 1)
```
myguitar, line 1133: Do we need 'else 1'?

mauricema: Yes,  we need it.  If the UPD size is 0, the build will fail because:
UINT8    FspmUpd[FixedPcdGet32 (PcdFSPMUpdSize)];  
We have to keep it a non-zero value.

myguitar: Got it. Thanks for the explanation.

myguitar (approved):
Looks good to me

mauricema:
Fixed indent issue.

PR #893, opened 2020-11-03
Fix the compiler flags for MSVC 2017 and 2019

mgherzan:
* Enable frame pointers (via /Oy-) for 32-bit modules. From the
documentation: "If you specify a debug compiler option (/Z7, /Zi, /ZI),
we recommend that you specify the /Oy- option after any other
optimization compiler options."
* /O1b2s is equivaled to /O1 (/Ob2 and /Os are included in /O1)
* /Z7 for debug info is obsolete, replace with /Zi

Signed-off-by: Mircea Gherzan <mircea.gherzan@intel.com>


mauricema (approved):
These changes looks good to me.   It will have global impacts to all projects.
We will need to see if it will cause issue for other projects or not.

myguitar (approved):
Looks good to me.
As a next step, an override file in BootloaderCorePkg can be an option rather than touching BaseTools. We also have some compile flags in BootloaderCorePkg.dsc which can be merged together.

PR #894, opened 2020-11-03
DEBUG() macro with a smaller code footprint

mgherzan:
The original DEBUG() macro expanded to three function calls. And there
are around 3000 calls of this macro in the SBL code base. So this is
clearly a waste of .text section space and thus FV space. The code
footprint of the original macro forces the use of aggressive compiler
flags to meet the size requirements, flags which hurt the debugging
experience.

The solution is to reduce the DEBUG() macro to a single function call.
This might result in some performance impact (only on debug builds)
because of possible "useless" calls that quickly return, but this
performance impact is negligible compared to the impact of the serial
output. Even more important than that: this smaller DEBUG() paves the
way for less aggressive compiler optimizations which in turn lead to a
better source-level debugging experience.

Signed-off-by: Mircea Gherzan <mircea.gherzan@intel.com>

### DIFF
--- a/MdePkg/Include/Library/DebugLib.h
+++ b/MdePkg/Include/Library/DebugLib.h
@@ -367,12 +367,7 @@ DebugPrintLevelEnabled (
 
 **/
 #if !defined(MDEPKG_NDEBUG)
-  #define DEBUG(Expression)        \
-    do {                           \
-      if (DebugPrintEnabled ()) {  \
-        _DEBUG (Expression);       \
-      }                            \
-    } while (FALSE)
+  #define DEBUG(Expression) DebugPrint Expression
 #else
   #define DEBUG(Expression)
 #endif
--- a/BootloaderCommonPkg/Library/BootloaderDebugLib/DebugLib.c
+++ b/BootloaderCommonPkg/Library/BootloaderDebugLib/DebugLib.c
@@ -57,6 +57,13 @@ DebugPrint (
   ASSERT (Format != NULL);
 
   //
+  // Check that debug print is enabled at all
+  //
+  if (!DebugPrintEnabled ()) {
+    return;
+  }
+
+  //
   // Check driver debug mask value and global mask
   //
   if ((ErrorLevel & GetDebugPrintErrorLevel ()) == 0) {


MdePkg/Include/Library/DebugLib.h
```diff
@@ -367,12 +367,7 @@ DebugPrintLevelEnabled (
 
 **/
 #if !defined(MDEPKG_NDEBUG)
-  #define DEBUG(Expression)        \
-    do {                           \
-      if (DebugPrintEnabled ()) {  \
-        _DEBUG (Expression);       \
-      }                            \
-    } while (FALSE)
+  #define DEBUG(Expression) DebugPrint Expression
```
mauricema, line 370: I think this change might need a bit more discussion.

Doing this might prevent compiler from optimizing the code for SBL release build.  When debug level PCD is set properly for release build, the whole _DEBUG() call could be removed from binary.  But with the new change,  it is handled inside the DebugPrint, so DebugPrint funciton will always be linked in, which cause negative size impacts.   

To confirm my analysis, I did a SBL release build check on APL project, and this PR change did increase the image size.

Before your change:
FV Space Information
STAGE1A [55%Full] 24576 total, 13648 used, 10928 free
STAGE1B [73%Full] 73728 total, 54304 used, 19424 free
STAGE2 [47%Full] 348160 total, 163720 used, 184440 free
OSLOADER [55%Full] 319488 total, 177744 used, 141744 free
FWUPDATE [96%Full] 126976 total, 123120 used, 3856 free
Build report can be found at W:\SlimBoot\SlimFork\Report.log

After your changes:
FV Space Information
STAGE1A [58%Full] 24576 total, 14448 used, 10128 free
STAGE1B [79%Full] 73728 total, 58784 used, 14944 free
STAGE2 [50%Full] 348160 total, 175144 used, 173016 free
OSLOADER [63%Full] 319488 total, 201808 used, 117680 free
FWUPDATE [98%Full] 147456 total, 145232 used, 2224 free
Build report can be found at W:\SlimBoot\SlimFork\Report.log

Please make sure your change does not impact SBL release build size.

mgherzan: Shouldn't there be a macro to disable the DEBUG() macros *altogether* in Release builds? I was and still am assuming that MDEPKG_NDEBUG should be defined for the Release build.

mauricema: It is the case for EDKII.  But for SBL, DEBUG macro is always compiled in.  In this way, it will provide some critical message even in release mode.  Similarly SBL does not silently ignore ASSERT() in release build either.  We use the debug print level mask to control how much message should be carried in DEBUG and RELEASE build.  With RELEASE build, the mask is set to a very high level so that only very few message will be printed out.  All other DEBUG print will be removed by the compiler.

mgherzan: Interesting :) Two questions please:

1. What's the macro that differentiates between DEBUG and RELEASE builds in SBL?
1. Where can I get that size report?

mauricema: Currently we don't have special macro to be passed into compiler command line to indicate DEBUG or RELEASE build.  Instead, it uses the standard PCDs to get those info, such as PcdFixedDebugPrintErrorLevel, PcdDebugPropertyMask, etc.  The SBL code uses DebugCodeEnabled() to determine if we are in debug mode.

When you do a build, the build log will print the size info for each FV. The info is close to the end of the build log.  Please look for "FV Space Information" string.

mgherzan: Thanks, Would it be OK to add such a macro (/DDEBUG_BUILD) to indicate that it's a debug build? If yes, where would you prefer it?

mauricema: Ideally we want to avoid adding new /D flags.  We want to enforce SBL code to reuse the existing infrastructure to handle DEBUG/RELEASE.  I am concerned if this /D is added, then people will start to use "#ifdef DEBUG_BUILD" in their code. It makes it hard to converge into existing interfaces.  

What issue are you trying to address with PR ?  Only for size reduction ?  How much size does it reduce in DEBUG build with this PR ?

mgherzan: Got your point, it's good that the you wanna keep folks away from these #ifdefs in the actual code. I'll try to find something based on the PCDs. Yes, the purpose of the patch is size reduction in the case of even less optimized builds. In the current debug build it does not reduce the size by much, as the optimization is still aggressive:
```
Without the patch (current upstream MSVC flags):
STAGE1A  [48%Full]  49152 total,  23888 used,  25264 free
STAGE1B  [25%Full] 339968 total,  85136 used, 254832 free
STAGE2   [43%Full] 700416 total, 302280 used, 398136 free
OSLOADER [80%Full] 319488 total, 256128 used,  63360 free
FWUPDATE [99%Full] 180224 total, 178560 used,   1664 free

With the patch (current upstream MSVC flags):
STAGE1A  [48%Full]  49152 total,  23888 used,  25264 free
STAGE1B  [25%Full] 339968 total,  85168 used, 254800 free
STAGE2   [43%Full] 700416 total, 302280 used, 398136 free
OSLOADER [80%Full] 319488 total, 256000 used,  63488 free
FWUPDATE [98%Full] 180224 total, 178400 used,   1824 free
```
But this will make a difference if we want to switch from **/O1** to **/Og /Os**, which is less aggressive. Without this patch, the build fails:
```
Without the patch (/Og /Os):
STAGE1A  [52%Full]  49152 total,  25680 used,  23472 free
STAGE1B  [27%Full] 339968 total,  93232 used, 246736 free
STAGE2   [44%Full] 700416 total, 310632 used, 389784 free
OSLOADER [88%Full] 319488 total, 282976 used,  36512 free
FWUPDATE [99%Full] 196608 total, 196416 used,    192 free
Exception: File 'PAYLOAD.lz' size 0x28C61 is greater than padding size 0x28000 !

With the patch (/Og /Os):
STAGE1A  [51%Full]  49152 total,  25232 used,  23920 free
STAGE1B  [25%Full] 339968 total,  88336 used, 251632 free
STAGE2   [43%Full] 700416 total, 304968 used, 395448 free
OSLOADER [83%Full] 319488 total, 267168 used,  52320 free
FWUPDATE [99%Full] 184320 total, 182848 used,   1472 free
```

mauricema: In this specific examples,  if you want to try to explore /Os and /Og,  I am thinking the possible path could be:

- Try to reduce DEBUG() print/ASSERT() statement.  In SBL, I feel we have too many unnecessary DEBUG print and ASSERT eating up flash space. By doing so, it should be able to reduce the DEBUG build size.

- Consider to increase the SBL flash layout for debug build so as to accommodate the increased size (if it is not too big change)
   EX:  you can adjust PAYLOAD_SIZE in BoardConfig.py to provide a little bit more space for payload binary.

mauricema:
Since this PR needs more work here,  close it for now.    We can reopen it when new updates are ready.

PR #896, opened 2020-11-06
Update Microcode path to support customization

gdong1:
Microcode module locates in different place.
so each platform could specify the path by
MICROCODE_INF_FILE in BoardConfig.py.
By default, it uses the same path if there
is no MICROCODE_INF_FILE defined in platform.


mauricema (approved):
Looks good to me.

myguitar (approved):
Looks good to me

PR #897, opened 2020-11-09
Fix MicroCode path

gdong1:
Use the default path only when MICROCODE_INF_FILE
is not defined in platform board. Currently it will
override the one defined in platform.


jgutbub (approved):
Looks good to me.

PR #904, opened 2020-11-21
Add legacy EF segment memory config

mauricema:
Current SBL code will build pointers in E/F segment for ACPI
and SMBIOS table. On some platforms, E/F segment is not supported.
So a new configuration ENABLE_LEGACY_EF_SEG is added.  When
it is enabled, SBL will not use legacy E/F segment memory.


gdong1 (approved):

myguitar (approved):
Looks good to me

PR #906, opened 2020-11-23
Config Editor: Update default yaml file search dir

chinkeongang:
User experience improvement:
If there is no yaml found in current directory
then default to start browsing at platform directory

Signed-off-by: Chin Keong Ang <chin.keong.ang@intel.com>


mauricema (approved):
Looks good.

PR #907, opened 2020-11-24
Fix hang at STAGE1B when compile with MSVC 2017 for IA32 Debug

chinkeongang:
Removed the Oy- flag for DEBUG_VS2017_IA32_CC_FLAGS
which was introduced recently in commit 9f3d2b5fc39b45f34ce22d462b25d24c1d559ed8
it causes hang when calculating SHA384 by CryptoLib call by SecureBootLib driver

Signed-off-by: Chin Keong Ang <chin.keong.ang@intel.com>


mauricema (approved):
Looks good.

PR #911, opened 2020-12-01
Set up initial CI with Azure Pipelines

mauricema:
Migrated to Azure Pipelines for SBL CI to align with EDK2 project.
This is an initial commit to enable basic check. Currently it will
verify patch format and QEMU GCC build and test.


mauricema:
This patch will be manually merged since it will cause CI change.

PR #916, opened 2020-12-02
Allow PatchCheck only in PR condition

myguitar:
The master build trigger does not have PullRequest variable.
This will allow PatchCheck to be run only in PR condition.


mauricema (approved):

mauricema:
Thank you for improving it.   :)

myguitar:
> Thank you for improving it. :)

No problem.

PR #920, opened 2020-12-06
Add EXT4 to print in 'boot' command

jgutbub:
Since the EXT library we have in OS Loader
supports 2, 3, and 4 it will help to reduce
confusion for users who have noticed that
only 2 and 3 are currently printed.


mauricema (approved):
Looks good.

PR #922, opened 2020-12-09
[CFL]: Fix onboard display issue on CFL-S & CFL-H

Praveenhp7:
As per board design the DDI ports are routed as per below,

CFL-S:
DDI1-->PORTB-->LSPCON-->HDMI
DDI2-->PORTC-->DP
CFL-H:
DDI1-->PORTB-->DP1
DDI2-->PORTC-->DP2
DDI3-->PORTD-->LSPCON-->HDMI

Hence add separate VBT binaries accordingly.

-Set HPD of DDI port F to disable as default since it's not used by
 H/S/U processor line.

TEST: verified onboard display's ports booting to UEFI shell & Yocto.


Platform/CoffeelakeBoardPkg/CfgData/CfgData_GpuConfig.yaml
```diff
@@ -95,7 +95,7 @@
       help         : >
                      Enable or disable
       length       : 0x01
-      value        : 0x1
+      value        : 0x0
```
myguitar, line 98: The default value change can impact all boards. What about WHL? Just in case, we can keep this and update cflh/cfls delta files.

Praveenhp7: -Each DDC(B,C,D,E ) and HPD(B,C,D,E) signals  are valid for each DDI ports ( B,C,D,E). 
as per EDS doc, U-processor can only support 1 Edp and 2 DDI ports. Hence Port F Hot plug detect (HPD)  may not valid for WHL.

WHL-U CRB comes with 2 Display ports and I have tested both found working with this change.

mauricema (approved):
Looks good

Platform/CoffeelakeBoardPkg/CfgData/CfgData_GpuConfig.yaml
myguitar: Understood. Thank you for the confirmation.

myguitar (approved):
Looks good to me

PR #927, opened 2020-12-14
Openssl 1.1.1i tool support

lsubash:
Openssl 1.1.1 is an installable and it creates
openssl.exe in a bin folder.
default slimboot path: c:\Openssl\bin\openssl.exe
Download location: https://slproweb.com/products/Win32OpenSSL.html

Added support to populate OPENSSL_PATH
from bin dir incase its available.

### DIFF
--- a/BootloaderCorePkg/Tools/SingleSign.py
+++ b/BootloaderCorePkg/Tools/SingleSign.py
@@ -63,11 +63,15 @@ MESSAGE_SBL_KEY_DIR = (
 def get_openssl_path ():
     if os.name == 'nt':
         if 'OPENSSL_PATH' not in os.environ:
-            os.environ['OPENSSL_PATH'] = "C:\\Openssl\\"
-        if 'OPENSSL_CONF' not in os.environ:
-            openssl_cfg = "C:\\Openssl\\openssl.cfg"
-            if os.path.exists(openssl_cfg):
-                os.environ['OPENSSL_CONF'] = openssl_cfg
+            openssl_dir = "C:\\Openssl\\bin\\"
+            if os.path.exists (openssl_dir):
+                os.environ['OPENSSL_PATH'] = openssl_dir
+            else:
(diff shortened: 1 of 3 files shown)


BuildLoader.py
```diff
@@ -92,8 +92,6 @@ def prep_env ():
             sys.exit(1)
         if 'NASM_PREFIX' not in os.environ:
             os.environ['NASM_PREFIX'] = "C:\\Nasm\\"
-        if 'OPENSSL_PATH' not in os.environ:
-            os.environ['OPENSSL_PATH'] = "C:\\Openssl\\"
```
lsubash, line 96: We are setting default path in get_openssl_path() incase user do not set it.

BootloaderCorePkg/Tools/SingleSign.py
```diff
@@ -63,7 +63,11 @@
...
     if os.name == 'nt':
         if 'OPENSSL_PATH' not in os.environ:
-            os.environ['OPENSSL_PATH'] = "C:\\Openssl\\"
+            openssl_dir = "C:\\Openssl\\bin\\"
+            if os.path.exists (openssl_dir):
+                os.environ['OPENSSL_PATH'] = openssl_dir
+            else:
+                os.environ['OPENSSL_PATH'] = "C:\\Openssl\\"
         if 'OPENSSL_CONF' not in os.environ:
             openssl_cfg = "C:\\Openssl\\openssl.cfg"
```
mauricema, line 72: is this still correct for CFG file ?

lsubash: With latest openssl installer, default CFG file is installed at C:\Program Files\Common Files\SSL\
It looks with openssl_conf is not mandatory for Slimboot operations.

Updated PR to set openssl conf with older versions only.

mauricema (approved):
Looks good to me.

PR #934, opened 2020-12-22
Use Python lz4 module for LZ4 compression

jgutbub:
To make SBL scripts/tools more OS-agnostic the
lz4 compression module can be used instead of
relying on the BaseTools LZ4 package/executable.

TEST=Confirmed that the python lz4.block.compress
     routine is compatible with Lz4DecompressLib
     during SBL runtime to decompress LZ4 binaries.


jgutbub:
Just realized it does not handle the decompression part if you are extracting container images, let me work on that and re-submit.

PR #935, opened 2020-12-22
Use Python lz4 module for LZ4

jgutbub:
To make SBL scripts/tools more OS-agnostic the
lz4 (de)compression module can be used instead of
relying on the BaseTools LZ4 package/executable.

Need to update the azure pipeline to install
the python lz4 module before running the builds.

TEST=Confirmed that the python lz4.block.compress
     routine is compatible with Lz4DecompressLib
     during SBL runtime to decompress LZ4 binaries.


mauricema (approved):
Looks good to me.

PR #940, opened 2021-01-04
Make Python lz4 module optional

jgutbub:
To avoid any dependency on non-standard python modules
it is preferred to make the binary Lz4 compression the
default flow and try the python module as a backup. Also
recommend to use a known working version of the lz4
python module, 3.1.1.


gdong1 (approved):

mauricema (approved):
Looks good.

PR #942, opened 2021-01-05
Removed deprecated python imp module usage

mauricema:
Python 3.4 and above have deprecated imp module in favor of
importlib.  This patch removed imp module usage from SBL, and
used importlib instead.

### DIFF
--- a/Platform/CometlakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/CometlakeBoardPkg/Script/StitchIfwi.py
@@ -9,7 +9,6 @@
 import sys
 import os
 import re
-import imp
 import struct
 import argparse
 import zipfile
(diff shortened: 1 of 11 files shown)


Platform/CometlakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -441,7 +440,7 @@ def main():
     ap.add_argument('-fusa', dest='fusa', action = "store_true", default = False, help = "Patch IFWI to generate Fusa ifwi")
     args = ap.parse_args()
 
-    stitch_cfg_file = imp.load_source('StitchIfwiConfig', args.config_file)
+    stitch_cfg_file = load_source('StitchIfwiConfig', args.config_file)
     if args.work_dir == '':
```
gdong1, line 444: Please double check if load_source is available since StitchIfwi.py doesn't import utility modules.
maybe use SourceFileLoader if it is not available.

myguitar (approved):
Looks good to me

Platform/CometlakeBoardPkg/Script/StitchIfwi.py
mauricema: yes, it is available.  CommonUtility is indirectly imported.

gdong1 (approved):

PR #944, opened 2021-01-06
Check for '.exe' in run_process routine

jgutbub:
run_process is used in stitching scripts
and other places and when the command being
called might also have a similarly named
command in the same folder (ex. fit and fit.exe)
where one command is meant to be ran in Linux
and the other is meant to be ran in Windows, there
can be issues that require to check for which
OS the stitching script is running in to append
'.exe' if Windows is detected. Rather than change
it every where else we can change it in the
run_process routine to check for Windows and if
'.exe' is not already included in the command we
can append it.

### DIFF
--- a/BootloaderCorePkg/Tools/CommonUtility.py
+++ b/BootloaderCorePkg/Tools/CommonUtility.py
@@ -206,6 +206,10 @@ def get_openssl_path ():
 
 def run_process (arg_list, print_cmd = False, capture_out = False):
     sys.stdout.flush()
+
+    if os.name == 'nt' and os.path.splitext(arg_list[0])[1] == '':
+        arg_list[0] += '.exe'
+
     if print_cmd:
         print (' '.join(arg_list))
 
--- a/BootloaderCorePkg/Tools/GenerateKeys.py
+++ b/BootloaderCorePkg/Tools/GenerateKeys.py
@@ -96,8 +96,10 @@ def generate_rsa_priv_keys (openssl_path, key_dir, key_size):
                         replace_all = 'yes'
 
                 print("Generating private key %s" % priv_key_name)
-                cmd = '%s genrsa -F4 -out %s %s' % (openssl_path, priv_key_name, key_sz)
-                run_process (cmd.split())
+                cmd = 'genrsa -F4 -out %s %s' % (priv_key_name, key_sz)
+                cmd = cmd.split()
+                cmd.insert(0, '%s' % (openssl_path))
+                run_process (cmd)
                 print('\n' * 3)
 
         # Generate test public keys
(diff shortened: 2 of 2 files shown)


BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -206,6 +206,10 @@ def get_openssl_path ():
 
 def run_process (arg_list, print_cmd = False, capture_out = False):
     sys.stdout.flush()
+
+    if os.name == 'nt' and '.exe' not in arg_list[0]:
+        arg_list[0] += '.exe'
```
gdong1, line 211: Better check if there is file extension exist instead of checking '.exe' in arg_list.
For example, if a test.bat given, it should not append .exe.

mauricema: There might be more cases that need to be handled.  For example,  if the executable is "test.cmd",  then this logic will break it.  Similar for other Windows script, like vbs, ps, etc.

gdong1: How about change to the line below?
if os.name == 'nt' and os.path.splitext(arg_list[0])[1] == '':

mauricema: Yes, that should work better.

gdong1 (approved):

mauricema (approved):
Looks good to me.

jgutbub:
Let me see about resolving the openssl path in this patch and if we need to separate it into a second patch I will do that; I cannot really resolve it locally because my openssl path is not in the Program Files location that it seems to be in the Azure environment.

BootloaderCorePkg/Tools/GenerateKeys.py
```diff
@@ -120,14 +122,18 @@ def generate_rsa_pub_keys (openssl_path, key_dir, key_size, replace_key):
...
+            run_process (cmd)
 
             # Extract public key from private key
             pub_key_name = '%s/%s_Pub_RSA%s.pem' % (key_dir, key_file_name, key_size)
             print("Generating OS Public Key %s" % pub_key_name)
-            cmd = '%s rsa -pubout -in %s -out %s' % (openssl_path, priv_key_name, pub_key_name)
-            run_process (cmd.split())
+            cmd = 'rsa -pubout -in %s -out %s' % (priv_key_name, pub_key_name)
+            cmd = cmd.split()
+            cmd.insert(0, '%s' % (openssl_path))
```
jgutbub, line 135: If there is a more optimal way to split the text up without modifying the openssl part please let me know, this is just what I came up with.

gdong1: I guess there is space in openssl_path, so that run_process doesn't really get the file extension, right?
If openssl_path is got from get_openssl_path(), we only need change get_openssl_path().

jgutbub: I don't think that will help in this case, in Azure environment it looks like openssl is located in "C:\Program Files", how can we remove the space there? Maybe I am not understanding though, let me cancel this PR so that you can get it done more quickly than me, I don't want to hold this up.

gdong1: I think the failure should be caused by run_process adding ".exe" wrongly. we could enhance run_process to avoid that.

PR #945, opened 2021-01-07
Add '.exe' extension if missing for windows build

gdong1:
run_process is used several places. Sometime
the extension name is not given for windows build.
In this case, if same file without file extension
exist (e.g. fit and fit.exe), then fit will be used
for windows build. It is not expected, so this patch
will add the missing .exe extension for this case.


mauricema (approved):
Looks good to me.

PR #946, opened 2021-01-07
Update correct BarType field for VF Bar program

stalamudupula:
ProgramBar() routine uses 'OrgBarType' field to
determine BarType. So, if a BarType is valid, copy
it to OrgBarType for VF also.


mauricema (approved):
Looks good.

PR #948, opened 2021-01-08
Update traditional Linux load position

gdong1:
One boot option in SBL could load multiple OS image
including pre-OS image, OS image, extra image, etc.
The traditional Linux should follow the same rule to
be loaded into normal boot OS image.

### DIFF
--- a/PayloadPkg/OsLoader/LoadImage.c
+++ b/PayloadPkg/OsLoader/LoadImage.c
@@ -804,7 +804,32 @@ LoadBootImages (
       Status = GetBootImageFromRawPartition (OsBootOption, LoadedImage);
     }
 
-    DEBUG ((DEBUG_INFO, "LoadBootImage ImageType-%d Image\n", Index));
+    if (EFI_ERROR (Status) && (Index == LoadImageTypeNormal)) {
+      //
+      // Launch Traditional Linux for debugging purpose only
+      //
+      if ((DebugCodeEnabled () || !FeaturePcdGet (PcdVerifiedBootEnabled)) && (FsHandle != NULL)) {
+        // Free loaded images previously, but keep LoadedImagesInfo structure
+        UnloadBootImages ((EFI_HANDLE)(UINTN)LoadedImagesInfo, TRUE);
+        LoadedImage = LoadedImagesInfo->LoadedImageList[LoadImageTypeNormal];
+        if (LoadedImage == NULL) {
+          LoadedImage = (LOADED_IMAGE *)AllocateZeroPool (sizeof (LOADED_IMAGE));
+          if (LoadedImage == NULL) {
+            return EFI_OUT_OF_RESOURCES;
+          }
+        }
+        LoadedImage->HwPartHandle   = HwPartHandle;
+        LoadedImage->LoadImageType  = LoadImageTypeNormal;
+        LoadedImagesInfo->LoadedImageList[LoadImageTypeNormal] = LoadedImage;
+        Status = GetTraditionalLinux (FsHandle, &LoadedImage->Image.Linux);
+        if (!EFI_ERROR (Status)) {
+          LoadedImage->Flags |= LOADED_IMAGE_LINUX;
+          DEBUG ((DEBUG_INFO, "LoadBootImage TraditionalLinux ImageType-%d Image\n", Index));
+        }
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -804,7 +804,32 @@ LoadBootImages (
...
     }
 
-    DEBUG ((DEBUG_INFO, "LoadBootImage ImageType-%d Image\n", Index));
+    if (EFI_ERROR (Status) && (Index == LoadImageTypeNormal)) {
+      //
+      // Launch Traditional Linux for debugging purpose only
+      //
+      if ((DebugCodeEnabled () || !FeaturePcdGet (PcdVerifiedBootEnabled)) && (FsHandle != NULL)) {
+        // Free loaded images previously, but keep LoadedImagesInfo structure
+        UnloadBootImages ((EFI_HANDLE)(UINTN)LoadedImagesInfo, TRUE);
```
myguitar, line 813: This will clear all pre-loaded images. Is this still valid?

gdong1: I just move the code, didn't change the logic. not sure why it is added before.
when loaded IAS image failure before, do we need free anything left?

myguitar: This whole code is for debugging purpose only. If there is no valid boot image found in for loop, this code will clean-up all allocated memory and load linux files for debugging purpose. Do we need to move the code from the original location to inside for loop?

gdong1: For each boot option, it could load preOS, normal OS, A/B support OS, and extra image.
The traditional Linux should be part of the boot option loading when loading normal OS failure to make the loading flow consistent. And we could also test PreOS + traditional linux case with this change.
If we need UnloadBootImages after each loop, we could separate it from traditional linux loading. 
BTW, if UnloadBootImages is needed, should we also check it regardless debug/release build?

mauricema: The new code logic has altered the current code flow now.  We need to think it carefully.

For example:  the old code will fallback to traditional Linux boot no matter which image loading fails.  The new logic will only do that for the 1st image loading failure.   

I agree with Aiden on this.  It might not be necessary to make this change since Traditional Linux is not considered as a normal boot option, instead, it is a fallback option when the official boot option fails.  So it makes sense to me if we handle it outside of the for loop as an error handler.

gdong1: The new logic is expected for the enhancement.
For a simple boot option (only load normal OS), the new logic is same with old logic.
For other boot options (e.g. PreOS + traditional Linux), old logic will change the boot option to only load traditional Linux since it will free PreOS. The new logic fixed it.
I checked UnloadBootImages (), it is ok to keep it there in case there is some memory allocated during IAS image loading.

mauricema: The new logic is not the same as old one because you added condition "Index == LoadImageTypeNormal".

mauricema: BTW,  as discussed, traditional Linux is not considered as a normal boot option. So PreOS+Traditional Linux doesn't seem to be a valid combination in production.   In production, it should be PreOS + Linux Container.

gdong1: That's because traditional Linux will be loaded into Image[LoadImageTypeNormal] below.
For the simple boot option (only load normal OS), other image load will be skipped. so the new logic is same with old logic ( that is, when normal OS loaded failure, it will load traditional Linux).

gdong1: I agree Traditional Linux is not for production, this change doesn't change that.
It would help developer/user to test PreOS with traditional Linux for debug purpose.

mauricema: Give an example, if the old boot option needs to load two images.  The first image loading is successful, but the 2nd image loading fails. With the old code logic, it will try to run Traditional Linux at the end.  But with the new code it will not.

myguitar: I'd prefer to keep original code. It would be risky to add debug code inside main loop. We may need some unit test result with different boot options and boot image combinations. If this is useful to debug a specific case, "Testing" branch would be good place.

PR #952, opened 2021-01-11
Updated Openssl for Linux

rcperni:
Updated openssl for linux build

Signed-off-by: Perni <ramesh.chandra.perni@intel.com>


gdong1 (approved):

BootloaderCorePkg/Tools/CommonUtility.py
```diff
@@ -201,6 +201,12 @@ def get_openssl_path ():
                     openssl_cfg = "C:\\Openssl\\openssl.cfg"
                     if os.path.exists(openssl_cfg):
                         os.environ['OPENSSL_CONF'] = openssl_cfg
+    else :
+        if 'OPENSSL_PATH' not in os.environ:
```
mauricema, line 205: why do we do this ?  
This might not work in many Linux env.   openssl might be on some other place on the PATH env, but not necessary on /usr/bin.

gdong1: It tried to fix the build issue where openssl is not in the PATH I think.
Maybe check if os.path.join(os.environ.get ('OPENSSL_PATH', ''), 'openssl') works or not, then check other cases?

mauricema: When openssl is installed in Linux, it will be added into PATH.    I don't think we should do this change.  It will break more things.

gdong1: it looks this could be fixed from signing script. so just close this one.

PR #953, opened 2021-01-11
[CFL]Fix payload selection issue

Praveenhp7:
WHL CRB board is configured to boot different payload based on GPIO pin
status, which is not working currenly.
this patch ensures to boot to Osloader payload when PayloadSelGpioData=1
otherwise boot to UEFI payload.

TEST= Boot test on WHL CRB and verified board able to switch between
OSloader & UEFI payload based on DIP switch R4H1B_4 status.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -894,7 +894,7 @@ UpdatePayloadId (
     Status = GpioGetInputValue (PayloadSelGpioPad, &PayloadSelGpioData);
     if (!EFI_ERROR (Status)) {
       if (PayloadSelGpioData == 1) {
-        PayloadId = UEFI_PAYLOAD_ID_SIGNATURE;
+        PayloadId = 0;
       } else {
         if ((GenericCfgData != NULL) && (GenericCfgData->PayloadId == AUTO_PAYLOAD_ID_SIGNATURE)) {
           PayloadId = UEFI_PAYLOAD_ID_SIGNATURE;


Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -894,7 +894,7 @@ UpdatePayloadId (
     Status = GpioGetInputValue (PayloadSelGpioPad, &PayloadSelGpioData);
     if (!EFI_ERROR (Status)) {
       if (PayloadSelGpioData == 1) {
-        PayloadId = UEFI_PAYLOAD_ID_SIGNATURE;
+        PayloadId = 0;
```
mauricema, line 897: Why do you revert the logic ?
It was working,  so wondering what is the real issue.

mauricema: The old logic is level high to boot UEFI payload.  
Now you changed to level high to boot OsLoader.

Praveenhp7: by default, the value of  "PayloadSelGpioData" is 1 when the DIP switch is OFF  and vice versa.  
 with existing code, UEFI payload is configured for both level and board always boot to UEFI payload ( this is due to changes introduced in commit 418e31ad )

myguitar (approved):
Looks good to me

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: OK, I see.  Looks like some bug was introduced previously.  Make sense to me now.

mauricema (approved):
Looks good to me.

PR #954, opened 2021-01-11
[TGL] Add missing binary files

gdong1:
SPD config files and TsnMacAddr bin are missing.
So add them to fix the build failure.


mauricema (commented):
For MAC binary,  it needs to be platform specific.  
I am not sure providing a fixed binary is the correct way.

gdong1:
> For MAC binary, it needs to be platform specific.
> I am not sure providing a fixed binary is the correct way.
Thanks for the comments. Updated BoardConfig.py not to require this binary for the build.
User could stitch this binary if required.

mauricema (approved):
Looks good to me.

PR #958, opened 2021-01-14
use different workspace to prepare binaries

gdong1:
Different projects might use different repo to get binaries.
or different project might be built at same time. so change
to use different workspace for different project to avoid
potential conflict.


mauricema (approved):
Looks good to me.

PR #961, opened 2021-01-14
Enhance PreOS support

gdong1:
SBL support to load PreOS and normal OS in a single boot option.
This patch tries to standardize the PreOS support.
The PreOS could be TrustyOS, PreOsChecker or others.
As long as PreOS flag is set in boot option, SBL will load and
boot PreOS before normal OS. If the preOS has specific requirement,
it could be addressed using PreOS image type.

### DIFF
--- a/PayloadPkg/OsLoader/LoadImage.c
+++ b/PayloadPkg/OsLoader/LoadImage.c
@@ -21,6 +21,59 @@ STATIC CONST CHAR16  *mConfigFileName[3] = {
 };
 
 /**
+  Get boot image from a IFWI container component
+
+  This function will load component from a container inside the IFWI
+  using the container signature and the component name from boot option.
+  After Boot image is loaded into memory, its information will be saved
+  to LoadedImage.
+
+  @param[in]      BootOption      Current boot option
+  @param[in, out] LoadedImage     Loaded Image information.
+
+  @retval  RETURN_SUCCESS     If a boot image was loaded successfully
+  @retval  Others             If a boot image was not loaded.
+**/
+EFI_STATUS
+GetBootImageFromIfwiContainer (
+  IN     OS_BOOT_OPTION      *BootOption,
+  IN OUT LOADED_IMAGE        *LoadedImage
+  )
+{
+  RETURN_STATUS              Status;
+  VOID                       *Buffer;
+  UINT32                     ImageSize;
+  CONTAINER_IMAGE           *Image;
+
+  Image = &BootOption->Image[LoadedImage->LoadImageType].ContainerImage;
+  if ((Image->Indicate != '!') || (Image->BackSlash != '/')) {
+    return EFI_INVALID_PARAMETER;
+  }
+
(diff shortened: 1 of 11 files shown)


PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -20,6 +20,59 @@ STATIC CONST CHAR16  *mConfigFileName[3] = {
...
+    return Status;
+  }
+
+  LoadedImage->Flags |= LOADED_IMAGE_COMPONENT;
+  LoadedImage->ImageData.Addr = Buffer;
+  LoadedImage->ImageData.Size = ImageSize;
+  LoadedImage->ImageData.AllocType = ImageAllocateTypePage;
+
+  if ( *((UINT32 *) Buffer) == CONTAINER_BOOT_SIGNATURE ) {
+    LoadedImage->Flags      |= LOADED_IMAGE_CONTAINER;
```
mauricema, line 67: where is LOADED_IMAGE_CONTAINER flag used ?  Why need a special flag here ?

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -55,149 +108,127 @@ GetBootImageFromRawPartition (
...
+  // Make sure to round the image size to be block aligned in bytes.
+  //
+  ContainerHdr = (CONTAINER_HDR *)BlockData;
+  if (ContainerHdr->Signature == CONTAINER_BOOT_SIGNATURE) {
+    ImageSize = ContainerHdr->DataOffset + ContainerHdr->DataSize;
+  } else if (ContainerHdr->Signature == IAS_MAGIC_PATTERN) {
+    ImageSize = IAS_IMAGE_SIZE ((IAS_HEADER *) BlockData);
+  } else {
+    DEBUG ((DEBUG_INFO, "No valid image header found !\n"));
+    return EFI_LOAD_ERROR;
```
mauricema, line 181: how do you handle BlockData free ?  
many places just return.  Please enhance error handling.

gdong1: it should be handled in UnloadBootImages() if loaded failure.
This patch doesn't change the original logic.

PayloadPkg/OsLoader/LoadImage.c
gdong1: This patch just separate container load from raw image load, doesn't change original logic.
From code, it will be used during parse image phase ParseContainerImage() in OsLoader.c

mauricema (approved):
Looks good to me.

PayloadPkg/OsLoader/LoadImage.c
mauricema: OK

PayloadPkg/OsLoader/LoadImage.c
mauricema: OK.

PR #962, opened 2021-01-15
Update get_openssl_path to return absoulte path

lsubash:
Tools as MEU used for signing and generating key manifests
expects to pass abosulte openssl paths. Updating
get_openssl_path to return paths for linux cases.


gdong1 (approved):

mauricema (approved):
Looks good to me.

PR #967, opened 2021-01-19
[BoardConfig] Support Inherited BoardConfig

myguitar:
This will allow inherited BoardConfig from the existing one
in the same BoardPkg. It will be useful when a new BoardConfig
has very minimum difference from the existing one.

See Platform/QemuBoardPkg/BoardConfigOverride.py as a reference.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -24,6 +24,7 @@ import multiprocessing
 from   ctypes import *
 from   BuildUtility import *
 
+import glob
 
 def rebuild_basetools ():
     exe_list = 'GenFfs  GenFv  GenFw  GenSec  Lz4Compress  LzmaCompress'.split()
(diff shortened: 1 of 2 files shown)


BuildLoader.py
```diff
@@ -1378,26 +1379,31 @@ def main():
...
     board_cfgs   = []
     board_names  = []
+    module_names = []
 
     # Find all boards
     search_dir = os.environ['SBL_SOURCE']
     if 'PLT_SOURCE' in os.environ:
         search_dir = os.path.abspath(os.path.join(search_dir, os.path.pardir))
     board_pkgs = os.listdir (search_dir)
+    board_cfgs = []
```
gdong1, line 1389: This is not required since it is done in line 1380.

myguitar: Thanks. Removed the unnecessary line.

mauricema (approved):
Looks good to me.

PR #968, opened 2021-01-20
Fix 32bit payload calling convention

mauricema:
Current Execute32BitCode() passes in 64bit arguments into stack.
It should use 32bit arguments instead according to 32bit calling
convention. This patch fixed this issue.


myguitar (approved):
Looks good to me

PR #969, opened 2021-01-21
Fix Secureboot status in PSD

Praveenhp7:
CFL, CML, EHL, TGL platforms are using PSD version 0.3.
as per PSD Spec v0.3 secureboot status indication as ber below,

000 – Secure boot is Disabled
001 – UEFI Secure boot is enabled
010 – Boot Guard is Enabled
100 – Bootloader Verified boot is Enabled


myguitar:
It looks PSD library can be a single common library for multiple platforms. Is it available?

mauricema (approved):
Looks good.

mauricema:
BTW, the code seems to be very similar across platforms,  is there any chance to make it more common ?

PR #970, opened 2021-01-21
Fix paging issue when PCI PMEM64 resource exists

mauricema:
In 32 bit SBL, when PCI PMEM64 exists, the OsLoader will hang
during boot in CreateIdentityMappingPageTables().  The function
is inteneded to be used in X64 mode only, and cannot handle
32bit well. So OsLoader should not call it for 32 bit build.

This patch also zeroed the allcated memory to ensure the unused
entries are all 0.


myguitar (approved):
Looks good to me

PR #974, opened 2021-01-23
Fix a bug in PAE 2MB paging

myguitar:
This will fix invalid PDP/PDE generation with PAE 2MB Pages,
and high bit calculation at PayloadEntry additionally.


mauricema (approved):
Looks good to me.

PR #979, opened 2021-01-24
Change PCI resource PCD to patchable type

mauricema:
Current PCI resource PCD bases are defined as fixed type. It
makes it impossible to dynamically adjust the base at runtime.
This patch changed it to be module patchable so that platform
can update when required.


gdong1 (approved):

myguitar (approved):
Looks good to me

PR #987, opened 2021-01-26
[CFL] Add TPM2 ACPI table

Praveenhp7:
similar to CML, CFL also uses both IOB2 & IOB3 ports to pass the info and generate an SMI.
where as common code ACPI implementation uses only IOB2, hence TPM2 ACPI tables are ported
from CML platforms. This patch fixes physical presence query and TPM clear issue on windows.

TEST: boot to windows on CFL-H platform and verified TPM physical presence and TPM clear command is
successfull.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/AcpiTables/AcpiTables.inf
+++ b/Platform/CoffeelakeBoardPkg/AcpiTables/AcpiTables.inf
@@ -42,7 +42,7 @@
   Fwst/Fwst.aslc
   Platform/CommonBoardPkg/AcpiTables/Tpm/Tpm2.aslc
   Platform/CommonBoardPkg/AcpiTables/Fpdt/Fpdt.aslc
-  Platform/CommonBoardPkg/AcpiTables/Tpm/TpmSsdt.asl
+  Ssdt/TpmSsdt.asl
   Ssdt/Thermal.asl
   CpuSsdt/Cpu0Cst.asl
   CpuSsdt/Cpu0Ist.asl
--- a/Platform/CoffeelakeBoardPkg/AcpiTables/Ssdt/TpmSsdt.asl
+++ b/Platform/CoffeelakeBoardPkg/AcpiTables/Ssdt/TpmSsdt.asl
@@ -1,0 +1,369 @@
+/** @file
+  The TPM2 definition block in ACPI table for TCG2 physical presence
+  and MemoryClear.
+
+Copyright (c) 2017 - 2018, Intel Corporation. All rights reserved.<BR>
+SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+DefinitionBlock (
+  "Tpm.aml",
+  "SSDT",
+  2,
+  "INTEL ",
+  "Tpm2Tabl",
+  0x1000
+  )
+{
+  Scope (\_SB)
+  {
+    External(SMIE)
+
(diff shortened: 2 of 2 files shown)


Platform/CoffeelakeBoardPkg/AcpiTables/Dsdt/Platform.asl
```diff
@@ -190,15 +190,6 @@ Field (SPRT, ByteAcc, Lock, Preserve) {
...
 }
 
-//
-// Operational region for ACPI Control (SMI_EN) access
-//
-OperationRegion (SMIN, SystemIO, (ACPI_BASE_ADDRESS + R_ACPI_IO_SMI_EN), 0x4)
-Field (SMIN, AnyAcc, NoLock, Preserve)
-{
-  SMIE, 32
-}
```
lsubash, line 200: You could have this code in platform.asl and remove it in TpmSsdt.asl.

lsubash (approved):
Looks good.

Platform/CoffeelakeBoardPkg/AcpiTables/AcpiTables.inf
```diff
@@ -42,7 +42,7 @@
   Fwst/Fwst.aslc
   Platform/CommonBoardPkg/AcpiTables/Tpm/Tpm2.aslc
   Platform/CommonBoardPkg/AcpiTables/Fpdt/Fpdt.aslc
-  Platform/CommonBoardPkg/AcpiTables/Tpm/TpmSsdt.asl
+  Ssdt/TpmSsdt.asl
```
gdong1, line 45: should we update common TpmSsdt.asl to add 0xB3 support so that each platform doesn't copy this file?

Praveenhp7: UEFI payload TCG2SMM Implementation for  CFL/CML is different compare to other platforms. am not sure what is the impact on other platforms if we move it to common code.

Platform/CoffeelakeBoardPkg/AcpiTables/Dsdt/Platform.asl
Praveenhp7: done

Platform/CoffeelakeBoardPkg/AcpiTables/AcpiTables.inf
gdong1: B3 port is used for data. it should have no impact if it is not used previously by other platforms.
Subash could comment on this one since he did this for all the platforms.

lsubash: Regarding IOB3 to common TpmSsdta.asl, we can do it. It would be don't care bit for EHL/TGL. We would need to write same value to B2 & B3 ports. 

It's better to adapt SMM dispatcher to TGC2SMM for CML/CFL platforms and use common asl file.
There are some issues i have seen earlier in CML for SMI generations after Windows boot and continued older TGC2SMM implementation.  Can we add a as seperate asl for CFL? Once SMI is fixed with SMM dispatcher we could move to common file for CML & CFL.

gdong1: sure.

gdong1 (approved):

PR #990, opened 2021-01-27
[TGL] Enlarge Payload region Size

gdong1:
Since image form Linux build is a little bigger,
so enlarge the payload region size to fix linux build.


jgutbub (approved):
Looks good to me.

PR #997, opened 2021-02-02
Add splash support post PCI enumeration

mauricema:
In some cases FSP does not support GFX and does not produce
GFX hob. But platform will be able to initialize its GFX after
PCI enumeration. This patch allows splash to be displayed post
PCI if the splash has not been displayed yet.


myguitar (approved):
Looks good to me

gdong1 (approved):

PR #998, opened 2021-02-02
[PCI] Add an option to allocate PCI PMEM resource first

myguitar:
This introduces an additional PCI Enumeration option.
- self._PCI_ENUM_FLAG_ALLOC_PMEM_FIRST

By deafult, the option will allocate PCI resource by ascending order
(MEM32->PMEM32->MEM64->PMEM64). If it's set to 1, by reversed order.

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dec
+++ b/BootloaderCorePkg/BootloaderCorePkg.dec
@@ -100,8 +100,13 @@
   # } PCI_RES_DOWNGRADE;
   #
   # typedef struct {
+  #   UINT16            AllocPmemFirst  : 1;  // default:0 (Mem32->Pmem32->Mem64->Pmem64), 1 (Pmem64->Mem64->Pmem32->Mem32)
+  #   UINT16            Reserved        : 15; // 0 initialized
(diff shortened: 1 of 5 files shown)


mauricema (commented):
Do you think we should use policy to allow an option to control it ?

myguitar:
> Do you think we should use policy to allow an option to control it ?

I don't think we need an option for this unless MEM alignment is larger than PMEM alignment, which is not common case.
To make sure which one goes first, we can sort BAR type by alignment size and program its resources by sorted order. But I think it would be same result as this reverted loop.

mauricema:
I am OK with that approach. I am just a little bit concerned about the compatibility issue, especially for 32 bit resource.

gdong1 (approved):

mauricema (approved):

myguitar:
> I am OK with that approach. I am just a little bit concerned about the compatibility issue, especially for 32 bit resource.

Thanks Maurice. Can you let me know your concern about 32 bit resource?

mauricema (approved):
Looks good to me.

PR #1000, opened 2021-02-03
commonSocPkg: SpiFlashLib: Permissions bit read per read/write

AdithyaBaglody:
The RegionPermission doesn't usually represent the current state of
the Region. There is a need to re-read the permission bit for each
read/write. There by making the variable accurately represents the
HW status.

Signed-off-by: Adithya Baglody <adithya.nagaraj.baglody@intel.com>


mauricema (approved):
Looks good to me.

PR #1001, opened 2021-02-04
[CFL] Fix CSME firmware update failure

Praveenhp7:
during CSME firmware update process, CSME update library throw error
"Could not access PCI device".this patch fixes this issue by adding
back "PciReadBuffer".

TEST=Verified CSME FWU on CFL-H & WHL platforms.


mauricema (approved):
Looks good to me.

PR #1002, opened 2021-02-05
Avoid build duplicated GFX HOB

mauricema:
In some cases Bootloader will build GFX HOB. This patch updated
SBL to check the bootloader GFX HOB first before building a new
GFX HOB from FSP. This is to avoid duplicated GFX HOB to be
generated in bootloader HOB.


gdong1 (approved):

myguitar (approved):
Looks good to me

PR #1003, opened 2021-02-05
Delay MP init done for OsLoader payload

mauricema:
There is request to utilize MP in OsLoader. To support it, it is
desired to delay MP init done signal to the end of the OsLoader.
This patch moved the MP init done signal into board ReadyToBoot
notification so that MP is still alive in OsLoader phase.


gdong1 (approved):

myguitar (approved):
Looks good to me

PR #1004, opened 2021-02-06
Build MP CPU TASK info hob

gdong1:
With this hob, user could run a task from AP in Osloader.


BootloaderCommonPkg/Include/Guid/MpCpuTaskInfoHob.h
```diff
@@ -0,0 +1,50 @@
...
+  EnumCpuReady,
+  EnumCpuStart,
+  EnumCpuEnd,
+} CPU_STATE;
+
+typedef UINT32 (*CPU_TASK_PROC)          (UINT32 Arg);
+
+typedef struct {
+  // Refer CPU_STATE
+  UINT32           State;
```
mauricema, line 29: considering we are supporting 64bit.  
Should we to have all argument to update to 64bit instead ?  
It is HOB now, need to avoid changing later   // CPU task function
typedef struct {
  UINT8             State;
  UINT8             Request;
  UINT8             Reserved[6];
  UINT64           CProcedure;
  UINT64           Argument;
  UINT64           Result;
} CPU_TASK;on.

BootloaderCommonPkg/Include/Guid/MpCpuTaskInfoHob.h
```diff
@@ -0,0 +1,50 @@
...
+
+  // CPU task function
+  CPU_TASK_PROC    CProcedure;
```
mauricema, line 32: Better to use a  UINT64 instead of pointer since the size will be different between 32bit and 64bit environment.

BootloaderCommonPkg/Include/Guid/MpCpuTaskInfoHob.h
gdong1: yes. we could update to use UINT64. what the Request field is used for?

BootloaderCommonPkg/Include/Guid/MpCpuTaskInfoHob.h
mauricema: This is the flag to request to start running the CProcedure.   The old code use CProcedure itself as a flag to trigger.  It is OK to do that as well if we can ensure the write is atomic.   Using a separate byte filed might be safer.

gdong1: Based on current impl, it should be ok as long as we always set State to EnumCpuStart at the end.

mauricema (approved):
Looks good to me.

PR #1013, opened 2021-02-10
Fix the MP hang issue

gdong1:
The ApDataPtr->CProcedure was wrongly updated in previous patch.
This patch fixed it and CPU task name from CProcedure to TaskFunc
to avoid confusion.


mauricema (approved):
Looks good to me.

myguitar (approved):
Thanks Guo for the quick fix!

PR #1014, opened 2021-02-10
Add CPU X2APIC support

mauricema:
This patch added X2APIC support. It is to enable the case when
APIC ID is greater than 255. This patch only handle core wakeup
portion. Platform still needs to handle ACPI related changes for
X2APIC.

X2APIC lib is backward compatible with XAPIC lib. So there is no
need to use XAPIC lib anymore.


myguitar (approved):
Looks good to me.

APIC is verified on a platform which supports X2APIC, but X2APIC is not ready for testing on the platform yet. Will follow up with this code base later.

mauricema:
> Looks good to me.
> 
> APIC is verified on a platform which supports X2APIC, but X2APIC is not ready for testing on the platform yet. Will follow up with this code base later.

Thank you, Aiden!   I have tested X2APIC on APL platform and it worked fine in SBL phase.

PR #1031, opened 2021-02-18
Move X2APIC enabling to common function

mauricema:
This patch removed duplicated X2APIC enabling code. Instead, it
enables X2APIC in a common function. By doing so, the very first
waking up will be done in APIC mode. Afterwards, it will be using
X2APIC mode if enabled by PCD.
This patch also fixed an X2APIC ACPI MADT issue.


myguitar (approved):
Looks good to me

PR #1034, opened 2021-02-23
Fix MP waking up issue on some platforms

mauricema:
It was reported that some platform had MP waking up issue after
switching to using X2APIC library. By comparing the library, found
X2APIC removed 2nd IPI sending in the flow. This 2nd IPI is
required per IA specification. The patch added it back.

Tests have been done and confirmed it fixed the issues seen on
thos platforms.


gdong1 (approved):

mauricema:
The failure was caused by ACPICA tool URL.  The old tool is not available any more.  Will fix it in separate patch.

ywho79 (approved):

PR #1036, opened 2021-02-24
[CFL] Set Tseg size to 16MB.

Praveenhp7:
This Patch did the following

-  TsegSize config options is defined in CfgData_Memory.yaml.
-  with 64GB RAM, slow boot issue was reported on ubuntu_20.0.
   Root cuase of the issue is due out of MTTR's and unable to cover
   portion of higher memory ranges. this patch fixes this issue.


mauricema (approved):
Looks good to me.

ywho79 (approved):

PR #1037, opened 2021-02-25
Update CsmePciReadBuffer callback prototype params

lsubash:
CsmePciReadBuffer function prototype is aligned as
per the CSME fwupdate lib. It fixes issue with
CSME capsule fw update.


mauricema (approved):
Looks good to me.

PR #1038, opened 2021-02-25
Add FSP swapping script

mauricema:
This patch added FSP swapping script to support replace a FSP
binary inside SBL IFWI or BIOS image.

To use it, use
  python BootloaderCorePkg\Tools\FspSwap.py ifwi.bin fsp.bin
The default output directory is "Out".


myguitar (approved):
Looks good to me

PR #1040, opened 2021-02-26
[APL] Fix Build error when SOURCE_DEBUG is enabled

Praveenhp7:
This patch fixes the multiple build issues which are observed
when ENABLE_SOURCE_DEBUG config is set to 1.


myguitar (approved):
Looks good to me

mauricema (approved):
Looks good.

PR #1041, opened 2021-02-27
Fix bug while updating binary not aligned to 4k

raghavag:
This patch fixed an error that occur during updating a
binary that is not aligned to 4k. Current code will always
update next 4K block. This patch adjusted the block length
to remaining size other than 4K.


gdong1 (approved):

mauricema (approved):
Looks good.

PR #1042, opened 2021-03-01
Program BAR0/1 for PPB

stalamudupula:
Current PCI Enum Lib scopes for only Apperture resources
for a PPB. But some OSes (like ESXi) expect BAR0 & BAR1
(Offset 0x10/0x14) to be allocated resources accordingly.
Otherwise, PPB enumeration doesnt happen correctly and
devices behind PPB are not registered at all.

This patch adds the functionality to assign valid resources
to BAR0(0x10) and BAR1(0x14) for a PPB also.


mauricema (commented):
Looks good to me.

mauricema (approved):

PR #1048, opened 2021-03-17
Build option to disable os verified boot disable

lsubash:
Add -n to sbl build to disable verified boot while
loading os through osloader. Container header verification
would be disabled when this option is enabled. This option
is to be used for developer use only.

### DIFF
--- a/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
+++ b/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
@@ -391,7 +391,12 @@ AutheticateContainerInternal (
       AuthData = (UINT8 *)ContainerHdr + ALIGN_UP(ContainerHdrSize, AUTH_DATA_ALIGN);
       if ((AuthType == AUTH_TYPE_NONE) && FeaturePcdGet (PcdVerifiedBootEnabled)) {
         Status = EFI_SECURITY_VIOLATION;
-      } else {
+      } else if ((ContainerHeader->Signature == CONTAINER_BOOT_SIGNATURE)
+                                                    && FeaturePcdGet (PcdVerifiedBootEnabled)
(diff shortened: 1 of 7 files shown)


BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -391,7 +391,12 @@ AutheticateContainerInternal (
       AuthData = (UINT8 *)ContainerHdr + ALIGN_UP(ContainerHdrSize, AUTH_DATA_ALIGN);
       if ((AuthType == AUTH_TYPE_NONE) && FeaturePcdGet (PcdVerifiedBootEnabled)) {
         Status = EFI_SECURITY_VIOLATION;
-      } else {
+      } else if ((ContainerHeader->Signature == CONTAINER_BOOT_SIGNATURE)
+                                                    && FeaturePcdGet (PcdVerifiedBootEnabled)
+                                                    && FeaturePcdGet (PcdOsVerifiedBootDisabled)) {
+          DEBUG((DEBUG_INFO, "NOTE: OS VERIFIED BOOT DISABLED!!\n"));
+          Status = EFI_SUCCESS;
+        } else {
```
lsubash, line 399: Skip RSA verification of Container header when signature is BOOT & PcdOsVerifiedBootDisabled is set. For MONO, Verification for components inside the OS container would happen. These components hashes are stored as part of container header.

myguitar:
Looks like another discussion is on-going. Hold on for a while?

lsubash:
> Looks like another discussion is on-going. Hold on for a while?

Yes, we can hold this patch. If fixed key option is supported by yocto, this patch might not be required.

mauricema:
let us close this for now.   Will reopen it when the decision is made.

PR #1049, opened 2021-03-17
[EHL] Operational region for ACPI Control SMI_EN access

lsubash:
SMIE is checked in Tpm asl before SMI is triggered.
Removal of this definition causes physical presence
usecases failure.

TEST=Verified TPM clear usecases with windows boots


mauricema (approved):
Looks good to me.

PR #1053, opened 2021-03-19
Common boot guard signing as separate and standalone script

lsubash:
This patch adapts common btg utility and restructures
stitch script and separates security functionality.

BtgSign.py - Common Boot guard signing and oem key manifest generation.
			 It also supports signing in standalone mode. Stanalone mode can
			 be used sign bootloader bin and stitch using FIT tool.

security_stitch_help.py - Interface functionality for Btg manifest and
			 update xml with security updates. This is as per the client
			 platform FIT xml format.

Currently TGL, EHL uses these scripts.


mauricema (approved):
Looks good to me.

PR #1054, opened 2021-03-19
[EHL] Fix UART init issue and set UART PCI mode as default

LeanSheng:
This CL fixes the long time issue where EHL SBL failed to init
properly and solely relies on FSP to handle the UART unit, and
hence the limitation of Hidden Mode UART only, as we observed
the UART output gone missing after PCI enumeration if we set
respective UART port into PCI mode in FSP. By hiding the UART,
OS will not be able to see the UART device as PCI device and
lose control to the UART device.

Due to hardware design, different uart could use different
LPSS_IO_MEM_PCP register offset for UART clock setup.

This CL includes dynamic configuration for clock setup by
reading the size of UART control register. Since this is pretty
generic for most of platforms, will plan to move more UART codes
to common codes in the future.

Second fix changes the default UART mode for both FSP-T and FSP-S
to skip uart init, and let SBL solely handles it and setup as a
PCI device.

Signed-off-by: LeanSheng <lean.sheng.tan@intel.com>


mauricema (approved):
Looks good to me!

ywho79 (approved):

PR #1057, opened 2021-03-22
[TGLU] Oem Key revocation feature support

lsubash:
TGL supports multiple OEM keys and their revocation
by CSE. This patch supports HECI APIs for OemkeyRevoke
and to get key status. This is port from TGLH platform
implemetation.

### DIFF
--- a/Silicon/TigerlakePchPkg/Include/Library/HeciInitLib.h
+++ b/Silicon/TigerlakePchPkg/Include/Library/HeciInitLib.h
@@ -377,4 +377,18 @@ HeciArbSvnCommitMsg (
   VOID
   );
 
+/**
+  This command is used to indicate to the FW to revoke a OEM key.
+  It should be executed when MBP data indicates that a key is pending for
+  revocation.
+
+  @retval EFI_SUCCESS             Command succeeded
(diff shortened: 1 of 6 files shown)


mauricema (commented):
Is the key revocation done automatically?   Does user have any control over it ?
Will this cause side effect if something is called by mistake ?

lsubash:
> Is the key revocation done automatically? Does user have any control over it ?
> Will this cause side effect if something is called by mistake ?

In Slimboot, OEM Key revocation would be initiated through Capsule Fwupdate CMDI mode. Automatic revocation based on MbpData has not yet enabled. For initiating revocation,
- By default, IFWI would contain BIOS (Btg KM & BPM signed with Key1). 
- Create an Oem Key manifest where user specifies key hash to be revoked.
- User need to update BIOS & CSME signed with second key though capsule FwUpdate process.
- Once System boots with updated BIOS & CSME (signed with Key2), CSME detects revocation is available. User can initiate Capsule Fwupdate CMDI mode to revoke key.

mauricema (approved):
Thanks.  Looks good to me.

gdong1 (approved):

PR #1059, opened 2021-03-23
[EHL] Fix SPI clock source gating s0ix issue

LeanSheng:
SPI clock source was on and gating s0ix entry, due to linux OS could
trigger SPI Write Protection Disable bit and hence set the SPI
Synchronous SMI Status bit. This CL fixes it by clearing the SPI
Synchronous SMI Status bit prior to S0ix entry.

Traditionally, it is expected for bootloader to clear this bit in
SMI handler upon S0ix entry, due to SBL being free of SMI, clearing
this bit from ACPI table has the same effect.

Signed-off-by: LeanSheng <lean.sheng.tan@intel.com>


ywho79 (approved):
Looks good

gdong1 (approved):

myguitar (approved):
Looks good to me

PR #1062, opened 2021-03-23
Fix Shell MTRR print issue

mauricema:
Current MTRR lib assumes the MTRR number is always 10. Instead,
this patch follows the IA manual to get the actual MTRR number
through MTRR capability register.


myguitar (approved):
Looks good to me

gdong1 (approved):

PR #1068, opened 2021-03-25
Add Visual Studio toolchain version select option

mauricema:
Current SBL always uses the latest VS toolchain detected to build
SBL. However, sometimes a different verion might be preferred.
This patch added a build option "-t" to specify the perferred VS
toolchain version. It is useful when mulitple VS toolchain versions
are installed.

For example, add '-t vs2015' to use VS2015 as the toolchain for
SBL build.


myguitar (approved):
Looks good to me

PR #1070, opened 2021-03-26
[APL] Firmware update flow fix

mauricema:
This patch enforces to use the SBL version for current boot
to check against the new version in capsule. Only when the
new SVN version in capsule is greater than or equal to current
SBL SVN version, the update is allowed.


raghavag (approved):
Looks good to me

PR #1071, opened 2021-03-26
Update MP lib to fix CPU task

gdong1:
Need initialize global variable mSysCpuTask.CpuCount to actual
CPU count.


mauricema (approved):
Looks good.

PR #1072, opened 2021-03-27
[TGL] Add TCC V2 support

gdong1:
1) Add PCD PcdTccEnabled so that TCC could build out when disabled
2) Add HOB gTccRtctHobGuid produced by FSP if FSP support TCC V2
3) Add a common TCC config data in common platform package
4) Add RTCM support from OsLoader

### DIFF
--- a/PayloadPkg/OsLoader/Rtcm.c
+++ b/PayloadPkg/OsLoader/Rtcm.c
@@ -1,0 +1,83 @@
+/** @file
+
+  Copyright (c) 2019 - 2021, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Library/DebugLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/BootloaderCommonLib.h>
+#include "OsLoader.h"
+
+#define SIGNATURE_IPFW    SIGNATURE_32 ('I', 'P', 'F', 'W')
+#define SIGNATURE_RTCM    SIGNATURE_32 ('T', 'C', 'C', 'R')
+
+typedef VOID (*RTCM_ENTRY)(VOID *Params);
+
+/**
+  Load and Run RTCM.
+
+  This function will load and run RTCM.
+
+  @retval  EFI_NOT_FOUND    RTCM binary file was not found.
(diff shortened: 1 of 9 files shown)


PayloadPkg/OsLoader/Rtcm.c
```diff
@@ -0,0 +1,83 @@
...
+  Load and Run RTCM.
+
+  This function will load and run RTCM.
+
+  @retval  EFI_NOT_FOUND    RTCM binary file was not found.
+  @retval  EFI_LOAD_ERROR   RTCM binary file loading failed
+  @retval  EFI_SUCCESS      RTCM load and run successfully.
+**/
+EFI_STATUS
+LoadAndRunRtcm(
```
mauricema, line 29: I think this function should be renamed to a generic function LoadPeImage(), and it can be put into BootloaderCommonPkg.    LoadComponent() and call the actual entry point can be left out in OsLoader.
Also do you think we can use the similar boot option mechanism to handle it ?  The only difference is this boot option will return ?

PayloadPkg/OsLoader/Rtcm.c
```diff
@@ -0,0 +1,83 @@
...
+  VOID         *RtcmRunTimeBuffer;
+  RTCM_ENTRY   EntryPoint;
+  VOID         *RtcmEntry;
+
+  Buffer = NULL;
+  Length = 0;
+  RtcmRunTimeBuffer = 0;
+
+  Status = LoadComponent(SIGNATURE_IPFW, SIGNATURE_RTCM, &Buffer, &Length);
+  if (EFI_ERROR(Status) || (Buffer == NULL) || (Length == 0))
```
mauricema, line 45: Check status should be good enough.

PayloadPkg/OsLoader/Rtcm.c
gdong1: yes, we could use boot option mechanism to handle it. we need load this kind of image using extra image slots, and we need extend boot option flags and boot image info to indicates this kind of info. let me explore it.

PayloadPkg/OsLoader/Rtcm.c
gdong1: agree.

gdong1:
close this one since it is replaced by https://github.com/slimbootloader/slimbootloader/pull/1080

PR #1073, opened 2021-03-27
[TGL] Add TCC V2 support

gdong1:
Update the TCC subregion layout.
Use the common TCC config data
Use the new BIOS settings from TCC subregion
Rename TCC variable to follow TCC V2 naming
Pending the FSP release for TCC V2 test.


Platform/TigerlakeBoardPkg/AcpiTables/Rtct/Rtct.act
```diff
@@ -0,0 +1,58 @@
+/** @file
+
```
mauricema, line 2: If it is empty table, consider constructing the table at runtime. Don't need the RTCT file in flash at all.   Refer to mAcpiTblTmpl() in AcpiInitLib.  I think we can add one API in AcpiInitLib to allow platform to append new ACPI table template into mAcpiTblTmpl so that the ACPI core code can help create the table at runtime.

gdong1: yes, it would be great if the AcpiInitLib could provide a generic way to let platform to append a new table. Then we don't need have RTCT ASL code at all.

mauricema: I have created a PR https://github.com/slimbootloader/slimbootloader/pull/1075  to allow this.   Please take a look.

gdong1:
close this one since it is replaced by https://github.com/slimbootloader/slimbootloader/pull/1081

PR #1076, opened 2021-03-29
Set current boot option in OsLoader

mauricema:
The boot logic has been adjusted in OsLoader. As a result, the
mCurrentBoot does not point to the current boot option any more.
It is always 0. This patch adjusted the value for it to indicate
the current boot option.


jgutbub (approved):
Looks good to me.

gdong1 (approved):

PR #1077, opened 2021-03-29
Add extra images boot support

gdong1:
Before normal OS boot, some images need run firstly.
These images will return to SBL after complete.
This patch uses OS boot option extra images to
support this use case.

It uses a string in the boot option to indicate this
container image need run before OS.
Container component name format: "!XXXX\YYYY:ZZZZ"
"XXXX" and "YYYY" are container and component signature.
For "ZZZZ", Value "POSN" means this component will be
run before OS. Value "POSR" means this component need
runtime memory to run it.


mauricema (approved):
Looks good to me.

PR #1079, opened 2021-03-29
Disable stitching customizations that can cause stitching errors

pqueirosint:
- Remove OEM key manifest file and  key hash static
  configurations.
  These tags are already handled by oem km helper
  functions.

- Disable RomTraceFiltering which causes build errors, and is not
  required for stitching.


lsubash (approved):

LeanSheng (approved):
Good catch. Thanks for uploading this CL! :)

mauricema (approved):
Looks good.

PR #1088, opened 2021-03-31
[UP2] GPIO fix for PCIe slot

mauricema:
This patch configured UP2 GPIO213 default to be output with level
high. It will select the PCIe interface as default. The original
default was mSATA.


myguitar (approved):
Looks good to me

PR #1090, opened 2021-04-01
Fix the boot option

gdong1:
When PreOS is configured from OS boot option data, the common function
FillBootOptionListFromCfgData () need update it to OS correct boot option
image LoadImageTypePreOs. Similarly when extra image is specified, need
update to extra image.
Update ImageType value and fix an image load issue for RTCM.

### DIFF
--- a/Platform/CommonBoardPkg/CfgData/Template_BootOption.yaml
+++ b/Platform/CommonBoardPkg/CfgData/Template_BootOption.yaml
@@ -2,7 +2,7 @@
 #
 #  Slim Bootloader Platform CFGDATA Template File.
 #
-#  Copyright (c) 2020, Intel Corporation. All rights reserved.<BR>
+#  Copyright (c) 2020 - 2021, Intel Corporation. All rights reserved.<BR>
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 ##
@@ -18,10 +18,12 @@ BOOT_OPTION_TMPL: >
     - ImageType_$(1) :
         name         : Image Type
         type         : Combo
-        option       : 0:Default, 1:Android, 2:ClearLinux, 3:Acrn, 4:Fastboot, 0xFE:Addendum, 0xFF:Not used
+        option       : 0:Default, 1:Android, 2:ClearLinux, 3:Acrn, 4:Fastboot, 0x1E:Extra image, 0x9E:PreOS image, 0xFF:Not used
         help         : >
                        Specify boot image type.
-                       Specially 'Addendum' indicates this option is not a standalone boot option. Instead, it provides additional information for the previous boot option.
+                       Extra and PreOS image type indicates this option is not a standalone boot option. Instead, it provides additional information for the previous boot option.
+                       Extra image means load and run extra image before normal OS image. The extra image will return to SBL and SBL will continue run normal OS.
+                       PreOs image means load and run PreOS image before normal OS image. SBL prepares normal OS info and passes to PreOS so the PreOs will not return to SBL.
                        'Not used' indicates this option will be ignored.
         length       : 0x01
         value        : $(2)
--- a/Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
+++ b/Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2019 - 2021, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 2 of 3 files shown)


Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -98,6 +99,17 @@ FillBootOptionListFromCfgData (
       CopyMem (BootOption, BootOptionCfgData, OFFSET_OF (OS_BOOT_OPTION, Image[0]));
     }
 
+    Image = &BootOptionCfgData->Image[0].ContainerImage;
+    if ((UpdateFlag == 2) && (Image->Indicate == '!') && (Image->BackSlash == '/')) {
+      // Currently addendum boot option only support RTCM and preOs
+      if ((Image->ContainerSig == SIGNATURE_32 ('I', 'P', 'F', 'W')) && (Image->ComponentName == SIGNATURE_32 ('T', 'C', 'C', 'R'))) {
```
mauricema, line 105: For this function, I think it is better to let the config item to indicate which image type the addendum is.  
So here we don't have to compare the signature.  Also it is not good to assume PreOs image for any addendum entry.

gdong1: Sure. let me update the patch

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -78,48 +79,50 @@ FillBootOptionListFromCfgData (
...
 
-    UpdateFlag = 0;
     if (BootOptionCfgData->ImageType == IMAGE_TYPE_ADDENDUM) {
-      // This entry is an addendum for previous boot option entry
-      if (ImageIdx < ARRAY_SIZE(BootOption->Image)) {
-        UpdateFlag = 2;
-        BootOption = &OsBootOptionList->OsBootOption[PrevBootOptionIndex];
+      // This entry is an extra image addendum for previous boot option entry
+      BootOption = &OsBootOptionList->OsBootOption[PrevBootOptionIndex];
+      if (BootOptionCfgData->PreOsImageType > 0x3) {
```
mauricema, line 89: Use Enum type instead of 3

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -78,48 +79,50 @@ FillBootOptionListFromCfgData (
...
+        continue;
       }
+      ImageIdx = LoadImageTypeExtra0 + BootOptionCfgData->PreOsImageType;
+    } else if (BootOptionCfgData->ImageType == IMAGE_TYPE_NOT_USED) {
```
mauricema, line 93: Why NOT_USED is for PreOS ?   It is a little bit confusing to me.

Platform/CommonBoardPkg/CfgData/Template_BootOption.yaml
```diff
@@ -18,10 +18,12 @@ BOOT_OPTION_TMPL: >
     - ImageType_$(1) :
         name         : Image Type
         type         : Combo
-        option       : 0:Default, 1:Android, 2:ClearLinux, 3:Acrn, 4:Fastboot, 0xFE:Addendum, 0xFF:Not used
+        option       : 0:Default, 1:Android, 2:ClearLinux, 3:Acrn, 4:Fastboot, 0x1E:load and run extra image, 0x1F:PreOS default, 0xFF:Not used
         help         : >
                        Specify boot image type.
-                       Specially 'Addendum' indicates this option is not a standalone boot option. Instead, it provides additional information for the previous boot option.
+                       Image type 0x1E or 0x1F indicates this option is not a standalone boot option. Instead, it provides additional information for the previous boot option.
+                       0x1E means it need load and run extra image before normal OS image. The extra image will return to SBL and SBL will continue run normal OS.
```
mauricema, line 25: These values is not visisual in ConfigEditor. We should state the actual string in help.

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
gdong1: 0x3 is max extra image the boot option supported. so have a MACRO for it.

Platform/CommonBoardPkg/CfgData/Template_BootOption.yaml
gdong1: It is visual in Config editor. that's why we listed the options for preOS and extra image.

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
gdong1: Only when ImageType is not used, the PreOsImageType is used as an indication for PreOsImageType.
OK. let me change to use IMAGE_TYPE_ADDENDUM.

mauricema (approved):
looks good to me.

PR #1098, opened 2021-04-05
Print Payload ID in human readable format

mauricema:
This patch printed Playload ID in human readable string format
instead of HEX string.

### DIFF
--- a/BootloaderCorePkg/Stage2/Stage2.c
+++ b/BootloaderCorePkg/Stage2/Stage2.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2016 - 2019, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2016 - 2021, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -68,11 +68,11 @@ PreparePayload (
   UINT32                         Dst;
   UINT32                         DstLen;
   VOID                          *DstAdr;
-  BOOLEAN                        IsNormalPld;
   UINT32                         PayloadId;
   UINT32                         ContainerSig;
   UINT32                         ComponentName;
   UINT8                          BootMode;
+  UINT64                         SignatureBuf;
 
   BootMode = GetBootMode();
   //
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -98,6 +96,15 @@ PreparePayload (
...
   }
 
+  if (ContainerSig < COMP_TYPE_INVALID) {
+    SignatureBuf = SIGNATURE_32 (' ', ' ', ' ', ' ');
+  } else {
+    SignatureBuf = ContainerSig;
+  }
+  DEBUG ((DEBUG_INFO, "Loading Payload ID %4a", (CHAR8 *)&SignatureBuf));
+  SignatureBuf = ComponentName;
+  DEBUG ((DEBUG_INFO, ":%4a\n", (CHAR8 *)&SignatureBuf));
```
gdong1, line 106: Thanks adding this print. 
Better adding this kind of pint to containLib so that it could cover all the container components including IPFW components?

mauricema: ContainLib is already using this print format to print string instead of HEX.
For example:
DEBUG ((DEBUG_INFO, "Registering container %4a\n", (CHAR8 *)&SignatureBuffer));

gdong1: I don't remember it prints the container name and component name when loading TCC components.
Better also add this info to LoadComponent()?

sindhuragrandhi: It does print like this: From ADL logs captured at different places:
Registering container KEYH
Registering container IPFW
Registering container EPLD

mauricema: thx!   Anyway, let me add more print to print both container name and component name.

mauricema:
Updated a new commit to print in ContainerLib.

gdong1 (approved):

BootloaderCorePkg/Stage2/Stage2.c
gdong1: Thanks for the update.

myguitar (approved):

PR #1099, opened 2021-04-05
[PCI] Disable PCI devices bus master by default

myguitar:
This will disable all PCI bus master by default, and enable it only if
- the original bus master was enabled before PCI enumeration
- Or the device is PCI Bridge

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
@@ -1224,9 +1224,13 @@ EnablePciDevice (
   CurrentLink = Parent->ChildList.ForwardLink;
   while (CurrentLink != NULL && CurrentLink != &Parent->ChildList) {
     PciIoDevice = PCI_IO_DEVICE_FROM_LINK (CurrentLink);
-    PciExpressOr16 (PciIoDevice->Address + PCI_COMMAND_OFFSET, \
-                    EFI_PCI_COMMAND_IO_SPACE | EFI_PCI_COMMAND_MEMORY_SPACE | EFI_PCI_COMMAND_BUS_MASTER);
+    PciExpressOr16 (PciIoDevice->Address + PCI_COMMAND_OFFSET,
+                    EFI_PCI_COMMAND_IO_SPACE | EFI_PCI_COMMAND_MEMORY_SPACE);
     if (PciIoDevice->ChildList.ForwardLink != &PciIoDevice->ChildList) {
+      if (IS_PCI_BRIDGE (&(PciIoDevice->Pci))) {
+        PciExpressOr16 (PciIoDevice->Address + PCI_COMMAND_OFFSET,
+                        EFI_PCI_COMMAND_BUS_MASTER);
+      }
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -1224,8 +1229,10 @@ EnablePciDevice (
   CurrentLink = Parent->ChildList.ForwardLink;
   while (CurrentLink != NULL && CurrentLink != &Parent->ChildList) {
     PciIoDevice = PCI_IO_DEVICE_FROM_LINK (CurrentLink);
-    PciExpressOr16 (PciIoDevice->Address + PCI_COMMAND_OFFSET, \
-                    EFI_PCI_COMMAND_IO_SPACE | EFI_PCI_COMMAND_MEMORY_SPACE | EFI_PCI_COMMAND_BUS_MASTER);
+    if (IS_PCI_BRIDGE (&(PciIoDevice->Pci))) {
```
mauricema, line 1232: should we check if it has a child ?
If a bridge has no child device,  maybe we should not enable the command register ?

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -505,7 +505,12 @@ GatherDeviceInfo (
     }
   }
 
-  PciExpressWrite16 (PCI_EXPRESS_LIB_ADDRESS (Bus, Device, Func, PCI_COMMAND_OFFSET), Value);
+  //
+  // Restore PCI command register only if the original bus master was enabled.
+  //
+  if ((Value & EFI_PCI_COMMAND_BUS_MASTER) != 0) {
```
mauricema, line 511: Will this disable UART debug msg ?  I guess it might be good to keep this ?

myguitar: This will not disable UART debug message since UART PCI device already enabled its bus master before PCI enumeration.

mauricema: UART does not have to enable bus master.  For example, it can just enable MMIO or IO alone.  In that case, this will disable the UART, right ?

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: Sure. Let me update this.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
myguitar: Okay. Got it what you mean. Let me remove this unnecessary if condition.

myguitar:
Updated according to review comment.

mauricema (approved):
Thank for the patch.  It looks good to me.

myguitar:
Let me do a little more testing.

myguitar:
Updated to cover UEFI payload in QEMU.

PR #1109, opened 2021-04-09
Update boot option Template

gdong1:
Update Boot Flags to list PreOs and Extra image.
Hide other options when preOS or Extra Image is selected.


mauricema (approved):
Looks ok to me.

PR #1112, opened 2021-04-12
Enable SBL call into extra module in boot option

mauricema:
SBL allows extra module to be called before tranfering into the main
boot option. For example, RTCM module can be called for boot option
with TCC feature support. This patch enabled this support.  Since the
extra module might have different ARCH mode from current SBL mode,
thunk will be provided if mismatching is detected.

### DIFF
--- a/PayloadPkg/Include/Library/PayloadEntryLib.h
+++ b/PayloadPkg/Include/Library/PayloadEntryLib.h
@@ -8,6 +8,19 @@
 #ifndef __PAYLOAD_ENTRY_LIB_H__
 #define __PAYLOAD_ENTRY_LIB_H__
 
+typedef  struct {
+  UINT32   Signature;
+  UINT32   HeapBase;
+  UINT32   HeapSize;
+  UINT32   RsvdBase;
+  UINT32   RsvdSize;
+  UINT32   Argc;
+  UINT32   Argv[4];
+} PLD_EXTRA_MOD_ARGS;
+
+typedef  EFI_STATUS (EFIAPI *PLD_MODULE_ENTRY) \
+         (VOID *HobList, PLD_EXTRA_MOD_ARGS *Params);
+
 /**
   The payload common Entry Point for C code.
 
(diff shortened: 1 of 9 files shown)


PayloadPkg/Include/Library/PayloadEntryLib.h
```diff
@@ -8,6 +8,19 @@
...
+  UINT32   HeapBase;
+  UINT32   HeapSize;
+  UINT32   RsvdBase;
+  UINT32   RsvdSize;
+  UINT32   Argc;
+  UINT32   Argv[4];
+} PLD_EXTRA_MOD_ARGS;
+
+typedef  EFI_STATUS (EFIAPI *PLD_MODULE_ENTRY) \
+         (VOID *HobList, PLD_EXTRA_MOD_ARGS *Params);
```
gdong1, line 22: It could be great if we could use a single entry point for PRE_OS_ENTRYPOINT and PLD_MODULE_ENTRY.
Maybe we could check it late

mauricema: Yes, I think it needs to be aligned later on.  MOD_ENTRY has more parameters to pass.

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -1038,9 +1038,10 @@ StartBootImages (
       break;
     }
     if ((LoadedExtraImage != NULL) && ((LoadedExtraImage->Flags & LOADED_IMAGE_RUN_EXTRA) != 0)) {
-      Status = StartBooting (LoadedExtraImage);
+      // For now, only RTCM will be supported.
+      Status = CallExtraModule (PLD_EXTRA_MOD_RTCM, LoadedExtraImage);
       if (EFI_ERROR (Status)) {
-        return Status;
+        break;
       }
```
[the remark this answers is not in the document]
mauricema: ok, let me change it back to try next boot option.

mauricema:
Updated commit per review comments

gdong1 (approved):

PR #1113, opened 2021-04-13
[WHL] Clear power button status on normal boot flow

mauricema:
When S3 resume fails for some reason, the power button status will
be not be cleared on next boot. It might trigger shutdown event in
payload. This patch will clear power button status bit on normal
boot flow so that system can still boot properly.


myguitar (approved):
Looks good to me

PR #1115, opened 2021-04-14
[TGL] Fix boot issue and enable SMM rebase

gdong1:
With latest core code change, it would boot hang.
So need set SMM related PCD early and update platform
to enable SMM rebase.


mauricema (approved):
Looks good.

PR #1116, opened 2021-04-14
Fix buffer overflow for copy in S3SaveRestore lib

stalamudupula:
For appending Save/Restore structs in TSEG area,
bootloader should reserve space for TotalSize and
for certain structs, only header info should be
actually populated. Rest should be all Zeros.

### DIFF
--- a/BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
+++ b/BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
@@ -61,6 +61,7 @@ TriggerPayloadSwSmi (
   designated for S3 save/restore purpose.
 
   @param    DataPtr               Address of the structure to be copied to TSEG
+  @param    IsHdrOnly             Reserve TotalSize, but populate only Header info
 
   @retval   EFI_OUT_OF_RESOURCES  If SmmSize is exceeding 4KiB
   @retval   EFI_OUT_OF_RESOURCES  If appeding new struct exceeds SmmSize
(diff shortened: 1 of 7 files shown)


BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -95,8 +95,18 @@ AppendS3Info (
...
   }
+
   CopyMem ((VOID *)CommHdr, DataPtr, ((BL_PLD_COMM_HDR*)DataPtr)->TotalSize);
 
+  //
+  // If the ID=SMMBASE_INFO_COMM_ID, then we are only copying the
+  // header and rest of TotalSize is placeholder for payload to fill in.
+  // So, do a ZeroMem on SmmBaseInfo data (TotalSize - sizeof(header)) 
+  //
+  if (CommHdr->Id == SMMBASE_INFO_COMM_ID) {
```
mauricema, line 106: this seems to be very specific to certain ID.
Do you think we should add a function parameter to indicate we want to zero the payload part or not while appending ?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
```diff
@@ -95,8 +95,18 @@ AppendS3Info (
...
+    ZeroMem ((UINT8 *)CommHdr + sizeof(BL_PLD_COMM_HDR), ((BL_PLD_COMM_HDR*)DataPtr)->TotalSize - sizeof(BL_PLD_COMM_HDR));
```
mauricema, line 107: BTW, I saw many typecasting for (BL_PLD_COMM_HDR*)DataPtr.  So why don't we define a new variable  BL_PLD_COMM_HDR  *PldCommHdr to avoid these ?

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: Agree. Will add another param to indicate valid payload data.

BootloaderCorePkg/Library/S3SaveRestoreLib/S3SaveRestore.c
stalamudupula: Sure. Let me change it.

mauricema (approved):
Looks good to me.

PR #1117, opened 2021-04-14
[CML] enable SMM rebase

stanleyintel:
Set SMM related PCD early and update platform
to enable SMM rebase


mauricema (approved):
Looks good to me.

PR #1119, opened 2021-04-14
Add support for symbolic links to Ext23Lib

jgutbub:
Some of the boot option file paths used when
booting with OS Loader payload are failing
because the Ext23Lib does not support symbolic
soft links (e.g. ln -s <file> <link>). This
patch adds support for loading the soft link
succesfully.


mauricema (approved):
Looks good to me.  Thank for for enhancing it.

PR #1125, opened 2021-04-16
Set PayloadId earlier in Stage2

mauricema:
QEMU currently set PayloadId in Post PCI enumeration. To be in
sync with other platform, this patch moves it to PreSiliconInit
hook point.


gdong1 (approved):
This is QEMU platform specific change. Better add [QEMU] in the patch title.

myguitar (approved):
Looks good to me

PR #1130, opened 2021-04-22
[EHL] fix s0ix issue for pcie attach

koktong-ong:
S0ix failed to enter when pcie card is attaching.
The _OFF method of RP01 power resource didnt get
called upon entering s0ix.
Workaround with performing the action in _OFF method
in low power idle entering callback.


mauricema (approved):
Looks good.

PR #1132, opened 2021-04-26
Add ELF image support for OsLoader

mauricema:
This patch enhanced OsLoader to support ELF image boot on top of
PE, FV, Multiboot and Kernel. As a result, a new image flag
LOADED_IMAGE_ELF is added to indicate ELF image type.

Verified SBL can boot uboot ELF image (non-Multiboot) on QEMU.


myguitar (approved):
Looks good to me

Issue #1136, opened 2021-04-27
Dockerfile installs incompatible git version


fitzged:
With the change added to verify correct tool versions are being used (bc5584d176878a154bb79cd8a94487dceb215670), the Dockerfile needs updating to pull the latest git version for Ubuntu 18.04 from the ppa:git-core/ppa repo, otherwise builds fail:
Checking Toolchain Versions...
- /usr/bin/python: Version 3.6.9 (>= 3.6.0) [PASS]
- /usr/bin/openssl: Version 1.1.1 (>= 1.1.0g) [PASS]
- /usr/bin/nasm: Version 2.13.02 (>= 2.12.02) [PASS]
- /usr/bin/iasl: Version 20160831 (>= 20160422) [PASS]
- /usr/bin/git: Version 2.17.1 (>= 2.20.0) [FAIL]
- /usr/bin/gcc: Version 7.5.0 (>= 7.3) [PASS]
...Failed! Please check toolchain versions!

myguitar:
Thanks for reporting 'git' version check issue. #1137 will print recommended 'git' version if not met and continue to build.

fitzged:
Great, thanks @myguitar that will work for me.

myguitar:
No problem. Let me close this.

PR #1137, opened 2021-04-27
[Build] Make 'git' version checking optional

myguitar:
The build system is using 'git', it doesn't require the latest git.
This will show 'git' recommend version if not met,
and will address issue#1136.


mauricema (approved):
Agree.  This might be a better approach.

myguitar:
This will address Issue #1136 .

PR #1139, opened 2021-04-29
Allow Ext23Lib symbolic links to have longer paths

jgutbub:
In the Ext23Lib support for symbolic links was recently enabled
but there was a limitation imposed to use the CFG data boot
option filepath limit of 16 bytes which does not need to be also
be imposed on symbolic link paths. This will allow symbolic link
paths to be up to 260 characters long.


myguitar (approved):
Looks good to me. Thanks for the quick fix!

mauricema (approved):
Looks good.

PR #1141, opened 2021-05-05
[EHL] ProcHot update

koktong-ong:
Current PROCHOT configuration has PROCHOT output set to disable.
i.e. SoC is not able to send PROCHOT to external platform.
In order to compliance for FuSa Condition of Use per Safety Manual.
For FuSa requirement, external prochot signaling
shall be enable (bit 21 = 0),
we need SBL IFWI to set bit 21 = 0 for msr(0x1fc).
VID: U0G3T36100707 (Static part)


Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Power.yaml
```diff
@@ -368,7 +368,7 @@
       help         : >
                      Enable or Disable PROCHOT# signal being driven externally; 0- Disable; <b>1- Enable</b>.
       length       : 0x01
-      value        : 0x01
+      value        : 0x00
```
mauricema, line 371: So we disable PROCHOT# signal being driven externally ?
Is this the requirement for FuSa ?

koktong-ong: We need to enable ProcHotOut for Fusa requirement.
The UPD is indicating "Disable"ProcHotOut so we're disable this to enable.

koktong-ong: Its already verified by PMC team with this upd changes.

mauricema: OK. got it.   But does it has impact on other SKU ?    
Do you think it is better to detect FuSa SKU at runtime and then enable it in SBL ?

koktong-ong: let me check with PMC team again for this

koktong-ong: Confirmed with PMC team its only enable for Fusa sku.
Patch submitted as well.

mauricema (approved):
Looks good to me.

myguitar (approved):

PR #1143, opened 2021-05-06
[TGL] Remove bpmgen2.params

lsubash:
User have to generate defaut bpmgen params using
BpmGen2GUI.exe available with tool package and use
it while Ifwi stitch. Sbl stitching utility overrides
the some of the default params with ones defined
in stitch config files.


mauricema (approved):
Looks good.

PR #1146, opened 2021-05-10
[EHL] GpioLock update

koktong-ong:
Issue seen in Yocto RT kernel that PSE GBE0 Transmission.
Found out that GPPC_A_5.pmode=0 (RGMII0_TXCTL) which expected
to be 0x1.
override become 0x0.
After investigation these PADCFGLOCK_GPP_A_0 and PADCFGLOCKTX_GPP_A_0
registers should be 0xFFFFFF to lock the GPIO GPPC_A_x to prevent
modificaiton to the GPIO.
Implemented GpioLock function in GpioLib.c to fix this issue.

### DIFF
--- a/Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -751,6 +751,7 @@ BoardInit (
     if (GetBootMode() != BOOT_ON_FLASH_UPDATE) {
       UpdatePayloadId ();
     }
+    GpioLockPads();
     break;
   case PostSiliconInit:
     // Set TSEG base/size PCD
--- a/Silicon/ElkhartlakePkg/Library/GpioLib/GpioHelpersLib.c
+++ b/Silicon/ElkhartlakePkg/Library/GpioLib/GpioHelpersLib.c
@@ -25,6 +25,8 @@ typedef struct {
   //
   UINT32  OutputState;
 } GPIO_UNLOCK_HOB_DATA;
+GLOBAL_REMOVE_IF_UNREFERENCED GPIO_UNLOCK_HOB_DATA *mGpioUnlockData = NULL;
+GLOBAL_REMOVE_IF_UNREFERENCED UINT32 mGpioUnlockDataRecords = 0;
 
 /**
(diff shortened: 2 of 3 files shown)


Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -751,6 +751,7 @@ BoardInit (
     if (GetBootMode() != BOOT_ON_FLASH_UPDATE) {
       UpdatePayloadId ();
     }
+    GpioLockPads();
```
stalamudupula, line 754: A better place to call this would in GpioInit.c inside GpioConfigurePads() or GpioPadConfigTable() after the Gpio Programming is done?

stalamudupula: Yes, GpioLockPads() is supported in CommonSoc GpioLib. But I didnt call it from GpioInit.c in CommonSoc lib. Just because it is not clear yet if all projects want to lock their configs in Stage1/2 after Gpio programming.

Moving to CommonSoc GpioLib in general is a good idea , makes code much cleaner and we can get rid of redundant code. But I dont think it is going to help you with LockPads here. You still have to do it the way it is done in this patch-set.

koktong-ong: @stalamudupula if GpioLockPads called once either in end of GpioConfigurePads() or GpioPadConfigTable() then we need to unlock before configuring it? Or will it be better if we just lock it once all GPIO configuration is done?

stalamudupula: You are correct. SInce we are doing Gpio init multiple times, it is difficult to identify inside GpioInit.c when to lock. You can leave it as it is now. Thanks.

koktong-ong: @LeanSheng As per discussion i'll open another PR for the GPIO common lib porting and let's merge this patch for GpioLock fix.

mauricema (approved):

myguitar (approved):

PR #1151, opened 2021-05-13
[QEMU] Fix compiler error for NOOPT build in Windows

vincentchchen:
- fix __aullshr link error for NOOPT build in Windows
  due to compiler intrinsics functions
- adjust Stage1A/TopSwap/OsLoader FD size for NOOPT target
- adjust Stage2 size for NOOPT target when DEBUG FSP is used

### DIFF
--- a/PayloadPkg/FirmwareUpdate/CsmeFwUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/CsmeFwUpdate.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2017 - 2019, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2021, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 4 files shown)


PayloadPkg/FirmwareUpdate/CsmeFwUpdate.c
```diff
@@ -130,9 +130,9 @@ CsmePciReadBuffer (
     return EFI_INVALID_PARAMETER;
   }
 
-  Bus = (UINT8)((StartAddress >> 24) & 0xFF);
-  Device = (UINT8)((StartAddress >> 16) & 0xFF);
-  Function = (UINT8)((StartAddress >> 8) & 0xFF);
+  Bus = (UINT8)((RShiftU64(StartAddress, 24)) & 0xFF);
```
mauricema, line 133: since upper 32 bit is not used at all for Bus/Device/Function calculation.
I would recommend to use a new UINT32 to hold the PciAddress.  Then it is safe to do direct shift without using RShiftU64.
PciAddress = (UINT32)StartAddress;
Bus = (UINT8)((PciAddress >> 24) & 0xFF);

Platform/QemuBoardPkg/BoardConfig.py
```diff
@@ -123,6 +125,8 @@ def __init__(self, *args, **kwargs):
             self.NON_REDUNDANT_SIZE = 0x400000
         else:
             self.TOP_SWAP_SIZE      = 0x010000
+            if self.NO_OPT_MODE:
```
mauricema, line 128: TOP_SWAP_SIZE should be adjusted by build.  It has to match the hardware topswap region size set in FIT.  For QEMU, it is virtual, but should match the QEMU firmware update script.

mauricema (approved):
Looks good to me.

myguitar (changes requested):
Looks good to me, but can you rebase this PR on top of the latest master?
`Fix PCI address check in CsmeFwUpdate` commit has been already merged.

myguitar (approved):

PR #1154, opened 2021-05-14
Fix eMMC controller driver issue

### LINKED ISSUE #1153: [APL] eMMC boot does not work any more
With latest code on APL,  We could not boot from eMMC devices.
All data read from eMMC failed.

mauricema:
When the bus master decoding was disabled by default in PCI bus, the
device driver should try to enable it on its own. However, eMMC
driver does not have the enabling code. It caused the eMMC access
failure.

It fixed #1153.


myguitar (approved):
Looks good to me

PR #1155, opened 2021-05-17
Generate signed slimboot binary to output path

lsubash:
Currently final IFWI gets generated output path.
Copy the signed slimboot binary to out path.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
@@ -501,6 +501,11 @@ def main():
                     type=str, required=True,
                     help='specify the platform specific stitch config file')
 
+    ap.add_argument('-op',
+                     dest='outpath',
+                     default = '',
+                     help = "Specify path to write output IFIW and signed bin files")
+
     args = ap.parse_args()
 
     stitch_cfg_file = load_source('StitchIfwiConfig', args.config_file)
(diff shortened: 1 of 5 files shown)


Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -544,6 +544,10 @@ def main():
     ifwi_file_name = 'sbl_ifwi_%s.bin' % (args.platform)
     shutil.copy(generated_ifwi_file, ifwi_file_name)
 
+    generated_signed_sbl =  os.path.join(work_dir, 'Temp', 'SlimBootloader.bin')
+    sbl_file_name = 'SlimBootloader_%s.bin' % (args.platform)
```
gdong1, line 548: This patch would copy the signed SBL image to current folder. Late capsule image and other images might also be added by StitchIfwi.py. To avoid copy too many files to current folder, maybe we could copy the signed SBL image to other places. e.g. copy to Outputs\<platform> if the folder exists.
For the signed SBL image, maybe we could rename it sbl_<platform>_signed.bin?

lsubash: @gdong1 would you recommend to add new parameter for stitch outputs as "-outpath .."?

gdong1: Adding output path is also fine to me. then we could use current folder if  outpath is not provided.

lsubash: Yes added outpath option for user to specify.

myguitar (approved):
Looks good to me. But, it would be better to make a common StitchIfwi script.

gdong1:
> good to me. But, it would be better to make a common StitchIfwi
yes, I think the current StitchIfwi is almost same for different project. we could do it in a separate patch.

gdong1 (approved):

mauricema (commented):
Looks good.

PR #1158, opened 2021-05-19
Fix return code of SearchDirectory

stanleyintel:
This patch changes the return code of SearchDirectory
to EFI_NOT_FOUND when a file to search cannot be found


mauricema (approved):
Looks good to me.

PR #1159, opened 2021-05-19
[QEMU] Support usb-storage with qemu-xhci or nec-usb-xhci

stanleyintel:
This patch adds a new platform device to support a usb-storage
on a emulated xhci bus device (qemu-xhci or nec-usb-xhci) where
the XHCI controller locates at PCI 0x0:0x3:0x0.


mauricema (commented):
We tested XHCI on QEMU before. 
The XHCI device  number could change depends on your QEMU command option.   
For example, "-device qemu-xhci,id=xhci,bus=pcie.0,addr=4" will put XHCI on device 4.   Since device 3 has been used for SD and NVMe,  it is not recommended to use 3 for XHCI again.  
The proper way to handle this is to specify the PCI address in QEMU command line.

stanleyintel:
> We tested XHCI on QEMU before.
> The XHCI device number could change depends on your QEMU command option.
> For example, "-device qemu-xhci,id=xhci,bus=pcie.0,addr=4" will put XHCI on device 4. Since device 3 has been used for SD and NVMe, it is not recommended to use 3 for XHCI again.
> The proper way to handle this is to specify the PCI address in QEMU command line.

Thanks. The QEMU command option is a better solution. Let me try the solution first and close the pull request if everything is okay.

PR #1162, opened 2021-05-21
Support ELF32/64 format payloads regardless of operation mode

myguitar:
This allows to load and execute ELF32 or ELF64 format payloads
regardless of Ia32 or X64 SBL operation modes.


mauricema (approved):
Looks good to me.

PR #1164, opened 2021-05-25
[FWU] Fix a pointer misusing in capsule name

myguitar:
This will fix an invalid pointer operation while copying capsule filename
from capsule info config data to a unicode buffer.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
+++ b/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
@@ -283,7 +283,7 @@ LoadCapsuleImage (
   // Find capsule using the name provided in configuration data
   //
   if (CapsuleInfo->FileName[0] != 0) {
-    AsciiStrToUnicodeStrS ((CONST CHAR8 *)(&CapsuleInfo->FileName), FileName, MAX_FILE_LEN);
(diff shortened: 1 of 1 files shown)


PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -283,7 +283,7 @@ LoadCapsuleImage (
   // Find capsule using the name provided in configuration data
   //
   if (CapsuleInfo->FileName[0] != 0) {
-    AsciiStrToUnicodeStrS ((CONST CHAR8 *)(&CapsuleInfo->FileName), FileName, MAX_FILE_LEN);
+    AsciiStrToUnicodeStrS ((CONST CHAR8 *)&CapsuleInfo->FileName[0], FileName, MAX_FILE_LEN);
```
mauricema, line 286: I think it might be better to use  sizeof(CapsuleInfo->FileName) for the 3rd argument to ensure we don't have buffer overflow.   the MAX_FILE_LEN could be changed by people,  but it does not have impact to CAPSULE_INFO_CFG_DATA structure buffer size.

mauricema (approved):
Looks good.

PR #1168, opened 2021-05-26
[CFL] Resolve StitchIfwi.py issue

jgutbub:
Recent addition of '-op' argument may have
unintentionally created an issue by using an
unassigned variable 'work_dir' where the
intended variable to be used may have been
'stitch_dir'. This patch resolves the issue
and able to successfully stitch an IFWI once
again.

### DIFF
--- a/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/CoffeelakeBoardPkg/Script/StitchIfwi.py
@@ -544,11 +544,11 @@ def main():
         raise Exception ('Stitching process failed !')
     os.chdir(curr_dir)
 
-    generated_ifwi_file = os.path.join(work_dir, 'Temp', 'Ifwi.bin')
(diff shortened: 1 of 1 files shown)


lsubash (approved):
Looks good

mauricema (commented):
Do we have similar issue for the same script for other platforms ?

jgutbub:
> Do we have similar issue for the same script for other platforms ?

From what I can see it looks like no but I will keep an eye on the upcoming Jenkins build if this same issue occurs on any other stitching flows. From the recent patch where this issue was introduced I see 'work_dir' used in the other stitching scripts, CFL appears to be the only one using 'stitch_dir' (more motivation to create a common script it seems):

https://github.com/slimbootloader/slimbootloader/pull/1155/files

mauricema:
OK, looks good.

lsubash:
I have verified with EHL, TGL , CML when this change got pushed. 
Missed verifying CFL ;)

PR #1169, opened 2021-05-27
[TGL] Fix PlatformId override issue

gdong1:
If the PlatformId is already set by stitch data, the platform code
should skip platform settings to avoid PlatformID override.
This patch adds the PlatformId check in platform code.


mauricema (approved):
Looks good to me.

PR #1172, opened 2021-05-27
Disable debug prints before serial init

stalamudupula:
Set debug print error level to 0 by default.
This will prevent code to print debug messages via
UART, if it is not yet initielized.

After PostTempRamInit, when platforms did UartInit,
we can set it back to default PCD value, or leave it
as it is if the platform has set a custom level.


myguitar (approved):
Looks good to me

mauricema (approved):
Looks good to me.

PR #1179, opened 2021-06-02
Get RTCM from OS boot media instead IFWI

gdong1:
RTCM will sit in OS boot media instead of IFWI image
So update SBL logic to load RTCM from OS boot media.

If the RTCM could not be found, SBL should continue
boot normal OS.

### DIFF
--- a/PayloadPkg/OsLoader/LoadImage.c
+++ b/PayloadPkg/OsLoader/LoadImage.c
@@ -63,19 +63,6 @@ GetBootImageFromIfwiContainer (
   LoadedImage->ImageData.Size = ImageSize;
   LoadedImage->ImageData.AllocType = ImageAllocateTypePage;
 
-  if ((Image->ContainerSig == SIGNATURE_32 ('I', 'P', 'F', 'W')) && (Image->ComponentName == SIGNATURE_32 ('T', 'C', 'C', 'R'))) {
-    //
-    // TCC RTCM image need reserved memory
-    //
-    LoadedImage->ImageData.Addr = AllocateReservedPages (EFI_SIZE_TO_PAGES(ImageSize));
-    if (LoadedImage->ImageData.Addr == NULL) {
-      return EFI_OUT_OF_RESOURCES;
-    }
-    CopyMem(LoadedImage->ImageData.Addr, Buffer, ImageSize);
-    FreePages (Buffer, EFI_SIZE_TO_PAGES (ImageSize));
-    LoadedImage->Flags |= LOADED_IMAGE_RUN_EXTRA;
-  }
-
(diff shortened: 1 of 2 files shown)


PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -312,7 +299,16 @@ GetBootImageFromFs (
     goto Done;
   }
 
-  Image = AllocatePages (EFI_SIZE_TO_PAGES (ImageSize));
+  if (LoadedImage->LoadImageType == LoadImageTypeExtra0) {
+      //
```
mauricema, line 303: fix indentation.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -312,7 +299,16 @@ GetBootImageFromFs (
     goto Done;
   }
 
-  Image = AllocatePages (EFI_SIZE_TO_PAGES (ImageSize));
+  if (LoadedImage->LoadImageType == LoadImageTypeExtra0) {
+      //
+      // Currently extra image is used on by RTCM which need allocate reserved memory.
+      // Need revise this logic if there is other use case late.
+      //
+     Image = AllocateReservedPages (EFI_SIZE_TO_PAGES(ImageSize));
```
mauricema, line 307: Need to handle error condition if Image is NULL.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -857,6 +853,12 @@ LoadBootImages (
     LoadedImagesInfo->LoadedImageList[Index] = LoadedImage;
 
     if (EFI_ERROR (Status)) {
+      if (Index >= LoadImageTypeExtra0) {
+        // Continue boot if load extra image failed.
+        Status = EFI_SUCCESS;
+        continue;
```
myguitar, line 859: If Extra0 fails, do we need to load following Extra1/2/3..?

gdong1: there is no relation in Extra0/1/2/3. 
As long as Extra1 is configured in the boot option, current logic will load Extra1 regardless Extra0 load success or not.
But currently we don't have this use case.

PayloadPkg/OsLoader/LoadImage.c
gdong1: sure

mauricema (approved):
Looks good.

PayloadPkg/OsLoader/LoadImage.c
myguitar: Understood. Thanks!

myguitar (approved):
Looks good to me

PR #1184, opened 2021-06-04
[TGL] Set PayloadId to "AUTO" by default

gdong1:
Update PayloadId to "AUTO" to align with other platforms to
have consistent behavior.


mauricema (approved):
Looks good.

PR #1185, opened 2021-06-09
Add a Pci Enum Hook function

stalamudupula:
This patch adds a platform hook function ability
in Pci Enum Lib to enable platform to perform a
PCI Enum specific work-around routine to detect
some controllers.

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dsc
+++ b/BootloaderCorePkg/BootloaderCorePkg.dsc
@@ -282,6 +282,7 @@
   gPlatformModuleTokenSpaceGuid.PcdFuncCpuInitHook   | 0x00000000
   gPlatformModuleTokenSpaceGuid.PcdFspsUpdPtr        | 0x00000000
   gPlatformModuleTokenSpaceGuid.PcdAcpiTableTemplatePtr | 0
+  gPlatformModuleTokenSpaceGuid.PcdPciEnumHookProc   | 0x00000000
 
   gPlatformModuleTokenSpaceGuid.PcdPciResAllocTableBase    | 0x00000000
 
--- a/BootloaderCorePkg/Include/Library/PciEnumerationLib.h
+++ b/BootloaderCorePkg/Include/Library/PciEnumerationLib.h
@@ -8,6 +8,26 @@
 #ifndef __PCI_ENUMERTION_LIB_H__
 #define __PCI_ENUMERTION_LIB_H__
 
+typedef enum {
+  ///
+  /// This notification is only applicable to PCI-PCI bridges and
+  /// indicates that the PCI enumerator is about to begin enumerating
+  /// the bus behind the PCI-PCI Bridge. This notification is sent after
+  /// the primary bus number, the secondary bus number and the subordinate
+  /// bus number registers in the PCI-PCI Bridge are programmed to valid
+  /// (not necessary final) values
+  ///
+  EfiPciBeforeChildBusEnumeration,
+
+  ///
+  /// This notification is sent before the PCI enumerator probes BAR registers
+  /// for every valid PCI function.
+  ///
+  EfiPciBeforeResourceCollection
+} EFI_PCI_CONTROLLER_RESOURCE_ALLOCATION_PHASE;
+
+typedef VOID   (EFIAPI *PLATFORM_PCI_ENUM_HOOK_PROC) (UINT8 Bus, UINT8 Dev, UINT8 Fun, EFI_PCI_CONTROLLER_RESOURCE_ALLOCATION_PHASE Phase);
+
 /**
  Enumerates the PCI devices allocates the required memory resource.
  Program the allocated memory resource to PCI BAR.
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
@@ -13,6 +13,7 @@
 #include <Library/SortLib.h>
 #include <Library/HobLib.h>
 #include <InternalPciEnumerationLib.h>
+#include <Library/PciEnumerationLib.h>
 #include <Library/BootloaderCommonLib.h>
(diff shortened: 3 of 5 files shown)


BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -282,6 +282,7 @@
   gPlatformModuleTokenSpaceGuid.PcdFuncCpuInitHook   | 0x00000000
   gPlatformModuleTokenSpaceGuid.PcdFspsUpdPtr        | 0x00000000
   gPlatformModuleTokenSpaceGuid.PcdAcpiTableTemplatePtr | 0
+  gPlatformModuleTokenSpaceGuid.PcdFuncPciEnumHook   | 0x00000000
```
mauricema, line 285: maybe PcdPciEnumHookProc ?   It is easier to know it is a function pointer in this way.

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -861,6 +863,17 @@ PciScanBus (
...
 
+        //
+        // Some controllers require read-write 0xffffffff to be detected.
+        // Perform this in Hook function.
+        //
+        Address   = PCI_EXPRESS_LIB_ADDRESS (StartBusNumber, Device, Func, PCI_BRIDGE_SECONDARY_BUS_REGISTER_OFFSET);
+        Register  = (UINT8)PciExpressRead8 (Address);
+        PlatformPciEnumHook = (PLATFORM_PCI_ENUM_HOOK)(UINTN)PcdGet32 (PcdFuncPciEnumHook);
+        if ((Register != 0) && (PlatformPciEnumHook != NULL)) {
+          PlatformPciEnumHook ((UINT32)Register);
```
mauricema, line 874: Should we use a more generic function similar to EDK2 PCI bus driver?
PlatformPciEnumHookProc (Phase, PciAddress) ?

BootloaderCorePkg/Include/Library/PciEnumerationLib.h
```diff
@@ -8,6 +8,8 @@
 #ifndef __PCI_ENUMERTION_LIB_H__
 #define __PCI_ENUMERTION_LIB_H__
 
+typedef VOID   (*PLATFORM_PCI_ENUM_HOOK) (UINT32 SecBus);
```
myguitar, line 11: Add EFIAPI
`(EFIAPI *PLATFORM_PCI_ENUM_HOOK)`

Parameters should be `Bus, Device, Func, Phase`.

```
typedef VOID   (*PLATFORM_PCI_ENUM_HOOK) (UINT32 SecBus);
->
typedef VOID   (EFIAPI *PLATFORM_PCI_ENUM_HOOK) (UINT8 Bus, UINT8 Device, UINT8 Func, EFI_PCI_CONTROLLER_RESOURCE_ALLOCATION_PHASE Phase);
```

BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -861,6 +863,17 @@ PciScanBus (
...
+        }
+
```
myguitar, line 876:
```
//
// Nofify EfiPciBeforeChildBusEnumeration
//
PlatformPciEnumHook = (PLATFORM_PCI_ENUM_HOOK)(UINTN)PcdGet32 (PcdFuncPciEnumHook);
if (PlatformPciEnumHook != NULL) {
  PlatformPciEnumHook (StartBusNumber, Device, Func, EfiPciBeforeChildBusEnumeration);
}
```

BootloaderCorePkg/Include/Library/PciEnumerationLib.h
myguitar: Do we have a better place in BootloaderCommonPkg instead of BootloaderCorePkg?

stalamudupula: Theres PlatformHookLib.h in CommonPkg. How does it sound?

mauricema: PlatformHookLib is a lib class defined by EDK2.  So let us leave it.
Since this PCD is defined in BootloaderCorePkg.dec,  the related definition should also be in the same package.  PciEnumerationLib.h looks good to me.  Any special reason why we want to move it into BootlaoderCommonPkg ?  Which library in BootlaoderCommonPkg will use this ?

myguitar: This was to avoid BootloaderCorePkg reference from Platform package directly. We have PciRootBridgeInfoGuid.h in BootloaderCommonPkg for some PciEnumeration interfaces, but it looks not suitable file. I am still thinking we need to avoid CorePkg reference from Platform/Silicon package. Any advices on this?

mauricema (approved):
It looks good to me.

BootloaderCorePkg/Include/Library/PciEnumerationLib.h
mauricema: When we defined the package dependency rules,  at that time, we allowed Platform/Silicon library to consume Core package.  And there are many instances already.  Do you see any issue to allow that ?

myguitar: Might be my bad memory. I thought the dependency rule does not allow Platform/Silicon to consume Core package like these.
- Common: independent to others
- Core: Common
- Platform: Common and Silicon
- Silicon: Common

If Core package can be referenced from Platform/Silicon, ACPI, MP or etc. libraries can be also used. What about limiting BootloaderCoreLib only?

myguitar: Let's take this dependency rule discussion to offline. Let me merge this PR. Thanks.

myguitar (approved):

PR #1188, opened 2021-06-10
Tgl simplify tcc enable

gdong1:
Enable TCC support


Platform/TigerlakeBoardPkg/BoardConfig.py
```diff
@@ -195,13 +194,34 @@ def __init__(self, *args, **kwargs):
...
+                cfg_dlt_file  = os.path.join(brd_cfg_src_dir, dlt_file[len (self._generated_cfg_file_prefix):])
+                lines         = open (cfg_dlt_file).read()
+
+                # Enable TCC in dlt file
+                if self.ENABLE_TCC:
+                    lines += open (os.path.join(brd_cfg_src_dir, 'CfgData_Tcc_Feature.dlt')).read()
+
+                # Write to generated final dlt file
+                output_cfg_dlt_file = os.path.join(build._fv_dir, dlt_file)
+                open(output_cfg_dlt_file, 'w').write(lines)
```
mauricema, line 224: where do you delete the generated files ?   
If you don't do it,  the generated dlt files will not be updated even when the source files are modified.

BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -621,7 +621,10 @@ def gen_config_file (fv_dir, brd_name, platform_id, pri_key, cfg_db_size, cfg_si
     # copy delta files
     dlt_list  = cfg_int[1:] + cfg_ext
     for dlt_file in dlt_list:
-        copy_expanded_file (os.path.join (brd_cfg_dir, dlt_file), os.path.join (fv_dir, dlt_file))
+        src_dlt_file = os.path.join (brd_cfg_dir, dlt_file)
+        if not os.path.exists(src_dlt_file):
```
mauricema, line 625: It is better to add a feature control  self.ENABLE_AUTOGEN_CFGDATA_DELTA and do this only when it is enabled.
It is avoid any impacts to other projects that don't generate DLT at build time.

Platform/TigerlakeBoardPkg/BoardConfig.py
gdong1: It is not necessary to delete the generated one in build folder since BoardConfig.py will re-generate it from source in every build.

mauricema: OK.  But you might want to check if the generated CfgDataStitch.py in Build folder still works as before.
The python file will use the DELTA files.

BootloaderCorePkg/Tools/BuildUtility.py
mauricema: If the DLT file name is provided by mistake on other projects,  it will report that it cannot open the delta file in the build FV folder, which is confusing to people.   To ensure it has the same behavior,  the logic should be check if the delta file name does not exist in CFGDATA source folder and it follows certain pattern (ex: Autogen_*.dlt ),  only in this case we update the delta path to point to the one in FV folder.

gdong1: To avoid error confusion, how about add below check before line 627? This way we could avoid check hardcoded string pattern "Autogen".
if not os.path.exists(src_dlt_file):
        raise Exception ('dlt file %s could not be found!' % (os.path.join (brd_cfg_dir, dlt_file)))

Platform/TigerlakeBoardPkg/BoardConfig.py
gdong1: yes.  the generated CfgDataStitch.py works with autogenerated dlt files.

BootloaderCorePkg/Tools/BuildUtility.py
mauricema: I am OK with it as long as it reports the actual delta file path.

Platform/TigerlakeBoardPkg/BoardConfig.py
mauricema: OK.

mauricema (approved):
Looks good to me.

PR #1192, opened 2021-06-10
[TGL] Add EC-less board support

mauricema:
Current TGL implementation assumed EC always exists on the board.
It does not work for EC-less design. In this patch, a EcEnable
CFGDATA is added to control it.


mauricema:
duplication.

PR #1199, opened 2021-06-18
Rework Fusa PT settings

pqueirosint:
- Add knob for setting up the ISI configuration file (isi_conf)
 - Fix PT test mask value to be aligned with default PT binary
   capabilities. Only add mask when PT is enabled.
 - "pt" parameter always requires isi_conf file. Dependency added
   automatically. isi_conf is also externally configurable since can
   be used without PT.

Signed-off-by: Pedro Queiros <pedro.queiros@intel.com>


LeanSheng (approved):
Looks good!

mauricema (approved):
Looks good to me.

PR #1204, opened 2021-06-24
[WIP]Allow dynamic setting for payload ID GPIO pin config

LeanSheng:
Here are the changes:
1. Add payload ID GPIO pin selection option to Common CfgData YAML.
2. Let each platform use Common CfgData to set payload ID gpio pin
3. Update dlt files with the default GPIO pin used for payload ID selection
4. Replace existing board cfg data PayloadSelGpio with the common cfg data version

Signed-off-by: Lean Sheng <lean.sheng.tan@intel.com>

### DIFF
--- a/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -820,37 +820,33 @@ UpdatePayloadId (
   VOID
   )
 {
-  GEN_CFG_DATA       *GenericCfgData;
-  PLATFORM_CFG_DATA  *PlatCfgData;
-  UINT32              PayloadId;
-  UINT32              GpioLevel;
+  EFI_STATUS      Status;
+  GEN_CFG_DATA    *GenCfgData;
+  UINT32          PayloadId;
+  UINT32          PayloadSelGpioData;
 
   if (GetBootMode() == BOOT_ON_FLASH_UPDATE) {
     return;
   }
 
-  PayloadId = 0;
-  GenericCfgData = (GEN_CFG_DATA *)FindConfigDataByTag (CDATA_GEN_TAG);
-  if ((GenericCfgData != NULL) && (GenericCfgData->PayloadId != AUTO_PAYLOAD_ID_SIGNATURE)) {
-    // If PayloadId is not AUTO, use it directly
-    PayloadId = GenericCfgData->PayloadId;
-  } else {
-    // If PayloadId is AUTO and GPIO PayloadId selection is enabled,
-    // adjust the PayloadId accordingly per GPIO level (0:UEFI  1:OsLoader).
-    GpioLevel = 1;
-    PlatCfgData = (PLATFORM_CFG_DATA *)FindConfigDataByTag (CDATA_PLATFORM_TAG);
-    if ((PlatCfgData != NULL) && (PlatCfgData->PayloadSelGpio.Enable != 0)) {
-      // The default GPIOSet to Pull Up 20K
-      GpioGetInputValue (PlatCfgData->PayloadSelGpio.PadInfo, 0x0C, &GpioLevel);
-      if (GpioLevel != 0) {
-        PayloadId = 0;
-      } else {
-        PayloadId = UEFI_PAYLOAD_ID_SIGNATURE;
-      }
-      DEBUG ((DEBUG_INFO, "Set PayloadId to 0x%08X based on GPIO config\n", PayloadId));
-    }
+  GenCfgData = (GEN_CFG_DATA *)FindConfigDataByTag (CDATA_GEN_TAG);
+  if (GenCfgData == NULL) {
+    ASSERT (FALSE);
+    return;
   }
 
+  if (GenCfgData->PayloadId != AUTO_PAYLOAD_ID_SIGNATURE || !GenCfgData->PayloadSelGpio.Enable) {
+    SetPayloadId (GenCfgData->PayloadId);
+    DEBUG ((DEBUG_INFO, "PayloadId = 0x%X\n", GenCfgData->PayloadId));
+    return;
+  }
+
+  // Switch payloads based on configured GPIO pin
+  // The default GPIOSet to Pull Up 20K
+  Status = GpioGetInputValue (GenCfgData->PayloadSelGpio.PinNum, 0x0C, &PayloadSelGpioData);
+  ASSERT (Status == RETURN_SUCCESS);
+  PayloadId = PayloadSelGpioData ? 0 : UEFI_PAYLOAD_ID_SIGNATURE;
+  DEBUG ((DEBUG_INFO, "Set PayloadId to 0x%X based on GPIO config\n", PayloadId));
   SetPayloadId (PayloadId);
 }
 
--- a/Platform/CommonBoardPkg/CfgData/CfgData_Common.yaml
+++ b/Platform/CommonBoardPkg/CfgData/CfgData_Common.yaml
@@ -26,6 +26,30 @@
                      Specify payload ID string. Empty will boot default payload. Otherwise, boot specified payload ID in multi-payload binary.
       length       : 0x04
       value        : ''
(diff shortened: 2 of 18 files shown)


Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -591,39 +591,30 @@ UpdatePayloadId (
...
 
   GenCfgData = (GEN_CFG_DATA *)FindConfigDataByTag (CDATA_GEN_TAG);
   if (GenCfgData == NULL) {
     ASSERT (FALSE);
     return;
   }
-  SetPayloadId (GenCfgData->PayloadId);
 
-  if (GetPayloadId () != AUTO_PAYLOAD_ID_SIGNATURE) {
+  if (GenCfgData->PayloadId != AUTO_PAYLOAD_ID_SIGNATURE || !GenCfgData->PayloadSelGpio.Enable) {
```
koktong-ong, line 605: This condition should be AND instead of OR?
Should we set default payload if invalid option was set? (eg. GenCfgData->PayloadId = 'UEFI' and GenCfgData->PayloadSelGpio.Enable = 1)?

LeanSheng: it has to be 'or' condition.
if the uefi payload is set cfgdata (payload id) = uefi, then it will set as per cfgdata. 
but if it is set as auto detection while gpio pin selection is not set, then it will still use the default value (null), which is osloader.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -820,37 +820,33 @@ UpdatePayloadId (
...
+  if (GenCfgData->PayloadId != AUTO_PAYLOAD_ID_SIGNATURE || !GenCfgData->PayloadSelGpio.Enable) {
+    SetPayloadId (GenCfgData->PayloadId);
+    DEBUG ((DEBUG_INFO, "PayloadId = 0x%X\n", GenCfgData->PayloadId));
+    return;
   }
 
+  // Switch payloads based on configured GPIO pin
+  // The default GPIOSet to Pull Up 20K
+  Status = GpioGetInputValue (GenCfgData->PayloadSelGpio.PinNum, 0x0C, &PayloadSelGpioData);
+  ASSERT (Status == RETURN_SUCCESS);
```
mauricema, line 847: I don't believe we need ASSERT here.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -820,37 +820,33 @@ UpdatePayloadId (
...
-      if (GpioLevel != 0) {
-        PayloadId = 0;
-      } else {
-        PayloadId = UEFI_PAYLOAD_ID_SIGNATURE;
-      }
-      DEBUG ((DEBUG_INFO, "Set PayloadId to 0x%08X based on GPIO config\n", PayloadId));
-    }
+  GenCfgData = (GEN_CFG_DATA *)FindConfigDataByTag (CDATA_GEN_TAG);
+  if (GenCfgData == NULL) {
+    ASSERT (FALSE);
```
mauricema, line 834: It is not critical,   so should remove this ASSERT.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -820,37 +820,33 @@ UpdatePayloadId (
...
+    return;
+  }
+
+  if (GenCfgData->PayloadId != AUTO_PAYLOAD_ID_SIGNATURE || !GenCfgData->PayloadSelGpio.Enable) {
```
mauricema, line 838: PayloadGPIO selection should only take effect if PayloadId is AUTO_PAYLOAD_ID_SIGNATURE.
If PayloadId is something else,  we should already use that regardless of the state of PayloadSelGpio.Enable.

Platform/CommonBoardPkg/CfgData/CfgData_Common.yaml
```diff
@@ -26,6 +26,30 @@
                      Specify payload ID string. Empty will boot default payload. Otherwise, boot specified payload ID in multi-payload binary.
       length       : 0x04
       value        : ''
+  - PayloadSelGpio :
+    - $STRUCT      :
+        name         : GPIO pin for switching payload
+        struct       : PAYLOAD_SEL_GPIO_PIN
+        length       : 0x04
+        value        : 0
+    - PinNum       :
```
mauricema, line 35: Is this generic for all platforms ?    How can the end user determine the PinNum value ?
Which Group and which pin ?     
The old CFL allows you to select group and pin.

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
LeanSheng: Sure will remove it

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
LeanSheng: will do thanks

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
LeanSheng: Actually it is a 'shortcut' to save additional line later down: if (GenCfgData->PayloadSelGpio.Enable) {
Thats why i combine both together, logically it will still work:
- payloadid will take effect if it is not auto
- payloadid will be default (null) if PayloadSelGpio.Enable not selected even payloadid is configured as AUTO_PAYLOAD_ID_SIGNATURE.

Do you think this will work? or should i split it out?

Platform/CommonBoardPkg/CfgData/CfgData_Common.yaml
LeanSheng: Yes this is generic, and yes for now it seems like 'magic number' (just like APL and WHL), but user can refer to each platform GPIO lib (eg. Silicon\CoffeelakePkg\Include\GpioPinsCnlH.h) to check for the pin number .

Yes CFL did expose the option in CFL Sil cfg data, but it is not being used anywhere in code, hence it wont take effect at all if user configure any specific group.

What do you think?

Platform/ApollolakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: If the default PayloadId is 0 and PayloadSelGpio.Enable = 0,  the expected behavior is to use the default PayloadId regardless of the GPIO level.

mauricema: you can use your code, but the logic has to be correct as stated above.

Platform/CommonBoardPkg/CfgData/CfgData_Common.yaml
mauricema: From user perspective, I think we should make it easy for user to select directly using group and pin number, ex: GPIO A10, D11, etc.  Magic number is hard for user to config.   SBL code can use the Group and Pin number to calculate the magic number required.

mauricema:
BTW, [Common][EHL][[TGL][CFL][APL][CML][CMLV] is too long. :)
for many platforms, just remove the scope tag.

mauricema:
Move to close state since it has no updates for more than one month.

PR #1208, opened 2021-06-25
[EHL] Fix Stage2 Boot Option checking

koktong-ong:
Fix mPchSciSupported flag checking before changing the
flag value of boot option.
This bug was introduced from commit
b78cbcf128866b8835116a6ba77d9e7ac5c8ab00


mauricema (approved):
Looks good.

PR #1210, opened 2021-06-29
GpioLib header clean-up

stalamudupula:
Inconsistent and redundant header files are removed.
All projects going forward

 - Use API declared in GpioLib.h
 - Provide instance of GpioSiLib.h
 - Use common defines in GpioConfig.h

[QEMU][APL][CFL][CML][CMLV]
 - Follow above header model
 - Have own instance of GpioLib

[EHL][TGL]
 - Follow above header model
 - Use common GpioLib instance


mauricema:
It is lots of changes.  It might be better later on to break it down into smaller commits.

mauricema (approved):
Looks good to me.

stalamudupula:
GpioLibInternal.h is separately created from GpioLib.h , to keep functions local to Common GpioLib and not necessarily needed to be exposed as API.

gdong1:
> It is lots of changes. It might be better later on to break it down into smaller commits.

Agree with Maurice. For such big change, had better to have one commit per platform and you could send out all the commits in one patch.  Anyway, you could do it next time.

gdong1 (approved):

PR #1211, opened 2021-06-29
Fix Linux build issue when adding UefiVariableLib

cshur:
When adding UefiVariableLib, there is a build issue in Linux.
Fix the build issue.

Signed-off-by: CSHur <cs.hur@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Library/UefiVariableLib/UefiVariableLib.c
+++ b/BootloaderCommonPkg/Library/UefiVariableLib/UefiVariableLib.c
@@ -2,7 +2,7 @@
   Implement ReadOnly Variable Services required by PEIM and install
   PEI ReadOnly Varaiable2 PPI. These services operates the non volatile storage space.
 
-Copyright (c) 2006 - 2019, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2006 - 2021, Intel Corporation. All rights reserved.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/UefiVariableLib/UefiVariableLib.c
```diff
@@ -691,7 +690,7 @@ GetVariableHeader (
       //
       // Variable header pointed by Variable is inconsecutive,
       //
-      LibStatus = GetLibraryData (PcdGet8 (PcdUefiVariableLibId), (VOID **)&VarStoreLibData);
+      GetLibraryData (PcdGet8 (PcdUefiVariableLibId), (VOID **)&VarStoreLibData);
```
mauricema, line 693: Can we add check for VarStoreLibData to ensure it is not NULL ?

BootloaderCommonPkg/Library/UefiVariableLib/UefiVariableLib.c
```diff
@@ -692,6 +692,11 @@ GetVariableHeader (
       // Variable header pointed by Variable is inconsecutive,
       //
       LibStatus = GetLibraryData (PcdGet8 (PcdUefiVariableLibId), (VOID **)&VarStoreLibData);
+      if (LibStatus == EFI_NOT_FOUND) {
```
mauricema, line 695: If LibStatus == EFI_NOT_FOUND, I think we should just return FALSE ?  
Allocating a new buffer will not work since all critical members in the structure won't be initialized.

stalamudupula: Also, Status can be anything, other than EFI_SUCCESS in failure case. So please check for EFI_ERROR(Status), instead of only EFI_NOT_FOUND.

mauricema (approved):
Looks good to me.

PR #1212, opened 2021-07-01
[EHL] Enable POSC for all boot option

koktong-ong:
Enable POSC for all boot medium in Cfgdata_BootOption.yaml
by default
If Non-Fusa sku was detect and the boot flag will
change to exclude POSC
User can modify the boot flag to exclude the POSC in
yaml file as well


ywho79 (approved):

mauricema (approved):
Looks good.

PR #1219, opened 2021-07-08
[Common] Fix config data boot option index

koktong-ong:
Current index only support up to 0xF as the CFGHDR_TMPL
as it amend into 0x05 (eg. 0x05$(1))
Update the logic to (0x050 + $(1)) instead to support
index larger than 0xF

### DIFF
--- a/Platform/CommonBoardPkg/Include/ConfigDataCommonDefs.h
+++ b/Platform/CommonBoardPkg/Include/ConfigDataCommonDefs.h
@@ -11,8 +11,8 @@
 // Note: Any Config TAG that is less than 0x100 will be considered as Common TAG
 // and it is applicable for all platforms.
 
-// 0x050 to 0x05F is reserved for boot option tag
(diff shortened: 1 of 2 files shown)


Platform/CommonBoardPkg/Include/ConfigDataCommonDefs.h
```diff
@@ -12,7 +12,7 @@
 // and it is applicable for all platforms.
 
 // 0x050 to 0x05F is reserved for boot option tag
-#define  MAX_BOOT_OPTION_CFGDATA_ENTRY  16
+#define  MAX_BOOT_OPTION_CFGDATA_ENTRY  30
```
myguitar, line 15: Any reason to use 30 instread of 32?
Can you update comment as well? `// 0x050 to 0x05F is reserved for boot option tag`

myguitar (approved):
Looks good to me

gdong1 (approved):

PR #1222, opened 2021-07-09
Fix error logs for capsule file too small

stanleyintel:
If the capsule file is too small, it should not only
show "the guid is not expected" that may mislead users
using a release-build image. This patch prints out
specific error if opening a small-than-expected capsule.


myguitar (approved):
Looks good to me

PR #1224, opened 2021-07-14
Fix SMM rebase S3 issue

gdong1:
Currently it will return a valid SMMBASE_INFO if SMMBASE_INFO_COMM_ID
is found in SMM S3 resume memory. It will cause issue in S3 path if there
is no one fill correct data when MpInit uses it to rebase SMM.
This patch adds a check to SMMBASE_INFO to avoid this issue.


rprangar (approved):
Looks good

myguitar (approved):
Looks good to me

PR #1241, opened 2021-08-01
[CFL] Fix reading bpmgen2 params

lsubash:
Default Bpmgen params created by bpmgen tool
has unicode characters. This patch fixes reading the
file with unicode chars.


myguitar (approved):
Looks good to me

PR #1245, opened 2021-08-03
[CorePkg] Add GetTempRamInfo() API

myguitar:
Some platforms need TempRam Base & Size information to calculate
FspmArchUpd StackBase & Size at runtime.
The TempRam Base & Size info will be only valid until TempRamExit.

### DIFF
--- a/BootloaderCorePkg/Stage1B/Stage1B.c
+++ b/BootloaderCorePkg/Stage1B/Stage1B.c
@@ -655,6 +655,11 @@ ContinueFunc (
   AddMeasurePoint (0x2060);
   ASSERT_EFI_ERROR (Status);
 
+  LdrGlobal = (LOADER_GLOBAL_DATA *)GetLoaderGlobalDataPointer ();
+  ASSERT (LdrGlobal != NULL);
+  LdrGlobal->CarBase = 0;
+  LdrGlobal->CarSize = 0;
+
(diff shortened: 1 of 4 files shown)


BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -655,6 +655,10 @@ ContinueFunc (
   AddMeasurePoint (0x2060);
   ASSERT_EFI_ERROR (Status);
 
+  LdrGlobal = (LOADER_GLOBAL_DATA *)GetLoaderGlobalDataPointer ();
```
jgutbub, line 658: I know it wasn't being checked before, but do we need to check if this pointer is NULL to avoid a KW hit?

myguitar: The global data pointer is set to near IDT base by line#431
`LdrGlobal      = (LOADER_GLOBAL_DATA *)(UINTN)MemPoolCurrTop;`
 and line#472
`SetLoaderGlobalDataPointer (LdrGlobal);`
So, this pointer will not have NULL. Just in case, let me add ASSERT().

myguitar: ASSERT is added in new PR. If KW hits here, it needs to be disposed as not an issue.

PR #1248, opened 2021-08-04
Cleanup Platform code to access LoaderGlobalData via APIs

myguitar:
This adds additional APIs to make Platform code use APIs to access
LoaderGlobalData instead of accessing variables directly.
- GetS3DataPtr()
- SetFeatureCfg()
- ClearFspHob()
- GetVerInfoPtr()

### DIFF
--- a/BootloaderCorePkg/Library/BootloaderCoreLib/BootloaderCoreLib.c
+++ b/BootloaderCorePkg/Library/BootloaderCoreLib/BootloaderCoreLib.c
@@ -444,3 +444,72 @@ GetTempRamInfo (
 
   return EFI_SUCCESS;
 }
+
+/**
+  This function retrieves the address of S3Data.
+
+  @retval    S3Data pointer address.
+
+**/
+VOID *
+EFIAPI
+GetS3DataPtr (
+  VOID
+  )
+{
+  return GetLoaderGlobalDataPointer()->S3DataPtr;
+}
+
+/**
+  This function sets features configuration.
+
+  @param Features   The features the platform supports.
+
+**/
+VOID
+EFIAPI
+SetFeatureCfg (
+  IN  UINT32  Features
+  )
+{
+  GetLoaderGlobalDataPointer()->LdrFeatures = Features;
+}
+
+/**
+  This function clears FSP Hob Data and pointer.
+
+**/
+VOID
+EFIAPI
+ClearFspHob (
+  VOID
+  )
+{
+  VOID                        *FspHobList;
+  EFI_HOB_HANDOFF_INFO_TABLE  *HandOffHob;
+  UINT32                       Length;
+
+  FspHobList = GetLoaderGlobalDataPointer()->FspHobList;
+  if (FspHobList != NULL) {
+    HandOffHob = (EFI_HOB_HANDOFF_INFO_TABLE *) FspHobList;
+    Length     = (UINT32)((UINTN)HandOffHob->EfiEndOfHobList - (UINTN)HandOffHob);
+    ZeroMem (HandOffHob, Length);
+
+    GetLoaderGlobalDataPointer()->FspHobList = NULL;
+  }
+}
+
+/**
+  This function retrieves the Version Information pointer.
+
+  @retval    Version Information pointer address.
+
+**/
+VOID *
+EFIAPI
+GetVerInfoPtr (
+  VOID
+  )
+{
+  return GetLoaderGlobalDataPointer()->VerInfoPtr;
+}
--- a/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -522,30 +522,6 @@ UpdateBlRsvdRegion ()
 }
 
 /**
-  Clear FSP HOB data
-
-**/
-VOID
-ClearFspHob (
-  VOID
-  )
-{
-  LOADER_GLOBAL_DATA          *LdrGlobal;
-  EFI_HOB_HANDOFF_INFO_TABLE  *HandOffHob;
-  UINT32                      Length;
-
-  LdrGlobal  = (LOADER_GLOBAL_DATA *)GetLoaderGlobalDataPointer ();
-  if (LdrGlobal->FspHobList == NULL) {
-    return;
-  }
-
-  HandOffHob = (EFI_HOB_HANDOFF_INFO_TABLE  *) LdrGlobal->FspHobList;
-  Length     = (UINT32)((UINTN)HandOffHob->EfiEndOfHobList - (UINTN)HandOffHob);
-  ZeroMem (HandOffHob, Length);
-  LdrGlobal->FspHobList = NULL;
-}
-
-/**
   Add a Smbios type string into a buffer
 
 **/
@@ -598,7 +574,7 @@ InitializeSmbiosInfo (
   Index         = 0;
   PlatformId    = GetPlatformId ();
   TempSmbiosStrTbl  = (SMBIOS_TYPE_STRINGS *) AllocateTemporaryMemory (0);
-  VerInfoTbl    = GetLoaderGlobalDataPointer()->VerInfoPtr;
+  VerInfoTbl    = GetVerInfoPtr ();
 
   //
   // SMBIOS_TYPE_BIOS_INFORMATION
@@ -971,7 +947,7 @@ BoardInit (
   BL_SW_SMI_INFO            *BlSwSmiInfo;
   VTD_INFO                  *VtdInfo;
   EFI_PEI_GRAPHICS_INFO_HOB *FspGfxHob;
-  LOADER_GLOBAL_DATA        *LdrGlobal;
+  VOID                      *FspHobList;
 
   switch (InitPhase) {
   case PreSiliconInit:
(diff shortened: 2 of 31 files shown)


BootloaderCorePkg/Library/BootloaderCoreLib/BootloaderCoreLib.c
```diff
@@ -444,3 +444,72 @@ GetTempRamInfo (
...
+  @param Features   The features the platform supports.
+
+**/
+VOID
+EFIAPI
+SetFeatureCfg (
+  IN  UINT32  Features
+  )
+{
+  GetLoaderGlobalDataPointer()->LdrFeatures = Features;
```
rprangar, line 475: Better to OR a specific feature than assignment?

rprangar (approved):
Looks good to me

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -2128,7 +2106,11 @@ UpdateCpuNvs (
   ///
   /// Get CPU Init Data Hob
   ///
-  GuidHob = GetNextGuidHob (&gCpuInitDataHobGuid, LdrGlobal->FspHobList);
+  GuidHob = NULL;
+  FspHobList = GetFspHobListPtr ();
+  if (FspHobList != NULL) {
+    GuidHob = GetNextGuidHob (&gCpuInitDataHobGuid, FspHobList);
+  }
```
gdong1, line 2113: How about update the GetNextGuidHob() to return NULL when the second parameter is NULL.
so we could still use GetNextGuidHob() in this case.

Silicon/TigerlakePkg/Library/Stage1BSocInitLib/Stage1BSocInitLib.c
```diff
@@ -79,18 +79,21 @@ IsWarmReset (
...
-  LOADER_GLOBAL_DATA    *LdrGlobal;
+  VOID                  *FspHobList;
   MEMORY_PLATFORM_DATA  *MemPlatformData;
   UINT8                 ResetType;
   BOOLEAN               IsWarmReset;
 
   IsWarmReset     = FALSE;
-  LdrGlobal       = (LOADER_GLOBAL_DATA *)GetLoaderGlobalDataPointer ();
-  MemPlatformData = GetGuidHobData (
-                      LdrGlobal->FspHobList,
```
gdong1, line 90: How about changing LdrGlobal->FspHobList to GetFspHobListPtr () directly?

BootloaderCorePkg/Library/BootloaderCoreLib/BootloaderCoreLib.c
gdong1: We already have GetFeatureCfg(). May add a new API FeatureCfgAndThenOr (UINT32  AndData,  UINT32  OrData)?
Agree it could simplify the caller.

myguitar: This is to make CorePkg APIs simple with primitive Get()/Set() API pair. Using Get()/Set() will minimize changes from original code. Let me keep this implementation to avoid unexpected issues with AndThenOr(). If AndThenOr() is required, we can add it later.

```
From:
  LdrGlobal = (LOADER_GLOBAL_DATA *)GetLoaderGlobalDataPointer ();
  LdrGlobal->LdrFeatures |= FeaturePcdGet (PcdAcpiEnabled)?FEATURE_ACPI:0;
  LdrGlobal->LdrFeatures |= FeaturePcdGet (PcdVerifiedBootEnabled)?FEATURE_VERIFIED_BOOT:0;
  LdrGlobal->LdrFeatures |= FeaturePcdGet (PcdMeasuredBootEnabled)?FEATURE_MEASURED_BOOT:0;
  LdrGlobal->LdrFeatures &= ~FEATURE_ACPI;
  LdrGlobal->LdrFeatures &= ~FEATURE_MEASURED_BOOT;
  LdrGlobal->LdrFeatures &= ~FEATURE_VERIFIED_BOOT;
To:
  LdrFeatures  = GetFeatureCfg ();
  LdrFeatures |= FeaturePcdGet (PcdAcpiEnabled)?FEATURE_ACPI:0;
  LdrFeatures |= FeaturePcdGet (PcdVerifiedBootEnabled)?FEATURE_VERIFIED_BOOT:0;
  LdrFeatures |= FeaturePcdGet (PcdMeasuredBootEnabled)?FEATURE_MEASURED_BOOT:0;
  LdrFeatures &= ~FEATURE_ACPI;
  LdrFeatures &= ~FEATURE_MEASURED_BOOT;
  LdrFeatures &= ~FEATURE_VERIFIED_BOOT;
  SetFeatureCfg (LdrFeatures);
```

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
myguitar: I thought about it, but this is to keep EDKII interface as it is.

  **If Guid is NULL, then ASSERT().
  If HobStart is NULL, then ASSERT().**

```
In MdePkg\Include\Library\HobLib.h

/**
  Returns the next instance of the matched GUID HOB from the starting HOB.

  This function searches the first instance of a HOB from the starting HOB pointer.
  Such HOB should satisfy two conditions:
  its HOB type is EFI_HOB_TYPE_GUID_EXTENSION and its GUID Name equals to the input Guid.
  If there does not exist such HOB from the starting HOB pointer, it will return NULL.
  Caller is required to apply GET_GUID_HOB_DATA () and GET_GUID_HOB_DATA_SIZE ()
  to extract the data section and its size info respectively.
  In contrast with macro GET_NEXT_HOB(), this function does not skip the starting HOB pointer
  unconditionally: it returns HobStart back if HobStart itself meets the requirement;
  caller is required to use GET_NEXT_HOB() if it wishes to skip current HobStart.

  If Guid is NULL, then ASSERT().
  If HobStart is NULL, then ASSERT().

  @param  Guid          The GUID to match with in the HOB list.
  @param  HobStart      A pointer to a Guid.

  @return The next instance of the matched GUID HOB from the starting HOB.

**/
VOID *
EFIAPI
GetNextGuidHob (
  IN CONST EFI_GUID         *Guid,
  IN CONST VOID             *HobStart
  );
```

Silicon/TigerlakePkg/Library/Stage1BSocInitLib/Stage1BSocInitLib.c
myguitar: `LdrGlobal->FspHobList` is the removed code. This patch is using GetFspHobListPtr().

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
gdong1: Yes, EDKII will ASSERT when the second parameter is NULL since it is not expected.
Comparing with new code, did you see issue with below code?  The FspHob should not be NULL else it should be bug.
GuidHob = GetNextGuidHob (&gCpuInitDataHobGuid, GetFspHobListPtr ());

Silicon/TigerlakePkg/Library/Stage1BSocInitLib/Stage1BSocInitLib.c
gdong1: I mean instead of using new code from line 88 ~ 96, how about we update removed code to only change 1 line.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
myguitar: GetFspHobListPtr () is NULL until MemoryInit is done. Caller MUST guarantee that GetNextGuidHob() parameters are not NULL.

Silicon/TigerlakePkg/Library/Stage1BSocInitLib/Stage1BSocInitLib.c
myguitar: LdrGlobal->FspHobList or GetFspHobListPtr () can be NULL until MemoryInit is done. If the 1st parameter of GetGuidHobData() is NULL, GetGuidHobData will use SblHob instead of FspHob. Better to check if GetFspHobListPtr() is NULL by Caller.

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
gdong1: MemoryInit must be done in stage2 code.  Prefer to use previous logic to make code simple since this patch mainly changes how to get FSP HOB pointer (from global data or from a function), not change FSP HOB itself.
But I am ok if you insist to have a check every time use FSP HOB pointer.

gdong1 (approved):

Platform/CoffeelakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
myguitar: MemoryInit is done in Stage1B. Refer to Stage1B.c line#442. Until the line, FspHobList is NULL.
This patch added a safe code to make sure NOT NULL. If we change GetNextGuidHob(), caller codes would be simpler but we need to change MdePkg\Include\Library\HobLib.h interface description/definition.

gdong1: It could not be NULL in this code Stage2BoardInitLib.c, why we newly add the check for FspHobList. That's my concern.

myguitar: Right, it shouldn't be NULL in Stage2 until ClearFspHob(). This is to add a safe code since we have a chance to have NULL FspHobList.

gdong1: From my viewpoint, it is not necessary to add such check: a) Simplify the caller. b) Capture such logic error early by assertion (If it asserts, the code logic much be fixed why we get FSP hob while FspHobList is NULL.)

myguitar: Your first comment was to make GetNextGuidHob() return NULL when the second parameter is NULL instead of ASSERTION. Now, it's back to ASSERTION.

gdong1: At that time I had thought GetNextGuidHob() would hang if NULL is passed to it when I saw the new check was added.

gdong1: Let's this big patch firstly since adding additional check doesn't harm anything. we could change it back later if required.

PR #1250, opened 2021-08-05
[TGL] add control for TCSS PTM Enable

stanleyintel:
According to EDS, TCSS PTM Enable (PTME on [B0,D7,F0] offset 158h)
should not be set unless a associated downstream port already has
PTM Enable set.

This patch adds a CfgData control for each TCSS PCIe Root Port.
User can enable each of them when a downstream port meets the
requirement. The new CfgData control is similiar to the following
setup in UEFI BIOS Menu:

    Intel Advanced -> SA -> TCSS -> PCIE RP[]

Last, this patch also removes redundant PTM configurations,
because (a) PTM settings will be overridden by built-in
CfgData_Silicon.yaml or customized DLT; b) even for customized 
SBL, FSP will use default PTM settings.

### DIFF
--- a/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -1418,14 +1418,6 @@ UpdateFspConfig (
   // Enable IEH
   FspsConfig->IehMode = 0x1;
 
-  // PTM configuration
-  for (Index = 0; Index < GetPchMaxPciePortNum(); Index++) {
-    FspsConfig->PciePtm[Index] = 0x1;
-  }
-  for (Index = 0; Index < 4; Index++) {
-    FspsConfig->CpuPcieRpPtmEnabled[Index] = 0x1;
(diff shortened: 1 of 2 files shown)


myguitar (approved):
Looks good to me

Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1424,7 +1424,7 @@ UpdateFspConfig (
   }
   for (Index = 0; Index < 4; Index++) {
     FspsConfig->CpuPcieRpPtmEnabled[Index] = 0x1;
-    FspsConfig->PtmEnabled[Index] = 0x1;
+    FspsConfig->PtmEnabled[Index] = 0x0;
   }
```
gdong1, line 1428: it looks line 1425 ~ 1428 could be removed since both will be set again in this function.

stanleyintel: noted and updated.

gdong1 (approved):

PR #1261, opened 2021-08-11
[EHL] Fix StitchIfwiConfig build issue on missing IPFW component

LeanSheng:
During stitching, the error shown when any IPFW component is not
found in SBL binary while the component is included in stitching
IPFW folder. This CL add one more check to make sure that the
specific component can be also found in SBL binary IPFW region.

Signed-off-by: Lean Sheng Tan <lean.sheng.tan@intel.com>

### DIFF
--- a/Platform/ElkhartlakeBoardPkg/Script/StitchIfwiConfig.py
+++ b/Platform/ElkhartlakeBoardPkg/Script/StitchIfwiConfig.py
@@ -79,32 +79,32 @@ def get_component_replace_list():
       ('IFWI/BIOS/TS1/ACM0',      'Input/acm.bin',        'dummy',    '',                            ''),
     ]
 
-    if os.path.exists('IPFW/PreOsChecker.bin'):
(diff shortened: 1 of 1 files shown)


Platform/ElkhartlakeBoardPkg/Script/StitchIfwiConfig.py
```diff
@@ -79,32 +79,32 @@ def get_component_replace_list():
       ('IFWI/BIOS/TS1/ACM0',      'Input/acm.bin',        'dummy',    '',                            ''),
     ]
 
-    if os.path.exists('IPFW/PreOsChecker.bin'):
+    if os.path.exists('IPFW/PreOsChecker.bin') and os.path.exists('IFWI/BIOS/NRD/IPFW/POSC'):
```
gdong1, line 82: "IFWI/BIOS/NRD/IPFW/POSC" is not really file system file. it is the component path from flash path.

PR #1271, opened 2021-08-24
[CML] enable stitch tools for Linux

stanleyintel:
The patch enables stitching CML/CMLV IFWI under Linux.

  CML/CMLV stitch tools do not natively support for Linux.
  A wrapper using 'wine' is required and works.


myguitar (approved):
Looks good to me

gdong1 (approved):

PR #1278, opened 2021-08-30
[TGL] fix fw update hang due to rom size change

stanleyintel:
This patch fixes a hang issue during fw update caused by
mismatching bios rom size. For a fw update, the TopSwap
size, Redundant Region size, and total BIOS region must be
identical.

This patch also adds check on rom size during fw update.
The check only works for a running SBL previously built with
this patch.

### DIFF
--- a/Platform/TigerlakeBoardPkg/BoardConfig.py
+++ b/Platform/TigerlakeBoardPkg/BoardConfig.py
@@ -170,9 +170,19 @@ class Board(BaseBoard):
         Redundant_Components_Size = self.UCODE_SIZE + self.STAGE2_SIZE + self.STAGE1B_SIZE + self.FWUPDATE_SIZE + self.CFGDATA_SIZE + self.KEYHASH_SIZE
         if Redundant_Components_Size > self.REDUNDANT_SIZE:
             raise Exception ('Redundant region size 0x%x is smaller than required components size 0x%x!' % (self.REDUNDANT_SIZE, Redundant_Components_Size))
-        self.NON_REDUNDANT_SIZE   = 0x2B2000 + self.SIIPFW_SIZE
         self.NON_VOLATILE_SIZE    = 0x001000
-        self.SLIMBOOTLOADER_SIZE  = (self.TOP_SWAP_SIZE + self.REDUNDANT_SIZE) * 2 + self.NON_REDUNDANT_SIZE + self.NON_VOLATILE_SIZE
+        # For firmware update, please keep SLIMBOOTLOADER_SIZE unchanged!
+        # The info can be found in the 'RomSize' of Outputs/tgl/FlashMap.txt
+        # Max size for 16MB IFWI: 0xAD8000
+        # Default value in UEFI BIOS (32MB IFWI): 0xC00000
+        self.SLIMBOOTLOADER_SIZE  = 0xAD0000
+        self.NON_REDUNDANT_SIZE   = self.SLIMBOOTLOADER_SIZE - \
+                                    (self.TOP_SWAP_SIZE + self.REDUNDANT_SIZE) * 2 - \
+                                    self.NON_VOLATILE_SIZE
+        Non_Redundant_Components_Size = self.PAYLOAD_SIZE + self.EPAYLOAD_SIZE + self.MRCDATA_SIZE + \
+                                        self.UEFI_VARIABLE_SIZE + self.VARIABLE_SIZE + self.SIIPFW_SIZE
+        if self.NON_REDUNDANT_SIZE < Non_Redundant_Components_Size:
+            raise Exception ('Non redundant region size 0x%x is smaller than required components size 0x%x!' % (self.NON_REDUNDANT_SIZE, Non_Redundant_Components_Size))
 
         self.PLD_HEAP_SIZE        = 0x04000000
         self.PLD_STACK_SIZE       = 0x00020000
--- a/Silicon/TigerlakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
+++ b/Silicon/TigerlakePkg/Library/FirmwareUpdateLib/FirmwareUpdateLib.c
@@ -1,7 +1,7 @@
 /** @file
   This file contains the implementation of FirmwareUpdateLib library.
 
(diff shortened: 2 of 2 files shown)


stanleyintel:
Some checks fail because https://www.nasm.us/pub/nasm/releasebuilds/ is not reachable at this moment. Will force push to trigger another round of check once the nasm site is back.

Platform/TigerlakeBoardPkg/BoardConfig.py
```diff
@@ -170,9 +170,16 @@ def __init__(self, *args, **kwargs):
...
-        self.SLIMBOOTLOADER_SIZE  = (self.TOP_SWAP_SIZE + self.REDUNDANT_SIZE) * 2 + self.NON_REDUNDANT_SIZE + self.NON_VOLATILE_SIZE
+        # For firmware update, please keep SLIMBOOTLOADER_SIZE unchanged!
+        # The info can be found in Outputs/tgl/FlashMap.txt
+        self.SLIMBOOTLOADER_SIZE  = 0xAB0000
+        self.NON_REDUNDANT_SIZE   = self.SLIMBOOTLOADER_SIZE - \
+                                    (self.TOP_SWAP_SIZE + self.REDUNDANT_SIZE) * 2 - \
+                                    self.NON_VOLATILE_SIZE
+        Non_Redundant_Components_Size = 0x2B2000 + self.SIIPFW_SIZE
+        if self.NON_REDUNDANT_SIZE < Non_Redundant_Components_Size:
+            raise Exception ('Non redundant region size 0x%x is smaller than required components size 0x%x!' % (self.NON_REDUNDANT_SIZE, Non_Redundant_Components_Size))
```
gdong1, line 182: had better calculate Non_Redundant_Components_Size as below?
Non_Redundant_Components_Size =  self.SIIPFW_SIZE + self.VARIABLE_SIZE + self.UEFI_VARIABLE_SIZE + self.MRCDATA_SIZE + self.EPAYLOAD_SIZE + self.PAYLOAD_SIZE

stanleyintel: thanks. will be verified and updated.

Platform/TigerlakeBoardPkg/BoardConfig.py
```diff
@@ -170,9 +170,17 @@ def __init__(self, *args, **kwargs):
         Redundant_Components_Size = self.UCODE_SIZE + self.STAGE2_SIZE + self.STAGE1B_SIZE + self.FWUPDATE_SIZE + self.CFGDATA_SIZE + self.KEYHASH_SIZE
         if Redundant_Components_Size > self.REDUNDANT_SIZE:
             raise Exception ('Redundant region size 0x%x is smaller than required components size 0x%x!' % (self.REDUNDANT_SIZE, Redundant_Components_Size))
-        self.NON_REDUNDANT_SIZE   = 0x2B2000 + self.SIIPFW_SIZE
         self.NON_VOLATILE_SIZE    = 0x001000
-        self.SLIMBOOTLOADER_SIZE  = (self.TOP_SWAP_SIZE + self.REDUNDANT_SIZE) * 2 + self.NON_REDUNDANT_SIZE + self.NON_VOLATILE_SIZE
+        # For firmware update, please keep SLIMBOOTLOADER_SIZE unchanged!
+        # The info can be found in the 'RomSize' of Outputs/tgl/FlashMap.txt
+        self.SLIMBOOTLOADER_SIZE  = 0xAB0000
```
myguitar, line 176: I am wondering why SBL size is 10.xx MB. Any specific reason? I think it would be good to bed aligned with MB size. What about 0xB00000?

stanleyintel: Thanks for the question first. 
Currently, the TGL FIT can support up to 0xAD8000. Otherwise, it will stop and report "Error 25: [Csme Binary Gen] Buffer offset out of bounds". 
Here, not setting to max size (0xAD8000) is to left some room for any unexpected FIT change in the future. Especially we know, any change of SLIMBOOTLOADER_SIZE will result in failed FWU. Therefore, if this value is changed again in the future it will cause large impact to existing products. Conservatively, 0xAB0000 might be a better choice.

mauricema: I think it might be better to align the BIOS region size with CRB BIOS.   CRB BIOS region is 0xC00000.  As you said, it is not preferred to change the  BIOS region size across FWU flow.  So we should set the BIOS region size big enough to accommodate future usage, including OEM/ODM usage.

stanleyintel: Agree. But we cannot set it to any value larger than 0xAD8000; otherwise, FIT reports error 25.

Anyway, I will try with the latest FIT tomorrow, and set it to the largest acceptable value.

mauricema: If so, I consider it as a bug in our stitching process.   I did not see any reason why SBL cannot set to 0xC00000 while BIOS can.  Thank you for looking into it.

stanleyintel: Found the reason: it is because 16MB IFWI and 32MB IFWI. When stitching with option "32MB", FIT indeed can support up to 0x1AD8000.

Patch amended with additional comment for 16MB IFWI and Default UEFI BIOS

stanleyintel: Hi @mauricema 
Two questions are: 
  1. Current internal release.py only build for 16MB IFWI, should we change it to 32MB, or left it as 16MB?
  2. If we keep the internal SBL IFWI 16MB, should we set SLIMBOOTLOADER_SIZE to 0xAD8000 (full occupy), or reserve some margin like 0xAB0000 or 0xAD0000 ?

mauricema: That makes sense.  Thank you for providing the info.

@gdong1 ,  please comment on 16MB IFWI vs 32MB IFWI.  

For now, let us just keep it as 16MB and use 0xAD0000.  If we decide to change to 32MB IFWI later on, then we can submit separate patch to update BIOS region to 0xC00000.

mauricema (approved):
Looks good to me.

myguitar (approved):

PR #1279, opened 2021-08-31
Enable Oem key has generation for Btg0 profile

raghavag:
Latest BpmGen2 require Oem Key hash generation for
Btg legacy profile also. This patch enabled oem key generation
for Btg0 profile.


lsubash (approved):
Looks good

PR #1280, opened 2021-08-31
Resolve KW issue in boot command

jgutbub:
Boot command may attempt to use an index which
exceeds the max image type. Check if the value
is larger than the max allowed value and return
if there is an error.


raghavag (approved):
Looks good to me

PR #1281, opened 2021-09-06
Move BdatLib to CommonSocPkg

stalamudupula:
This patch adds BdatLib to CommonSocPkg so that all projects
can refer to one single instance of BdatLib. Also removed the
redundant platform-specific package folders.


myguitar (approved):

mauricema (approved):
Looks good.

PR #1282, opened 2021-09-08
[TGL] Add Fusa Enable CFG option

jgutbub:
Not every board will want to boot with
Fusa Enabled (ex. UPX i11 board). We can
allow for Fusa to be skipped using CFG
instead of just BoardConfig flag.

### DIFF
--- a/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -858,6 +858,7 @@ BoardInit (
   EFI_PEI_GRAPHICS_INFO_HOB *FspGfxHob;
   VOID                      *FspHobList;
   SILICON_CFG_DATA          *SiCfgData;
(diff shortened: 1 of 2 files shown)


Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -928,8 +929,11 @@ BoardInit (
       }
     }
     BuildOsConfigDataHob ();
-    if (FeaturePcdGet (PcdFusaEnabled)) {
-      FusaKeyOffProcessing();
+    FeaturesCfgData = (FEATURES_CFG_DATA *) FindConfigDataByTag(CDATA_FEATURES_TAG);
```
mauricema, line 932: can you put all these code under if (FeaturePcdGet (PcdFusaEnabled)) {} condition ?   In this way, if static build flag is off, all the related code will be excluded from compiling.

jgutbub: Done.

mauricema (approved):
Looks good.

PR #1284, opened 2021-09-10
[EHL] Adding option to select new gpio scheme

koktong-ong:
There is a discrepancy between how Linux reads GPIO and how bootloader
is presenting it. This causes Linux distros to crash, even in the installer,
unless a kernel module parameter has been passed:
'module_blacklist=pinctrl_elkhartlake'

The customer, Red Hat, is using RHEL 8 4.18.0-315.
There appears to be a discrepancy between how Linux reads GPIO and how
the Intel BIOS is presenting it.


Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Ext_IotgCrb.dlt
```diff
@@ -26,6 +26,7 @@ GPIO_CFG_DATA.GpioPinConfig1_GPP_D13.GPIOSkip_GPP_D13 | 0
 # Enable to test TCC mode & tuning
 # FEATURES_CFG_DATA.Features.Tcc    | 0x1
 FEATURES_CFG_DATA.Features.LowPowerIdle      | 0x1
+FEATURES_CFG_DATA.Features.NewGpioSchemeEnable      | 0x1
```
mauricema, line 29: Do you know if this change is compatible to all OS ?   EX: Does every POR OS move to the new GPIO scheme ?

PR #1285, opened 2021-09-10
[EHL] Adding option to select new gpio scheme

koktong-ong:
There is a discrepancy between how Linux reads GPIO and how bootloader
is presenting it. This causes Linux distros to crash, even in the installer,
unless a kernel module parameter has been passed:
'module_blacklist=pinctrl_elkhartlake'

The customer, Red Hat, is using RHEL 8 4.18.0-315.
There appears to be a discrepancy between how Linux reads GPIO and how
the Intel BIOS is presenting it.


mauricema (approved):
Looks good to me.

PR #1286, opened 2021-09-10
[TGL] Update UPX GFX and SMBIOS data

jgutbub:
For the UPX i11 board we need to set
the SA display table (same as DDR4 RVP),
populate the VBT callback routine (also
same DDR4 RVP), and update the SMBIOS
base board string.


jgutbub:
Need to double check the DP port, looks like maybe there is some difference on the TC port used versus RVP.

mauricema (approved):
Looks good.

mauricema:
> Need to double check the DP port, looks like maybe there is some difference on the TC port used versus RVP.

For UPX i11, the GPIO_A14 needs to be configured as native function 2,  DDSP_HPD3.   Current code is set to GPIO output.

myguitar (approved):
Looks good to me

myguitar:
Still need to be in Draft PR?

jgutbub:
Merging for now, DP port is not initializing during pre-OS however, need to complete in a future patch.

PR #1287, opened 2021-09-13
Fix build issue for UCODE repo clone

mauricema:
Ucode repo can be mapped to several different repos in SBL.
Current SBL uses the exactly same folder name for all UCODE
repo. It causes build issues when the UCODE repo switches
from one repo to another. To avoid this issue, this patch
allows to automatically use different folder name for UCODE
repo clone so that each UCODE repo has its own folder.


myguitar (approved):
Looks good to me

PR #1292, opened 2021-09-16
[TGL] Increase CFG data mempool and skip debug UART init

jgutbub:
With the recent additions to the CFG data (new DLT for
UPXi11 and new fields added) the CFG data size alloced
in memory seems to have run out of space, need to add
space to load the CFG data. Also, we need to specify
in FSP-M to skip UART init for the debug port we are
using when it is a PCH UART or we lose debug messages
in Stage2 onward if using a debug FSP.


mauricema (approved):
Looks good to me.
Added Guo to review as well.

gdong1 (approved):

PR #1296, opened 2021-09-17
[TGL] Group GPIO pin configurations

mauricema:
There are too many pins in GPIO pin configurations in TGL platform.
It is desired to group them into sub-pages to make it easy to find
the proper pin for configurations.  This patch implemented this.


myguitar (approved):
Looks good to me

PR #1307, opened 2021-09-27
Resolve fs cmd KW issue

jgutbub:
Currently the PLT_DEVICE_TABLE is limited
to only using OS_BOOT_MEDIUM_TYPE devices
but ideally should be capable of mapping
any system device into the table. For the
time being we will remove the OsBootDeviceGraphics
since it created some KW issue in the fs command.


mauricema (approved):
Looks good to me.
Let us enhance it later on.

PR #1308, opened 2021-09-28
[EHL] Enable Gbe TSN config in yaml

koktong-ong:
Enable Gbe TSN config in silicon yaml file below:
- PchTsnGbeSgmiiEnable
- PseTsnGbeSgmiiEnable
- PseTsnGbePhyInterfaceType

### DIFF
--- a/Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
+++ b/Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
@@ -1758,7 +1758,7 @@
       help         : >
                      Enable/Disable #AC check on split lock. <b>0- Disable</b>; 1- Enable.
       length       : 0x1
-      value        : 0x1
+      value        : 0x0
   - SgxSinitNvsData :
       name         : SgxSinitNvsData
       type         : EditNum, HEX, (0x0,0xFF)
@@ -1867,11 +1867,49 @@
                      When FALSE, it disables Timed GPIO 1.
       length       : 1b
       value        : 0x0
+  - PchTsnGbeSgmiiEnable :
+      name         : PCH TSN SGMII Support
+      type         : Combo
+      option       : $EN_DIS
+      help         : >
+                     Enable/disable PCH TSN SGMII support 0=Disable 1=Enable
+      length       : 1b
+      value        : 0x1
+  - PseTsnGbe0SgmiiEnable :
+      name         : PSE TSN 0 SGMII Support
+      type         : Combo
+      option       : $EN_DIS
+      help         : >
+                     Enable/disable PSE TSN 0 SGMII support 0=Disable 1=Enable
+      length       : 1b
+      value        : 0x00
+  - PseTsnGbe1SgmiiEnable :
+      name         : PSE TSN 1 SGMII Support
+      type         : Combo
+      option       : $EN_DIS
+      help         : >
+                     Enable/disable PSE TSN 1 SGMII support 0=Disable 1=Enable
+      length       : 1b
+      value        : 0x00
+  - PseTsnGbe0PhyInterfaceType :
+      name         : PSE TSN 0 Phy Interface Type
+      type         : EditNum, HEX, (0x00,0x3)
+      help         : >
+                     Set PSE TSN 0 Phy Interface Type 0: Not Connected, 1: RGMII, 2: SGMII, 3:SGMII+
+      length       : 2b
+      value        : 0x01
+  - PseTsnGbe1PhyInterfaceType :
+      name         : PSE TSN 1 Phy Interface Type
+      type         : EditNum, HEX, (0x00,0x3)
+      help         : >
+                     Set PSE TSN 1 Phy Interface Type 0: Not Connected, 1: RGMII, 2: SGMII, 3:SGMII+
+      length       : 2b
+      value        : 0x01
   - Dummy        :
       name         : SaPostMemTestRsvd
       type         : Combo
       option       : $EN_DIS
       help         : >
                      Reserved for SA Post-Mem Test
-      length       : 22b
+      length       : 15b
       value        : 0x0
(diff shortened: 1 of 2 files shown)


Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
```diff
@@ -1857,21 +1857,43 @@
...
+      length       : 0x2
+      value        : { 0x01, 0x01 }
   - Dummy        :
       name         : SaPostMemTestRsvd
       type         : Combo
       option       : $EN_DIS
       help         : >
                      Reserved for SA Post-Mem Test
-      length       : 22b
+      length       : 0x4
```
mauricema, line 1898: Why do you need to pad dummy 4 bytes?   
Dummy is only used if the structure is not 4 bytes aligned.  So the length should be 1 to 3.  Otherwise, dummy should be removed.

Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
```diff
@@ -1857,21 +1857,43 @@
...
+      length       : 0x01
       value        : 0x0
   - EnableTimedGpio1 :
       name         : Timed GPIO 1
       type         : Combo
       option       : $EN_DIS
       help         : >
                      When FALSE, it disables Timed GPIO 1.
-      length       : 1b
+      length       : 0x01
```
mauricema, line 1868: Why do you want to use 1 byte  instead of 1 bit for ENABLE/DISABLE?

Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
```diff
@@ -1857,21 +1857,43 @@
...
       value        : 0x0
+  - PchTsnGbeSgmiiEnable :
+      name         : PCH TSN SGMII Support
+      type         : Combo
+      option       : $EN_DIS
+      help         : >
+                     Enable/disable PCH TSN SGMII support 0=Disable 1=Enable
+      length       : 0x01
```
mauricema, line 1876: Why do you want to use 1 byte  instead of 1 bit for ENABLE/DISABLE?

Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
```diff
@@ -1857,21 +1857,43 @@
...
+      value        : 0x1
+  - PseTsnGbeSgmiiEnable :
+      name         : PSE TSN SGMII Support
+      type         : EditNum, HEX, (0x00,0xFFFF)
+      help         : >
+                     Enable/disable PSE TSN SGMII support 0=Disable 1=Enable
+      length       : 0x2
```
mauricema, line 1883: Why do you want to use 2 byte  instead of 1 bit for ENABLE/DISABLE?

mauricema: Why EditNum type ?

Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1532,14 +1532,24 @@ UpdateFspConfig (
...
-    Fspscfg->PseTsnGbeSgmiiEnable[0] = 0;
-    Fspscfg->PseTsnGbeSgmiiEnable[1] = 0;
-    Fspscfg->PseTsnGbePhyInterfaceType[0]    = 1;
-    Fspscfg->PseTsnGbePhyInterfaceType[1]    = 1;
-
+    Fspscfg->PseTsnGbeSgmiiEnable[0] = SiCfgData->PseTsnGbeSgmiiEnable[0];
+    Fspscfg->PseTsnGbeSgmiiEnable[1] = SiCfgData->PseTsnGbeSgmiiEnable[1];
+    Fspscfg->PseTsnGbePhyInterfaceType[0]    = SiCfgData->PseTsnGbePhyInterfaceType[0];
+    Fspscfg->PseTsnGbePhyInterfaceType[1]    = SiCfgData->PseTsnGbePhyInterfaceType[1];
+    DEBUG ((DEBUG_INFO, "------------------------------------------PCH and PSE TSN CONFIG----------------------------------------\n"));
```
mauricema, line 1542: Consider to use DEBUG_VERBOSE type.  It is too detailed debug info.

Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
mauricema: if you have two interfaces, then it should be broken into 2 item and each is a single bit.

Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
koktong-ong: ok i misunderstood the 1b ad byte and 0x1 as bit. Will update the patch

Platform/ElkhartlakeBoardPkg/CfgData/CfgData_Silicon.yaml
koktong-ong: i saw there are some similar multiple interface cfgdata with enable/disable option are using EditNum with bytes as well. 
Example:
  - SataPortsEnable :
      name         : Enable SATA ports
      type         : EditNum, HEX, (0x00,0xFFFFFFFFFFFFFFFF)
      help         : >
                     Enable/disable SATA ports. One byte for each port, byte0 for port0, byte1 for port1, and so on.
      length       : 0x08
      value        : { 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}
Should we change them as well?

mauricema: Yes, but I think it is more user friendly if we break it down into individual config item.

koktong-ong:
Closed this and start new PR https://github.com/slimbootloader/slimbootloader/pull/1414

PR #1310, opened 2021-09-28
Add platform device type enum definitions

mauricema:
SBL defined SetDeviceAddr/GetDeviceAddr to abstract platform
device info. However, currently only boot media devices are
defined. It has cases that other platform device info is needed,
such as GFX device. This patch extended the concept to define
some other platform devices as an extenstion to
OS_BOOT_MEDIUM_TYPE. The enum values should not overlap with
the ones defined in OS_BOOT_MEDIUM_TYPE.


jgutbub (approved):
Looks good to me.

PR #1311, opened 2021-09-29
[EHL] Disable New GPIO scheme by default

koktong-ong:
Disable the new GPIO scheme by default in dlt file
Only enable this option for kernel version 4.18.0-315
Alternatively with kernel module parameter
'module_blacklist=pinctrl_elkhartlake' will works
without enabling this GPIO scheme.


mauricema (commented):
Please provide more info in commit message.
It should cover in which cases it is required to use new GPIO scheme.

mauricema (approved):
Looks good to me.

PR #1312, opened 2021-09-30
[TGL] Update FSP and platform version since MR3 is released

vincentchchen:
- UP3 IoT FSP MR3
- change the FSP headers from FspBin folder to Include folder
- update TGL platform version to 1.3

### DIFF
--- a/Silicon/TigerlakePkg/FspBin/FspBin.inf
+++ b/Silicon/TigerlakePkg/FspBin/FspBin.inf
@@ -11,16 +11,16 @@
 
 [UserExtensions.SBL."CloneRepo"]
   REPO    = https://github.com/intel/FSP.git
-  COMMIT  = 4a31a4d972eaeb9aa41fef3b121df97f8f2a4e61
+  COMMIT  = 91a6117b66fca4c465d8480cc17408ad588290cd
 
 [UserExtensions.SBL."CopyList"]
   TigerLakeFspBinPkg/TGL_IOT/Fsp.fd                    : Silicon/TigerlakePkg/FspBin/FspDbg.bin
   TigerLakeFspBinPkg/TGL_IOT/Fsp.fd                    : Silicon/TigerlakePkg/FspBin/FspRel.bin
   TigerLakeFspBinPkg/TGL_IOT/Fsp.bsf                   : Silicon/TigerlakePkg/FspBin/Fsp.bsf
-  TigerLakeFspBinPkg/TGL_IOT/Include/FspUpd.h          : Silicon/TigerlakePkg/FspBin/FspUpd.h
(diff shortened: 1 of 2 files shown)


Silicon/TigerlakePkg/FspBin/FspBin.inf
```diff
@@ -11,16 +11,16 @@
...
 
 [UserExtensions.SBL."CopyList"]
   TigerLakeFspBinPkg/TGL_IOT/Fsp.fd                    : Silicon/TigerlakePkg/FspBin/FspDbg.bin
   TigerLakeFspBinPkg/TGL_IOT/Fsp.fd                    : Silicon/TigerlakePkg/FspBin/FspRel.bin
   TigerLakeFspBinPkg/TGL_IOT/Fsp.bsf                   : Silicon/TigerlakePkg/FspBin/Fsp.bsf
-  TigerLakeFspBinPkg/TGL_IOT/Include/FspUpd.h          : Silicon/TigerlakePkg/FspBin/FspUpd.h
-  TigerLakeFspBinPkg/TGL_IOT/Include/FsptUpd.h         : Silicon/TigerlakePkg/FspBin/FsptUpd.h
-  TigerLakeFspBinPkg/TGL_IOT/Include/FspmUpd.h         : Silicon/TigerlakePkg/FspBin/FspmUpd.h
-  TigerLakeFspBinPkg/TGL_IOT/Include/FspsUpd.h         : Silicon/TigerlakePkg/FspBin/FspsUpd.h
+  TigerLakeFspBinPkg/TGL_IOT/Include/FspUpd.h          : Silicon/TigerlakePkg/Include/FspUpd.h
```
mauricema, line 20: Why do we change the header FSP file location from  FspBin to Include folder ?
The intention is to put it under FspBin since we might need to support multiple FSP bianry/headers in a single project.  You might have conflicts if they are located under Include folder.

vincentchchen: Just noticed that only this project copies the headers into FspBin. Didn't know the intention of this design.
Will discard the change. But would like to know what would be the future design to support multiple FSP, since so far TGLU and TGLH share the same FspBin.inf. And before they are combined, they used different folders to download FSP files.

mauricema: That makes sense.  If your project only supports one FSP binary, then it is fine to put it under "Include".  However, if your project needs to support multiple FSP binaries. then it needs to move to "FspBin".   
Since you only have one FSP binary,  I am OK with your original change.  Thank you for explanation!

mauricema (approved):
Looks good to me.

PR #1316, opened 2021-10-02
[QEMU] Fix AUTO boot option index

mauricema:
On SBL, it can support boot option selection through QEMU command line
"-boot order" parameter. However, it does not work anymore. It was
because of the MAX_BOOT_OPTION_CFGDATA_ENTRY adjustment in other commit.
This patch decoupled internal boot option index with the CFGDATA boot
option index so that it does not have impacts on each other. With this
change, QEMU boot option can be altered through command line again.


myguitar (approved):
Looks good to me

PR #1317, opened 2021-10-02
Enable UEFI Universal Payload support 

mauricema:
This patch series enabled UEFI Universal Payload support on SBL.


myguitar (approved):
Looks good to me

PR #1318, opened 2021-10-03
Add gitignore file in Silicon and Platform folder

mauricema:
Current SBL build will show lots of FSP files as untracked files.
This patch re-orgnized the rules for .gitignore file so that it
has better control at different folder level. With these new rules
no untracked files will be shown for the FSP and VBT files.


Silicon/.gitignore
```diff
@@ -0,0 +1,7 @@
+Fsp.bsf
+FspDbg.bin
+FspRel.bin
+Fsp*.h
+Fsp*.h
```
myguitar, line 5: Please remove duplicated one.

myguitar (approved):
A minor comment on a duplicated one. Other than that, it looks good to me.

Silicon/.gitignore
mauricema: Sure, will do. thx!

mauricema: Fixed it.

PR #1319, opened 2021-10-03
Add accurate TSC frequency calculation support

mauricema:
Current SBL code uses MSR(0xce) to calculate the CPU TSC frequency.
However, it is not very accurate. A better way is to use CPUID to
calculate the TSC frequency. This patch added new API to get accurate
TSC frequency.  It also added APIs to allow conversion between time
and TSC ticks.

### DIFF
--- a/BootloaderCommonPkg/Library/TimeStampLib/TimeStampLib.c
+++ b/BootloaderCommonPkg/Library/TimeStampLib/TimeStampLib.c
@@ -1,10 +1,11 @@
 /** @file
 
-  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2021, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
+#include <Register/Intel/Cpuid.h>
 #include <Library/BaseLib.h>
 
 /**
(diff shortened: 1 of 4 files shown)


BootloaderCommonPkg/Library/TimeStampLib/TimeStampLib.c
```diff
@@ -44,3 +45,85 @@ GetTimeStampFrequency (
...
+  UINT32                 RegEbx;
+  UINT32                 RegEcx;
+
+  // Use CPUID leaf 0x15 Time Stamp Counter and Nominal Core Crystal Clock Information
+  // EBX returns 0 if not supported. ECX, if non zero, provides Core Xtal Frequency in hertz.
+  // TSC frequency = (ECX, Core Xtal Frequency) * EBX/EAX.
+  AsmCpuid (CPUID_TIME_STAMP_COUNTER, &RegEax, &RegEbx, &RegEcx, NULL);
+
+  // If EAX or EBX returns 0, the XTAL ratio is not enumerated.
+  if ((RegEax == 0) || (RegEbx == 0 )) {
```
myguitar, line 75: According to above comment, ECX could be also 0. The case may need to fallback to GetTimeStampFrequency as well.

mauricema: Good point, will make the change.

mauricema: Fixed it in new commit.

myguitar (approved):
Looks good to me.

PR #1320, opened 2021-10-04
[TGL] Disable Intel HD Audio (Azalia)

randylintw:
1. HD Audio and TSN share pins. These are mutually exclusive features.
 2. RVP board should be reworked to support legacy HD Audio mode.


Platform/TigerlakeBoardPkg/CfgData/CfgData_Silicon.yaml
```diff
@@ -278,7 +278,7 @@
       help         : >
                      0- Disable, 1- Enable (Default) Azalia controller
       length       : 0x01
-      value        : 0x01
+      value        : 0x00
```
mauricema, line 281: Is the intention to disable PchHdaEnable by default for all boards ?  
Or we just want to disable this for certain board ?  
If it is for specific  board,  we can do it in DLT file.

randylintw: Okey, disable TGL RVP board by default.

mauricema (approved):
Looks good to me.

PR #1322, opened 2021-10-05
[CFL] Increase PAYLOAD size

jgutbub:
Seems some recent patches may have added additional
size to OS Loader and now the internal pre-commit build
test in Jenkins is failing due to not enough space
for PAYLOAD. Other platforms are also using 29000h
so might be best to increase the CFL value as well.


mauricema (approved):
Looks good to me.

PR #1338, opened 2021-10-12
Enable FSP-V support

jinjhuli:
This patch enable FSP-V support in common code, include
adding FSP-V header, region, offset etc. It needs to pair
with FSP-V calling from each platform code base to run
FSP-V support.

Signed-off-by: jinjhuli <jin.jhu.lim@intel.com>


jinjhuli:
> FSP-V is not in any of the public EAS specification yet. I don't feel at this time it can be added into SBL core code. Please consider putting it into your own branch. If later on, the spec is published, then we can add the support easily.

hi Maurice, sure. Let me close this for now and re-submit again after we have the spec.

PR #1383, opened 2021-10-31
Fix GPIO CFGDATA database overriding issue

### LINKED ISSUE #1382: GPIO CFGDATA database was modified by GPIO programming code
On SBL, the GPIO CFGDATA database was modified by GPIO programming code in 
Silicon\CommonSocPkg\Library\GpioLib\GpioInit.c function FillGpioTable ().   It is not expected.

mauricema:
Current GPIO programming code in CommonSocPkg will override the
CFGDATA database. It tried to zero out the GPIO group ID and pad
ID from the GPIO CFGDATA entry. But it should only be applied to
the copied new entry, not the original entry in CFGDATA database.
This patch fixed this isue.
It fixed #1382.

### DIFF
--- a/Silicon/CommonSocPkg/Library/GpioLib/GpioInit.c
+++ b/Silicon/CommonSocPkg/Library/GpioLib/GpioInit.c
@@ -610,25 +610,21 @@ FillGpioTable (
 
 )
 {
-  UINT32             *GpioItem;
-  GPIO_PAD            GpioPad;
-  UINT8              *TableData;
+  GPIO_PAD           *GpioPad;
+  UINT8              *GpioData;
+  UINT8              *GpioCfg;
 
-  TableData = ((UINT8 *)GpioCfgHdr) + GpioCfgHdr->HeaderSize;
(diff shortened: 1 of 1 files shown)


gdong1 (approved):

stalamudupula:
@mauricema Is there any use-case to use Gpio Cfg item after the Gpio Programming for that particular pin has been done? The original thought was we dont need group Id and Pad Id from Cfg data after Gpio programming is done.

stalamudupula (approved):
Looks good to me.

mauricema:
> @mauricema Is there any use-case to use Gpio Cfg item after the Gpio Programming for that particular pin has been done? The original thought was we dont need group Id and Pad Id from Cfg data after Gpio programming is done.

Yes,  we do.  In later stage, we will rebuild the CFGDATA for a board using the database in memory.  Please refer to BuildConfigData() in ConfigDataLib.c.   In general, we should not modify any of the CFGDATA after it has been created in memory.

PR #1385, opened 2021-11-01
Add API to set fixed display mode in VBT table

mauricema:
In certain condition, it is required to use fixed display mode. This
path added a new API SetVbtFixedMode() in BoardSupportLib to provide
such function. It can be used to request a fixed resolution at runtime
to run an application, such as setup screen.


Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -533,3 +536,49 @@ GetVbtAddress ()
...
+      ModeBlk = (BLOCK51_FIXED_MODE_SET *)(VbtBuf + Offset);
+      // BlockId 51 is for fixed mode set
+      if (ModeBlk->BlockId == 51) {
+        ModeBlk->FeatureEnable = 1;
+        ModeBlk->XRes = Xres;
+        ModeBlk->YRes = Yres;
+        Status = EFI_SUCCESS;
+        break;
+      }
+      Offset += (ModeBlk->BlockSize + 3);
```
gdong1, line 580: It is difficult to understand the hard-coded number 3. could we use a meaning MACRO or calculation for it?

mauricema: sure, updated the code in the new commit.

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -533,3 +536,51 @@ GetVbtAddress ()
...
+      // BlockId 51 is for fixed mode set
+      if (BlkHdr->BlockId == 51) {
+        ModeBlk = (BLOCK51_FIXED_MODE_SET *) BlkHdr;
+        ModeBlk->FeatureEnable = 1;
+        ModeBlk->XRes = Xres;
+        ModeBlk->YRes = Yres;
+        Status = EFI_SUCCESS;
+        break;
+      }
+      Offset += (BlkHdr->BlockSize + sizeof(VBT_BLOCK_COMMON_HEADER));
```
gdong1, line 582: If VBT_BLOCK_COMMON_HEADER is added to Offset here, in the loop line 572 it is confused (VbtBuf + Offset) is still a pointer of VBT_BLOCK_COMMON_HEADER. Could you please double check?

mauricema: It points to next block header.  What is the problem here?

gdong1: if (VbtBuf + Offset) point to a block header in Line 572, then "Offset += BlkHdr->BlockSize" should point to next block header. here why it add sizeof(VBT_BLOCK_COMMON_HEADER)?

mauricema: Per spec, the BlockSize indicated in the header does not include the VBT_BLOCK_COMMON_HEADER itself.

gdong1: Got it.

gdong1 (approved):

PR #1386, opened 2021-11-01
[TGL] Gpio fix to allow unified group ID for TGL-U and TGL-H

mauricema:
This patch decouples the GPIO group ID used by GPIO library and the GPIO used by CFGDATA.


stalamudupula (approved):
Thanks for this enhancement. Looks good to me.

japelner-intc (approved):
Thanks for submitting this. This looks great.

PR #1388, opened 2021-11-02
Add capsule image location info

gdong1:
It is useful to have capsule image location info since
capsule image could be configured to different places.


mauricema (approved):
Looks good to me.

PR #1393, opened 2021-11-03
Add RxRaw field to Gpio config template

stalamudupula:
Current GpioLib uses 2 bits from OtherSettings to
configure RxRaw field in GPIO PAD CFG DWORD 0. But
Gpio config templates are missing the option to configure
this feature. This patch adds the option in template.

### DIFF
--- a/Platform/CometlakeBoardPkg/CfgData/Template_CfgData.yaml
+++ b/Platform/CometlakeBoardPkg/CfgData/Template_CfgData.yaml
@@ -381,14 +381,20 @@ GPIO_TMPL: >
         condition    : $GPIO_CFG_DATA.GpioPinConfig1_$(1).GPIOSkip_$(1) == 0
 
         length       : 4b
-
+    - RxRaw_$(1) :
+        name         : RxRaw
+        type         : Combo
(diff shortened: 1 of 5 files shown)


Platform/CometlakeBoardPkg/CfgData/Template_CfgData.yaml
```diff
@@ -381,14 +381,20 @@ GPIO_TMPL: >
         condition    : $GPIO_CFG_DATA.GpioPinConfig1_$(1).GPIOSkip_$(1) == 0
 
         length       : 4b
-
+    - RxRaw_$(1) :
+        name         : RxRaw
+        type         : Combo
+        option       : 0x0:Default, 0x2:NoOverride, 0x3:RxDrive_1_internally
```
mauricema, line 387: 1 for NoOverride ?

stalamudupula: Thanks for catching this. Updated in new patch.

mauricema (approved):
Looks good to me.  Thank you for the change.

PR #1398, opened 2021-11-05
Fix variable not initialized KW issue

myguitar:
This is to initialize a Boolean variable to fix KW issue


mauricema (approved):
Thank you for the fix, Aiden!

myguitar:
> Thank you for the fix, Aiden!

No problem! Thank you.

Issue #1396, opened 2021-11-04
Buffer overrun triggered by filesystem size


mehmetakbulut:
I have a disk image with a MBR table that contains a single partition with an EXT2 filesystem. The image file is always ~200MB. If the filesystem is <=1525MB in size, then everything works as expected. If the filesystem is >=1530MB in size, then I get the following error:
```
======== Try Booting with Boot Option 1 ========
BootMediumPciBase(0x1F02)
Getting boot image from SATA
Init AHCI controller E00FA000
AHCI port [0] has a [harddisk]
AHCI port [2] has a [cdrom]
Try to find boot partition
AhciDeviceRead Status = Device Error
Partition type: MBR  (1 logical partitions)
Find partition success
Init File system
Detected EXT on StartBlock 1 Part 0
BootSlot = 0x0
ASSERT [OsLoader] /home/mehmet.akbulut/code/av-sbl/BootloaderCommonPkg/Library/FullMemoryAllocationLib/Pool.c(546): Tail->Signature == ((('p') | ('t' << 8)) | ((('a') | ('l' << 8)) << 16))

STAGE_PAYLOAD: System halted!
```

This is built for and running on QEMU with:
`qemu-system-x86_64 -pflash ./SlimBootloader.bin -machine pc-q35-4.2,accel=kvm,usb=off,vmport=on,dump-guest-core=off -cpu qemu64 -m 8196 -smp 4 -hda ./disk.img --serial mon:stdio`

Following are the differences I have from master.
```
BuildLoader.py
-        self.PLD_HEAP_SIZE         = 0x02000000
+        self.PLD_HEAP_SIZE         = 0x0F424000

Platform/QemuBoardPkg/BoardConfig.py
-        self.HAVE_VERIFIED_BOOT       = 1
+        self.HAVE_VERIFIED_BOOT       = 0

Platform/QemuBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
-    PciOr8 (PCI_LIB_ADDRESS(0, 0, 0, MCH_ESMRAMC), MCH_ESMRAMC_T_EN);
+    //PciOr8 (PCI_LIB_ADDRESS(0, 0, 0, MCH_ESMRAMC), MCH_ESMRAMC_T_EN);
```

[...]

PR #1399, opened 2021-11-05
Enhance Ext2 filesystem library

mauricema:
For EXT2 filesystem revision 0, there are some fixed fields in the
super block structure according to the documentation. The code should
always use those fixed values for safe regardless of the value inside
the image.

### DIFF
--- a/BootloaderCommonPkg/Library/Ext23Lib/Ext2Fs.c
+++ b/BootloaderCommonPkg/Library/Ext23Lib/Ext2Fs.c
@@ -646,6 +646,11 @@ Ext2SbValidate (
     goto Exit;
   }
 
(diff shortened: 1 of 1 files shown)


myguitar (approved):
Glad to know this! May I know which image you were using for Ext2 Rev0?

mauricema:
> Glad to know this! May I know which image you were using for Ext2 Rev0?
Please look into issue #1396.   The tool genimage will generate EXT2 disk image with Rev0.    Without, it will cause divided by 0 exception for EXT2 library.

myguitar:
> > Glad to know this! May I know which image you were using for Ext2 Rev0?
> > Please look into issue #1396.   The tool genimage will generate EXT2 disk image with Rev0.    Without, it will cause divided by 0 exception for EXT2 library.

Thank you Maurice!

PR #1403, opened 2021-11-08
[EHL] Disabled AC split lock by default

koktong-ong:
Disabled AC split lock by default in CfgData yaml file.


mauricema (approved):
Looks good to me.

PR #1404, opened 2021-11-08
[TGL] Remove SGX Configurability

japelner-intc:
This patch removes SGX configurability from Slim Bootloader as
SGX is not supported on TGL.

Signed-off-by: Jim <jim.pelner@intel.com>


mauricema (approved):
Looks good to me.

Issue #1400, opened 2021-11-05
Consider using LF line endings in the SlimBootloader Git repository


mikecrowe:
The SlimBootloader Git repository uses DOS-style CRLF line endings on all its text files in the repository. It appears that this is intentional since I see various commits fixing files that had different line endings to use CRLF. The requirement may come from the [EDK2 coding standards)[https://edk2-docs.gitbook.io/edk-ii-c-coding-standards-specification/5_source_files] but no rationale is given there. It looks like it [might be for historic reasons](https://www.mail-archive.com/edk2-devel@lists.01.org/msg00865.html) that no longer apply.

Text files with CRLF line endings are annoying to use from Linux. Whilst most editors can be configured to accept and preserve them, the line endings will often be changed when running files through other tools. Even running the Python scripts included in SlimBootloader rewrites files to use only LF line endings.

Modern Git's support for CRLF line endings in work trees when checking out on Windows but LF line endings on Linux and MacOS is generally good. However, this requires that the files are stored in the repository with LF line endings. In my research I've been unable to find a standard way[1] for standard Git to use LF line endings in the work tree when the repository blobs use CRLF. If the blobs in the repository use LF then it is straightforward for clients on Linux and MacOS to use LF in the work tree whilst clients on Windows will use CRLF in the work tree. See https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings. If any files need to have CRLF line endings even on Linux then this is possible using `gitattributes` directives.

SlimBootloader is a great way to boot Linux on Intel hardware. Developers using Linux for their products will most likely be building their software on Linux and will want to edit and compile SlimBootloader from Linux.

It's clear that files with only LF endings are sneaking into the repository (there are currently at least six), and using Git's standard mechanisms ought to avoid inconsistent line endings being introduced in the future.

Please can you consider switching to using LF line endings in the repository?

Thanks.

Mike.

[1] It is possible to use filters to smudge files to change the line endings as they are checked out. However, this requires extra configuration after cloning the repository. It doesn't happen automatically.

mauricema:
Thank you for the suggestion.  The primary reason that we used CRLF is just to be in sync with EDK2 since we leverage some of the code direct from EDK2 project, and we don't want to cause unnecessary issue to us.  What you mentioned is reasonable to me.  Let me bring this topic back to the team and then update here on the results.

mikecrowe:
[...]

PR #1408, opened 2021-11-10
Use LF line endings in the repository

mikecrowe:
Convert the line endings stored for all text files in the repository to
LF. The majority previously used DOS-style CRLF line endings. Add a
.gitattributes file to enforce this and treat certain extensions as
never being text files.

This has implications depending on the client platform you are using the
repository from:

* Windows

The usual configuration for Git on Windows means that text files will
be checked out to the work tree with DOS-style CRLF line endings. If
that's not the case then you can configure Git to do so for the entire
machine with:

 git config --global core.autocrlf true

or for just the repository with:

 git config core.autocrlf true

Line endings will be normalised to LF when they are committed to the
repository. If you commit a text file with only LF line endings then it
will be converted to CRLF line endings in your work tree.

* Linux, MacOS and other Unices

The usual configuration for Git on such platforms is to check files out
of the repository with LF line endings. This is probably the right thing
for you. In the unlikely even that you are using Git on Unix but editing
or compiling on Windows for some reason then you may need to tweak your
configuration to force the use of CRLF line endings as described above.

* General

For more information see
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings .

Fixes: https://github.com/slimbootloader/slimbootloader/issues/1400


mikecrowe:
The check failures appear to be due to all the files that already existed in the repository that do not pass `PatchCheck.py`. Although I changed that script as part of this commit to require LF line endings, that isn't enough to stop it complaining about all the pre-existing problems.

gdong1:
> The check failures appear to be due to all the files that already existed in the repository that do not pass `PatchCheck.py`. Although I changed that script as part of this commit to require LF line endings, that isn't enough to stop it complaining about all the pre-existing problems.

Yes, the failures mainly caused by existing coding style issues (* Tab character used, Trailing whitespace found). These issues could be fixed separately.

mikecrowe:
> Yes, the failures mainly caused by existing coding style issues (* Tab character used, Trailing whitespace found). These issues could be fixed separately.

At least some of the complaints appeared to be in imported code, so I was wary of modifying the files. Also, some of the tab complaints are for makefiles which are required to contain tab characters. :( Nevertheless, I agree that these issues should be addresses separately.

mauricema:
let me take a look to see if I can get it fixed.

mikecrowe:
Line 48 of the commit message is the URL. In other projects I've contributed to the commit line length rule has not been applied for lines that have a reason to be too long. In this case, I can't make the line shorter without breaking the URL. Using a URL shortner would risk the URL no longer being valid in the future. Should I just remove the URL?

I've added your patch as two separate commits. I'm not sure whether the CI pipeline will like it that way, but it's worth a try.

mauricema:
> Line 48 of the commit message is the URL. In other projects I've contributed to the commit line length rule has not been applied for lines that have a reason to be too long. In this case, I can't make the line shorter without breaking the URL. Using a URL shortner would risk the URL no longer being valid in the future. Should I just remove the URL?
> 
> I've added your patch as two separate commits. I'm not sure whether the CI pipeline will like it that way, but it's worth a try.

Sure.  I am fine with not breaking the URL line.

mauricema:
Looks like separate commit in patch set will not make the patch checker happy since it checks for each patch separately.
At this moment, I would think we just force push this change given the facts that all builds passed and we know the checker will pass after all commits are merged in.

mikecrowe:
> Looks like separate commit in patch set will not make the patch checker happy since it checks for each patch separately.
At this moment, I would think we just force push this change given the facts that all builds passed and we know the checker will pass after all commits are merged in.

If you're happy to do that then so am I. However, I set you as the author of the two commits from your patch and didn't feel able to add Signed-off-by lines in your name. You might want to reword them too since my description was rather inadequate.

Thanks.

mauricema (approved):
Thanks for the PR changes pf the file EOL style.  :)
Looks good to me.  It has known PatchChecker failure as discussed.
Since it can be addressed by other commits in the patch set,  
we plan to merge this in.

mauricema:
BTW, for the other 2 patches I might want to need to add "Signed off" in commit msg.

mauricema:
Or we can handle them totally separately through different patches. 
You only submit your first patch, and in this way I can rework on the patches under my name.

mikecrowe:
> You only submit your first patch, and in this way I can rework on the patches under my name.

OK. I'll remove your two patches from this PR.

mauricema:
> > You only submit your first patch, and in this way I can rework on the patches under my name.
> 
> OK. I'll remove your two patches from this PR.

Great, thank you so much!  Sorry to make you go back and forth.

mauricema:
Force push this commit per previous discussions.

mikecrowe:
Thanks for landing this. It will be of great help to us. The messing about was mostly my fault for trying to keep the changes separate.

mauricema:
No problem at all.  It was merged.  And I have submitted remaining patches.
Thanks for contributing to SBL project!  😃

PR #1412, opened 2021-11-15
Add a PCD to set RTCM RSVD SIZE

stalamudupula:
If the number of cores are more and the RTCM is
required to support HyperThreading, then it needs
more reserved size, preferably 511 pages instead of
current 255 pages.

So, add a FixedPcd and let each platform override
the default 255 pages value to whatever is required.


cshur (approved):

mauricema (approved):
Looks good to me.

mauricema (approved):

PR #1413, opened 2021-11-16
Add SMBIOS type 4 for processor info

mauricema:
This patch added required SMBIOS type 4 for processor info.

### DIFF
--- a/BootloaderCorePkg/Library/SmbiosInitLib/SmbiosInitLib.c
+++ b/BootloaderCorePkg/Library/SmbiosInitLib/SmbiosInitLib.c
@@ -15,13 +15,25 @@
 #include <Library/SmbiosInitLib.h>
 #include <Library/PcdLib.h>
 #include <Library/BootloaderCommonLib.h>
-#include <Library/FspSupportLib.h>
+#include <Library/TimeStampLib.h>
+#include <Library/MpInitLib.h>
 #include "SmbiosTables.h"
 
-#define SMBIOS_STRING_UNKNOWN              "Unknown"
-#define SMBIOS_STRING_UNKNOWN_VERSION      "XXXX.XXX.XXX.XXX"
-#define SMBIOS_STRING_VENDOR               "Intel Corporation"
-#define SMBIOS_STRING_PLATFORM             "Intel Platform"
+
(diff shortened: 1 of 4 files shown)


BootloaderCorePkg/Library/SmbiosInitLib/SmbiosInitLib.c
```diff
@@ -519,6 +554,146 @@ InitSmbiosStringPtr (
...
+              (UINT32 *)&CpuBrandBuf[BufIdx+4],
+              (UINT32 *)&CpuBrandBuf[BufIdx+8],
+              (UINT32 *)&CpuBrandBuf[BufIdx+12]
+             );
+    BufIdx += 16;
+  }
+  if (CpuBrandBuf[0] == 0) {
+    AsciiStrCpyS (CpuBrandBuf, sizeof(CpuBrandBuf), BRAND_STRING_UNSUPPORTED);
+  }
+  CpuBrandBuf[48] = 0;
```
stalamudupula, line 596: Why are we restricting the string to only 48 bytes?

BootloaderCorePkg/Library/SmbiosInitLib/SmbiosInitLib.c
```diff
@@ -160,6 +172,8 @@ FinalizeSmbios (
   EFI_STATUS                    Status;
   SMBIOS_TABLE_ENTRY_POINT      *SmbiosEntry;
 
+  Status = BuildProcessorInfo ();
```
stalamudupula, line 175: Can we move this to SmbiosInit (), if we can get the Mp info correctly by then?

mauricema: Yes, let me do that.

BootloaderCorePkg/Library/SmbiosInitLib/SmbiosInitLib.c
mauricema: CPUID can return 16 bytes for one cycle. We only do 3 cycles, so it should be less than 16*3.

stalamudupula:
Looks good to me.

myguitar (approved):
Thank you Maurice for helping this!

PR #1415, opened 2021-11-19
Print code bytes around eip/rip when taking a fatal exception

buytenh:
When we die due to a fatal exception, we get a debug message that
looks like this:

	Exception #6 from 0x0020:0x000729A1 !!!

This message by itself is not incredibly useful for figuring out
which code caused the exception.

This patch borrows an idea from the Linux kernel, and extends the
exception information with a dump of the opcode bytes around the
instruction pointer, which means that the exception debug message
will now look like this:

	Exception #6 from 0x0020:0x000729A1 !!!
	Code: 89 E0 B9 40 00 00 00 48 8D 15 E3 12 00 00 E8 4D E0 FF FF 41 B8 00 00 07 00 B9 40 00 00 00 48 8D 15 E9 12 00 00 E8 36 E0 FF FF <0F> 0B E8 7D E5 FF FF BA 01 00 00 00 4C 8D 8C 24 90 00 00 00 4C 8D

The byte marked as <0F> indicates the byte that the instruction pointer
points to, and we additionally print a number of prologue and epilogue
bytes, with more prologue bytes than epilogue bytes, according to the
reasoning used by and documented in the Linux kernel tree:

	https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/x86/kernel/dumpstack.c?h=v5.15#n100

Tested by inserting a UD2 instruction into Stage1A and verifying on
qemu ia32 and on qemu x64 that the right information is printed.

### DIFF
--- a/BootloaderCorePkg/Library/CpuExceptionLib/Ia32/CpuExceptionHandler.c
+++ b/BootloaderCorePkg/Library/CpuExceptionLib/Ia32/CpuExceptionHandler.c
@@ -39,6 +39,8 @@ CommonExceptionHandler (
   }
 
   DEBUG ((DEBUG_ERROR, "Exception #%d from 0x%04X:0x%08X !!!\n", ExceptionType, Ptr[1], Ptr[0]));
+  DumpHex(0, Ptr[0] - 32, 64, (UINT8 *)(Ptr[0] - 32));
+
   CpuHalt (NULL);
 }
 
--- a/BootloaderCorePkg/Library/CpuExceptionLib/X64/CpuExceptionHandler.c
+++ b/BootloaderCorePkg/Library/CpuExceptionLib/X64/CpuExceptionHandler.c
@@ -39,6 +39,8 @@ CommonExceptionHandler (
   }
 
   DEBUG ((DEBUG_ERROR, "Exception #%d from 0x%04X:0x%08X !!!\n", ExceptionType, (UINT32)Ptr[1], (UINT32)Ptr[0]));
+  DumpHex(0, Ptr[0] - 32, 64, (UINT8 *)(Ptr[0] - 32));
+
   CpuHalt (NULL);
 }
 


mauricema:
:)  Thank you for your contribution to SBL.  Definitely a good enhancement for debug exception.
You patch failed the patch check because of a long line in commit msg.  The detailed info is as below:
"
Checking git commit: master..72aecfc8e60d4ca81b6db4fbdf0328b66beca348
The commit message format is not valid:
 * Line 17 of commit message is too long.
https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format
"

BootloaderCorePkg/Library/CpuExceptionLib/Ia32/CpuExceptionHandler.c
```diff
@@ -39,6 +41,18 @@ CommonExceptionHandler (
   }
 
   DEBUG ((DEBUG_ERROR, "Exception #%d from 0x%04X:0x%08X !!!\n", ExceptionType, Ptr[1], Ptr[0]));
+
```
mauricema, line 44: We have a common function called DumpHex().  
Do you think we can do  DumpHex (0, Ip - 42, 64, Ip - 42) to make it simpler ?  It won't have <> around the current IP, but with the address print, it should be easy to find the related instruction.

BootloaderCorePkg/Library/CpuExceptionLib/X64/CpuExceptionHandler.c
```diff
@@ -39,6 +41,18 @@ CommonExceptionHandler (
   }
 
   DEBUG ((DEBUG_ERROR, "Exception #%d from 0x%04X:0x%08X !!!\n", ExceptionType, (UINT32)Ptr[1], (UINT32)Ptr[0]));
+
```
mauricema, line 44: This part seems to be common for IA32 and X64.  Maybe we can move it out to up level so that we only need to maintain one copy.   But we can do it as separate patch I guess.  :)

BootloaderCorePkg/Library/CpuExceptionLib/Ia32/CpuExceptionHandler.c
buytenh: Sounds good!  I changed it to `DumpHex(0, Ptr[0] - 32, 64, (UINT8 *)(Ptr[0] - 32));`, which gives us a few fewer prologue bytes, but it aligns the faulting instruction to the start of the third line, which makes it easy to find.  I've also updated the commit message to match.

BootloaderCorePkg/Library/CpuExceptionLib/X64/CpuExceptionHandler.c
buytenh: Makes sense to me to do this as a separate patch. :)

BootloaderCorePkg/Library/CpuExceptionLib/Ia32/CpuExceptionHandler.c
mauricema: Looks good to me. 👍

mauricema (approved):
Looks good to me.

PR #1418, opened 2021-11-19
Allow building with openssl 1.0.2k

buytenh:
This is the openssl version that ships with RHEL7, and building with
this version seems to be working OK.

### DIFF
--- a/BootloaderCorePkg/Tools/BuildUtility.py
+++ b/BootloaderCorePkg/Tools/BuildUtility.py
@@ -36,7 +36,7 @@ build_toolchains = {
     'python'    : '3.6.0',
     'nasm'      : '2.12.02',
     'iasl'      : '20160422',
-    'openssl'   : '1.1.0g',
+    'openssl'   : '1.0.2k',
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -36,7 +36,7 @@
     'python'    : '3.6.0',
     'nasm'      : '2.12.02',
     'iasl'      : '20160422',
-    'openssl'   : '1.1.0g',
+    'openssl'   : '1.0.2k',
```
mauricema, line 39: Our project requires us to use newer openssl (at least 1.1.0).   It is part of the security requirement, and it just tries to avoid any potential vulnerability in the old openssl.  So might be hard to do this change on our side.

buytenh: OK, I understand.  Thank you!

mauricema: Thank you for understanding!

PR #1419, opened 2021-11-25
Fix cfgdata_dir argument issue in CfgDataStitch

mauricema:
This patch fixed the issue that "cfgdata_dir" argument is not
used properly by the CfgDataStitch script. The "cfgdata_dir"
will contain the CfgDataDef.yaml file and optional DLT files
if provided. Otherwise, the DLT files will be extracted from
the IFWI binary and stored in "cfgdata_dir".


myguitar (approved):
It looks good to me. Thanks a lot Maurice for helping this!

PR #1420, opened 2021-11-25
[EHL] Removed hardcoded PSE PWM pin enable

koktong-ong:
Removed hardcoded PSE PWM pin enable and adapt from
CfgData in Stage2.


mauricema (approved):
Looks good.

PR #1423, opened 2021-11-29
[EHL] Update FSP to MR2

kokweich:
Update FSP version to MR2 in inf file

Signed-off-by: kokweich <kok.wei.chan@intel.com>


mauricema (commented):
Looks good to me.

mauricema (approved):

PR #1425, opened 2021-12-01
[TGL][EHL] Fix regression for Flash descriptor lock

lsubash:
Add BootMediaWriteByType and use for flash descriptor
update.


randylintw:
validated and test pass.

raghavag (approved):
Changes look good

PR #1427, opened 2021-12-02
[X64] Fix invalid 1st timestamp at reset vector

myguitar:
The timestamp value is saved in edi:esi at reset vector in X64, but
PreparePagingTable is using esi and makes 1st timestamp value invalid.

This saves/restores edi:esi before/after PreparePagingTable to keep
proper timestamp value at reset vector.


mauricema (approved):
Looks good.

PR #1429, opened 2021-12-03
[TGL] Add stitch option to support TGL-H RVP Config 3

randylintw:
Config 1/3 RVP boards can't share same IFWI image
and it is required to adjust the FIT parameters.
Add -o cfg3 to support this.


mauricema (approved):
Looks good.

PR #1431, opened 2021-12-09
[TGL] Fix infinite reset loop caused by bad DSO

stanleyintel:
This patch solves an infinite reset loop issue caused by bad DSO with the scenario:
- After platform reset (due to WDT timeout), FSPm asks for another reset, but before that, WDT_FLAG_TCC_DSO is already cleaned. As a result, in the thrid reset, stage1B will have no idea about the DSO is corrupted and it continues boot with Tcc Tuning flow, which causes WDT timeout reset again.

This patch introduces a WDT_FLAG_TCC_BAD_DSO flag in WDT scrachpad (bit 18). The flag is a marker that is set when a bad DSO is detected. The new booting flow for "bad DSO" case if Tcc_Tuning enabled will be:
  - 1st boot: (after fwupdate)
      - TCC_DSO and WDT set by stage1b and stage2
      - FSP hangs and trigger WDT reset
  - 2nd boot:
      - Stage1b detects "bad DSO" because of WDT and TCC_DSO.
      - It clears TCC_DSO and WDT.
      - Then, it sets TCC_BAD_DSO.
      - Then it continues boot that will skip Tcc Tuning (because of TCC_DSO unset)
  - 3rd boot:
      - Stage1b detects "bad DSO" because of TCC_BAD_DSO
      - It continues boot that will skip Tcc Tuning (because of TCC_DSO unset)

The patch does not remove the 200-sec abnormal boot-up symptom because the symptom is noticeable to user. So user can be aware of something wrong (bad DSO).

The "bad DSO" flag will be clear before fwupdate, so a fwupdate with a correct DSO can solve the 200 sec abnormal boot up time.

### DIFF
--- a/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -902,8 +902,8 @@ BoardInit (
     Status = PcdSet32S (PcdAcpiTableTemplatePtr, (UINT32)(UINTN)mPlatformAcpiTables);
     break;
   case PostSiliconInit:
-    if (IsWdtFlagsSet(WDT_FLAG_TCC_DSO)) {
-      WdtDisable (WDT_FLAG_TCC_DSO);
+    if (IsWdtFlagsSet(WDT_FLAG_TCC_DSO_IN_PROGRESS)) {
+      WdtDisable (WDT_FLAG_TCC_DSO_IN_PROGRESS);
     }
     // Set TSEG base/size PCD
     TsegBase = MmioRead32 (TO_MM_PCI_ADDRESS (0x00000000) + R_SA_TSEGMB) & ~0xF;
@@ -1024,6 +1024,9 @@ BoardInit (
   case ReadyToBoot:
     if ((GetBootMode() != BOOT_ON_FLASH_UPDATE) && (GetPayloadId() == 0)) {
       ProgramSecuritySetting ();
+    } else if (GetBootMode() == BOOT_ON_FLASH_UPDATE) {
+      /* clear bad DSO mark (if have), so next boot is a fresh restart */
+      WdtClearScratchpad (WDT_FLAG_TCC_BAD_DSO);
     }
 
     break;
@@ -1190,7 +1193,7 @@ TccModePostMemConfig (
   FspsUpd->FspsConfig.TccErrorLogEn   = TccCfgData->TccErrorLog;
   FspsUpd->FspsConfig.IfuEnable       = 0;
 
-  if (!IsWdtFlagsSet(WDT_FLAG_TCC_DSO)) {
+  if (!IsWdtFlagsSet(WDT_FLAG_TCC_DSO_IN_PROGRESS)) {
     //
     // If FSPM doesn't enable TCC DSO timer, FSPS should also skip TCC DSO.
     //
@@ -1198,7 +1201,7 @@ TccModePostMemConfig (
     FspsUpd->FspsConfig.TccStreamCfgStatus = 1;
   } else if (TccCfgData->TccTuning != 0) {
     // Reload Watch dog timer
-    WdtReloadAndStart (WDT_TIMEOUT_TCC_DSO, WDT_FLAG_TCC_DSO);
+    WdtReloadAndStart (WDT_TIMEOUT_TCC_DSO, WDT_FLAG_TCC_DSO_IN_PROGRESS);
 
     // Load TCC stream config from container
     TccStreamBase = NULL;
--- a/Silicon/CommonSocPkg/Include/Library/WatchDogTimerLib.h
+++ b/Silicon/CommonSocPkg/Include/Library/WatchDogTimerLib.h
@@ -8,9 +8,9 @@
 #ifndef WATCH_DOG_TIMBER_LIB_H_
 #define WATCH_DOG_TIMBER_LIB_H_
 
-#define  WDT_TIMEOUT_TCC_DSO   200          // 200 seconds
-#define  WDT_FLAG_TCC_DSO      BIT17
-#define  WDT_FLAG_MASK         BIT17
+#define  WDT_TIMEOUT_TCC_DSO           200    // 200 seconds
+#define  WDT_FLAG_TCC_DSO_IN_PROGRESS  BIT17
+#define  WDT_FLAG_TCC_BAD_DSO          BIT18
 
 
 /**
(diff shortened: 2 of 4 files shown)


randylintw:
Tested with bad DSO updated and still can enter Yocto OS. 
Also can trigger fwupdate to replace the bad DSO with good one.

Silicon/CommonSocPkg/Library/WatchDogTimerLib/WatchDogTimerLib.c
```diff
@@ -98,14 +98,14 @@ WdtDisable (
...
+  Clear WDT TCC flags in scratchpad
 
-  @param[in] Flags             The timer flags that would be cleared.
+  @param[in] Flags             The scratchpad flags that would be cleared.
 
 **/
 VOID
 EFIAPI
-WdtClearFlags (
+WdtClearTccScratchpad (
```
mauricema, line 108: WdtDogTimerLib is for common usage, and it should not be specific to TCC.
Please keep the API interface generic and use the flag to indicate TCC use cases.

Silicon/CommonSocPkg/Include/Library/WatchDogTimerLib.h
```diff
@@ -44,14 +45,26 @@ WdtDisable (
...
+
+/**
+  Set WDT TCC flags in scratchpad
+
+  @param[in] Flags             The scratchpad flags that would be set.
+
+**/
+VOID
+EFIAPI
+WdtSetTccScratchpad (
```
mauricema, line 67: WdtDogTimerLib is for common usage, and it should not be specific to TCC.
Please keep the API interface generic and use the flag to indicate TCC use cases.

Silicon/CommonSocPkg/Library/WatchDogTimerLib/WatchDogTimerLib.c
```diff
@@ -116,6 +116,25 @@ WdtClearFlags (
...
 
+/**
+  Set WDT TCC flags in scratchpad
+
+  @param[in] Flags             The scratchpad flags that would be set.
+
+**/
+VOID
+EFIAPI
+WdtSetTccScratchpad (
```
mauricema, line 127: WdtDogTimerLib is for common usage, and it should not be specific to TCC.
Please keep the API interface generic and use the flag to indicate TCC use cases.

Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1024,6 +1024,8 @@ BoardInit (
   case ReadyToBoot:
     if ((GetBootMode() != BOOT_ON_FLASH_UPDATE) && (GetPayloadId() == 0)) {
       ProgramSecuritySetting ();
+    } else if (GetBootMode() == BOOT_ON_FLASH_UPDATE) {
```
mauricema, line 1027: why do we only clear it for FWU path ?

Silicon/CommonSocPkg/Include/Library/WatchDogTimerLib.h
```diff
@@ -10,7 +10,8 @@
 
 #define  WDT_TIMEOUT_TCC_DSO   200          // 200 seconds
 #define  WDT_FLAG_TCC_DSO      BIT17
-#define  WDT_FLAG_MASK         BIT17
+#define  WDT_FLAG_TCC_BAD_DSO  BIT18
```
mauricema, line 13: In this case, I feel we need to rename WDT_FLAG_TCC_DSO to be more specific so that it is clear on the difference with BAD_DSO.   Maybe we can use WDT_FLAG_TCC_DSO_IN_PROGRESS ?

Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
stanleyintel: if we don't clear the flags in FWU path, after FWU, the flags (e.g., BAD DSO), may still exist (warm boot path). Then boot flow will bypass the DSO tuning in new boot.

mauricema: I meant why don't we clear it on all boot path regardless to make it simpler ?  Do you see any issue ?

Silicon/CommonSocPkg/Library/WatchDogTimerLib/WatchDogTimerLib.c
stanleyintel: The original implementation (line 115) only takes care of bit fields within WDT_FLAG_MASK, which is to clear scratchpad and no other bit fields. Therefore, WdtClearScratchpad should be more precise. (I will rename it to WdtClearScratchpad () in the next revision)

Silicon/CommonSocPkg/Include/Library/WatchDogTimerLib.h
stanleyintel: Thx for the comment. Will rename it.

Silicon/CommonSocPkg/Library/WatchDogTimerLib/WatchDogTimerLib.c
stanleyintel: will be more generic, with a mask (so no need to be for scratchpad only)

Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
stanleyintel: Current logic allows that user only needs to experience 200-sec WDT for each cold boot (if BAD DSO exists).
In current logic, we keep the BAD DSO flag exists, so that in next boot, stage1b can know the existence of BAD DSO and skip Tcc tuning. Otherwise, user will experience 200-sec WDT for every warm boot.

It is OKAY, in terms of solving this infinite reset loop issue, to clear it on all boot paths. However, it is a bit ambiguous, in terms of programming simplification and UX. Please advise.

Silicon/CommonSocPkg/Include/Library/WatchDogTimerLib.h
stanleyintel: rename to WdtSetScratchpad() to keep common usage and focus on scratchpad only

Silicon/CommonSocPkg/Library/WatchDogTimerLib/WatchDogTimerLib.c
stanleyintel: rename to WdtSetScratchpad() to keep common usage and focus on scratchpad only

mauricema (approved):

Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
mauricema: OK.

PR #1437, opened 2021-12-16
Update BtgSign tool

gdong1:
Currently StitchLoader.py is under platform package, and
the common tool BtgSign.py should not depend on that tool.
And BtgSign.py indeed doesn't depend on it, so just update
it to make it could work without StitchLoader.py.


mauricema (approved):
Looks good.

PR #1439, opened 2021-12-17
[TGL] Invalidate bad DSO region

stanleyintel:
This patch invalidates the DSO region with a marker when a corrupted DSO hang is detected (by WDT timeout). With the marker, stage1b can know the DSO tuning should be skipped.

The bad DSO mark is defined as both signature and size in the TCCT component header are zero.

With this patch, the previous defined WDT scratchpad bit, WDT_FLAG_TCC_BAD_DSO, is removed..

### DIFF
--- a/Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
+++ b/Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
@@ -58,6 +58,117 @@ GetBoardId (
   OUT UINT8 *BoardId
 );
 
+
+/**
+  Check if the BAD DSO mark exists or not
+  BAD DSO mark is defined as
+    LOADER_COMPRESSED_HEADER of the component in flash
+    Its Signature' and 'Size' fields are 0 (zero).
+    Other fields are subject to change
+
+  @retval TRUE  DSO was marked as BAD DSO
+  @retval FALSE No BAD DSO mark found
+
+ */
+BOOLEAN
+EFIAPI
+IsMarkedBadDso (
+  VOID
+)
+{
+  EFI_STATUS               Status;
+  UINT32                   Length;
+  LOADER_COMPRESSED_HEADER *Hdr;
+  LOADER_COMPRESSED_HEADER BadDsoMark = {0};
+
+  DEBUG ((DEBUG_INFO, "Check BAD DSO mark\n"));
+
+  Status = LocateComponent (SIGNATURE_32 ('I', 'P', 'F', 'W'),
+                            SIGNATURE_32 ('T', 'C', 'C', 'T'),
+                            (VOID *)&Hdr, &Length);
+  if (EFI_ERROR (Status) || (Length < sizeof(LOADER_COMPRESSED_HEADER))) {
+    return FALSE;
+  }
+
+  if ((Hdr->Signature == BadDsoMark.Signature) && (Hdr->Size == BadDsoMark.Size)) {
+    DEBUG ((DEBUG_INFO, "BAD DSO(TCCT) detected!\n"));
+    return TRUE;
+  } else {
+    return FALSE;
+  }
+}
+
+/**
+  Invalidate BAD DSO
+
+  @retval EFI_SUCCESS
+  @retval EFI_NOT_FOUND          Unable to find IPFW/TCCT
+  @retval EFI_OUT_OF_RESOURCES   Bios region is too small
+  @retval Others                 Errors during SPI operations
+
+ */
+EFI_STATUS
+EFIAPI
+InvalidateBadDso (
+  VOID
+)
+{
+  EFI_STATUS  Status;
+  UINT32      Address;
+  UINT32      Length;
+  UINT32      BaseAddress;
+  UINT32      RegionSize;
+  CONTAINER_ENTRY  *ContainerEntry;
+  COMPONENT_ENTRY  *CompEntry;
+  CONTAINER_HDR    *ContainerHdr;
+  // BAD DSO mark: The 'Signature' and 'Size' fields are zero(0)
+  // Other fields are subject to change
+  LOADER_COMPRESSED_HEADER BadDsoMark = {0};
+
+  DEBUG ((DEBUG_INFO, "Invalidate BAD DSO region\n"));
+
+  Status = LocateComponentEntry (SIGNATURE_32 ('I', 'P', 'F', 'W'),
+                                 SIGNATURE_32 ('T', 'C', 'C', 'T'),
+                                 &ContainerEntry, &CompEntry);
+  if (EFI_ERROR (Status) || (ContainerEntry == NULL) || (CompEntry == NULL)) {
+    return EFI_NOT_FOUND;
+  }
+
+  /* a region should have 4KB as min erase size */
+  Length = SIZE_4KB;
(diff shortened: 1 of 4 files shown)


sindhuragrandhi:
Why do we need this complex method when we can simply get away with a scratch pad bit for indicating bad DSO?

sindhuragrandhi:
Also, if this is truly required, other projects will have to do the same. Please see if the main functions can be moved to a common library if possible.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -58,6 +58,112 @@ GetBoardId (
...
+  Status = LocateComponentEntry (SIGNATURE_32 ('I', 'P', 'F', 'W'),
+                                 SIGNATURE_32 ('T', 'C', 'C', 'T'),
+                                 &ContainerEntry, &CompEntry);
+  if (EFI_ERROR (Status) || (ContainerEntry == NULL) || (CompEntry == NULL)) {
+    return EFI_NOT_FOUND;
+  }
+  ContainerHdr = (CONTAINER_HDR *)(UINTN)ContainerEntry->HeaderCache;
+
+  Address = ContainerEntry->Base + ContainerHdr->DataOffset + CompEntry->Offset;
+  Length = KB_(4); /* minimum erase size */
```
mauricema, line 142: use SIZE_4KB instead.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -58,6 +58,112 @@ GetBoardId (
...
+  UINT32      Length;
+  UINT32      BaseAddress;
+  UINT32      RegionSize;
+  CONTAINER_ENTRY  *ContainerEntry;
+  COMPONENT_ENTRY  *CompEntry;
+  CONTAINER_HDR    *ContainerHdr;
+
+  DEBUG ((DEBUG_INFO, "Invalidate Bad DSO\n"));
+
+  Status = LocateComponentEntry (SIGNATURE_32 ('I', 'P', 'F', 'W'),
```
mauricema, line 133: You can use LocateComponent() instead.  It will get the address and size directly.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
mauricema: Also to be safe, need to check the component size is greater than or equal to 4KB before marking it bad.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -58,6 +58,112 @@ GetBoardId (
...
+  }
+
+  BaseAddress = ((UINT32)(~RegionSize) + 1);
+
+  Address -= BaseAddress;
+  if ((Address + Length) > RegionSize) {
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  Status = SpiFlashErase (FlashRegionBios, Address, Length);
```
mauricema, line 156: Can we just write 0 to the location as a bad marker ?  In this way, the erasing can be avoided to reduce flash erasing frequency.  For flash, you can write from 1 to 0 without erasing.  But to write from 0 to 1, erasing is required. So writing 0 does not need any erasing.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -58,6 +58,112 @@ GetBoardId (
...
+  UINT32      Address;
+  UINT64      Data;
+
+  CONTAINER_ENTRY  *ContainerEntry;
+  COMPONENT_ENTRY  *CompEntry;
+  CONTAINER_HDR    *ContainerHdr;
+
+  DEBUG ((DEBUG_INFO, "Check Bad DSO mark\n"));
+
+  Status = LocateComponentEntry (SIGNATURE_32 ('I', 'P', 'F', 'W'),
```
mauricema, line 87: Use LocateComponent()

mauricema:
> Also, if this is truly required, other projects will have to do the same. Please see if the main functions can be moved to a common library if possible.

This is a more generic approach, and it is common.   We can move it to common code once it is fully tested.  Scratch pad is project specific, and really hard to find available scratch pad register.

mauricema:
> Why do we need this complex method when we can simply get away with a scratch pad bit for indicating bad DSO?

I believe this is a better approach.  Scratch pad info will be lost after a power cycle, and it will retrigger the whole WDT timeout flow on next boot.   Marking it in flash makes it consistent so that this bad DSO will only be tried once.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
stanleyintel: Thanks. Will update accordingly.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
stanleyintel: Good suggestion. Thanks. Will update accordingly.

stanleyintel: Unfortunately, all-zero Component will break anti-rollback, so my new patch keeps "Svn" field non-zero.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -58,6 +58,116 @@ GetBoardId (
   OUT UINT8 *BoardId
 );
 
+
+/* BAD DSO mark:
+     The 'Signature' and 'Size' fields are zero(0).
+     Other fields, like Svn, are subject to change
+  */
+STATIC LOADER_COMPRESSED_HEADER mBadDsoMark = {0};
```
mauricema, line 66: For Stage1B, we cannot use global variable.  The reason is that Stage1B could be running from Flash.
Please check if we can avoid using this global variable.

stanleyintel: Thanks and will be updated.

Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
stanleyintel: LocateComponent() is now used in IsMarkedBadDso(). 
Here it still needs to use LocateComponentEntry() to get CompEntry->Size for further check

mauricema (approved):
Looks good to me.

PR #1440, opened 2021-12-18
Check fit header entry location

lsubash:
For bootguard event log fit header location is retrieved
from fit pointer. Added a condition to check the FIT header
range to fix KW error.

### DIFF
--- a/Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
+++ b/Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
@@ -512,7 +512,15 @@ FindFitEntryData (
   UINT32                         Index;
 
   FitTableOffset = *(UINT64 *)(UINTN)(BASE_4GB - 0x40);
+  // Fit table is located in Top Swap region.
+  // Adding check to validate Fit entry location
+  if ((FitTableOffset < (BASE_4GB - PcdGet32 (PcdTopSwapRegionSize)))  ||  (FitTableOffset >  BASE_4GB)) {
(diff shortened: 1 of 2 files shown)


Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
```diff
@@ -512,7 +512,14 @@ FindFitEntryData (
   UINT32                         Index;
 
   FitTableOffset = *(UINT64 *)(UINTN)(BASE_4GB - 0x40);
+  // Fit table is located in Top Swap region. Top swap region is always less than 4MB for slimboot.
```
mauricema, line 515: If we need to valid FIT, should we do it in a more common place ?  
Also , the 4MB might change later on.   If we want to limit,  I would recommend to introduce PCD to allow config.

lsubash: The limit should be in "Top Swap sizes". I will introduce a PCD to get top swap size configured from Boardconfig.
For the common FIT  functionality, I think we could plan add a separate lib as per the FIT specification later.

mauricema: Or we can consider check it in SBL build process when we generate FIT.

lsubash: Yes, We can check in SBL build process. Incase FIT pointer location is incorrect system doesn't boot.
These additional checks are added for Klocwork cases.

mauricema (approved):
Looks good.

PR #1441, opened 2021-12-20
add private key check for PKCS8 format

liux2085:
Since openssl 3.0.0, in commit 10203a34725e ("Support writing RSA
keys using the traditional format again"), the default format of
private key has been changed to PKCS#8.

Signed-off-by: Yongxin Liu <yongxin.liu@windriver.com>


mauricema (approved):
It looks good to me.  
Thank you for the patch!  :)

PR #1444, opened 2021-12-23
Move DSO update/check to TccLib

gdong1:
IsMarkedBadDso and InvalidateBadDso would be required for all
the platforms that support TCC. And the implementation is also
common, so just move them to common TccLib.
Also updated the implementation to remove SPI flash erase for
InvalidateBadDso().


mauricema (approved):
Looks good.

myguitar (approved):

PR #1446, opened 2021-12-23
[CMLV] Fix wrong value of PchMaxPciePort

randylintw:
According to H410 PCH EDS data, it should be 20.


randylintw:
co-work with Vincent Chen <vincent.chen@intel.com>

mauricema (approved):
Looks good.

PR #1447, opened 2021-12-24
Fix fwupdate check for BAD DSO mark

stanleyintel:
BAD DSO mark is simplified to only use Signature=0
when moving DSO update/check to TccLib.
Related check in fwupdate has to be updated too.


mauricema (approved):
Looks good.

PR #1445, opened 2021-12-23
[CFL] Fix Klocwork scanning issue

randylintw:
Fix the wrong bootmode comparision.

Signed-off-by: Randy Lin <randy.lin@intel.com>

PR #1448, opened 2021-12-27
[CFL] Fix Klocwork scanning issue

randylintw:
Fix the Expression BootMode can never reach the value.


randylintw:
The https://github.com/slimbootloader/slimbootloader/pull/1445 was replace by another one so that code change isn't merged, resubmit again.

mauricema (approved):
Looks good to me.

PR #1449, opened 2021-12-29
Update IASL to version 20190509

jgutbub:
For Azure and toolchain check we
can update to using IASL version
20190509 since some of the newer
SBL platforms require a newer
IASL version.


gdong1 (approved):

mauricema (approved):
Looks good.

PR #1450, opened 2022-01-05
[CML] Fix ACPI GPE 0x6F interrupt storm

randylintw:
RTD3 table isn't ready so that comment _L6F out.


mauricema (approved):
Looks good to me.

PR #1457, opened 2022-01-12
HelloWorld: Use ConsolePrint to simplify code and fix GCC build

mikecrowe:
This compiles, but I'm having some other problems which mean that I haven't been able to test the result yet.

I suspect that the automated checks will complain about the length of the quoted compiler output in the commit message. I can't really fix that without making the output unreadable or removing it altogether.

### DIFF
--- a/PayloadPkg/HelloWorld/HelloWorld.c
+++ b/PayloadPkg/HelloWorld/HelloWorld.c
@@ -22,8 +22,6 @@ PayloadMain (
   )
 {
   UINT8      Key;
-  CHAR8      Message[64];
-  UINTN      Len;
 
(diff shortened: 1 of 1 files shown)


mauricema:
Below is the Azure Pipeline check  error:
2022-01-12T20:41:57.0190508Z  * Missing Signed-off-by! (Note: this must be added by the code contributor!)
So please use "git commit --amend -s" to add signed-off-by.

Also it has warning on the long commit message line.
WARNING - Line 6 of commit message is too long (100 >= 76).
In this case, it might be OK since it is a reference to the code line.

mauricema (approved):
Looks good to me.

PR #1462, opened 2022-01-14
BuildUtility: Support GCC5_BIN prefix

mikecrowe:
The EDK II build tools support applying a prefix for GCC from an
environment variable (e.g. GCC5_BIN). For this to work correctly with
Slimbootloader it's also necessary to use the prefix for the version
check too.

Signed-off-by: Mike Crowe <mac@mcrowe.com>


gdong1 (approved):

mauricema (approved):
Looks good.

PR #1463, opened 2022-01-14
PrepareBuildComponentBin: Avoid network access if commit is already available

mikecrowe:
It may be necessary to build Slimbootloader in situations where Internet
access is not available, or where it is important to ensure that there
is also a local copy of the build dependencies.

It's possible for the correct repositories to be cloned into the right
places outside the Slimbootloader build system. However,
PrepareBuildComponentBin previously insisted on attempting to
fetch from the remote repository every time.

If the .git directory exists, let's first try to check out the required
commit and only if that fails attempt to fetch.

This means that it's possible to stop the Slimbootloader build system
accessing the Internet by arranging for the correct repositories to be
cloned in the right places beforehand and setting the https_proxy
environment variables to unresolvable names to avoid accidental use.

Signed-off-by: Mike Crowe <mac@mcrowe.com>


gdong1 (approved):

mauricema (approved):
Looks good.

PR #1465, opened 2022-01-20
[TGL] Fix UART0 access

stanleyintel:
This patch fixes no activity on UART0 pins when enabling it for serial
communication.

In TGL, there are two UART0 instances while the one (GPP_F0 to GPP_F4)
is shared with CNVI. This patch enables GPP_C8~C11 as the UART0 instance
to reduce the conflict with CNVI.

This patch also fixes the GPIO pins definition for TGL-H and moves
serial io initialization code to SerialIo.c to simplfy Stage2BoardInitLib.c.

Test: TGL-UP3 RVP and TGL-H RVP

### DIFF
--- a/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/SerialIo.c
+++ b/Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/SerialIo.c
@@ -1,0 +1,52 @@
+/** @file
+
+  Copyright (c) 2022, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <Library/DebugLib.h>
+#include <FspsUpd.h>
+#include <Library/PchInfoLib.h>
+#include <PlatformData.h>
+#include <PchAccess.h>
+#include <PchLimits.h>
+#include "SerialIo.h"
+
+/**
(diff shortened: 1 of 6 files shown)


gdong1 (approved):

Platform/TigerlakeBoardPkg/Library/Stage2BoardInitLib/SerialIo.c
```diff
@@ -0,0 +1,52 @@
...
+    // Inform FSP to skip debug UART init
+    FspsConfig->SerialIoDebugUartNumber = DebugPort;
+    FspsConfig->SerialIoUartMode[DebugPort] = 0x4;
+  } else if (S0IX_STATUS() == 1) {  // legacy UART
+    FspsConfig->SerialIoUartMode[2] = 1;  // Force UART to PCI mode to enable OS to have full control
+  }
+
+  // Two instants of UART0 in TGL PCH: GPP_C8 to GPP_C11 and GPP_F0 to GPP_F3, while the second
+  // instant can be used by CNVI.
+  if (IsPchLp ()) {
```
mauricema, line 39: Maybe better to have a CFG data for this ?  If user wants to use F0-F3 as UART,  then they don't have to change the code.
It can be a follow up patch though.

mauricema (approved):
Looks good.

PR #1472, opened 2022-01-27
[EHL] Remove FSP headers

koktong-ong:
Remove FSP headers due to the FSP download script
will check if FSP header exist. It will only replace
if there is no existing FSP headers.


mauricema (approved):
Looks good.

PR #1473, opened 2022-01-27
[TGL] Disable s0ix on TGLU RVP

randylintw:
s0ix feature enabling flag also turn off some FSP configs
so that default SBL image can't detect the onboard Lan
and type c devices.


mauricema (approved):
Looks good to me.

PR #1474, opened 2022-01-28
[TGL] Enable PCIe PM features

stanleyintel:
The patch enables PCIe PM features by,
    1. Store Root Port configuration before FSP-s.
    2. Configure Pcie RP in PostPciEnumeration with the stored RP config.

The feature is controlled by ENABLE_PCIE_PM and the corresponding PcdEnablePciePm

The implementation is silicon-dependent, because of registers definition.
The PciePmNull component is a generic implementation. This patch also implements PciePm for TGL.

Verified: TGL-U RVP

### DIFF
--- a/Silicon/TigerlakePchPkg/Library/BasePcieHelperLib/BasePcieHelperLib.c
+++ b/Silicon/TigerlakePchPkg/Library/BasePcieHelperLib/BasePcieHelperLib.c
@@ -1,0 +1,315 @@
+/** @file
+  This file contains routines that support PCI Express initialization
+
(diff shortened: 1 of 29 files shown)


Silicon/TigerlakePchPkg/Library/BasePcieHelperLib/BasePcieHelperLib.c
```diff
@@ -0,0 +1,315 @@
+/** @file
+  This file contains routines that support PCI Express initialization
```
mauricema, line 2: This file seems to be generic.  Shouldn't it be in SocCommonPkg ?

mauricema (approved):
Looks good to me.

PR #1478, opened 2022-02-08
Fix condition fields may disappear in Config Editor tool

randylintw:
When enable TGL IBECC, the IbeccProtectedRegionMask should be
visible too in the Config Editor but its not happening this way.

Siged-off-by: Randy Lin <randy.lin@intel.com>

### DIFF
--- a/BootloaderCorePkg/Tools/ConfigEditor.py
+++ b/BootloaderCorePkg/Tools/ConfigEditor.py
@@ -624,6 +624,7 @@ class application(tkinter.Frame):
     def combo_select_changed(self, event):
         self.update_config_data_from_widget(event.widget, None)
         self.update_widgets_visibility_on_page()
+        self.update_page_scroll_bar(False)
 
     def edit_num_finished(self, event):
         widget = event.widget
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Tools/ConfigEditor.py
```diff
@@ -624,6 +624,8 @@ def update_widgets_visibility_on_page(self):
     def combo_select_changed(self, event):
         self.update_config_data_from_widget(event.widget, None)
         self.update_widgets_visibility_on_page()
+        self.frame_right.update()
+        self.conf_canvas.config(scrollregion=self.conf_canvas.bbox("all"))
```
mauricema, line 628: Maybe better to call update_page_scroll_bar() instead ?

If you want to keep the cursor at current selected item, we can add a parameter to update_page_scroll_bar() to control if we need to reset the yview position to 0 or not.

randylintw: Sure, added parameter in update_page_scroll_bar.

mauricema (approved):
Looks good to me.

PR #1483, opened 2022-02-11
StitchIfwi: Indicate failure through exit status

mikecrowe:
When run as part of an automated system it's important to ensure that
any failure is reported to the calling process. Writing an error message
and then exiting indicating success leads to difficult-to-diagnose
problems.


mauricema (approved):
Thank you for the change.  It looks good to me.

gdong1 (approved):

PR #1484, opened 2022-02-11
Add configuration data length check in CfgDataTool

mauricema:
This patch added configuration array data size check to ensure it
is DWORD aligned. It will prevent some unexpected errors later on
while merging the configuration data.


stalamudupula:
I have an incorrect alignment size in gpio config for ADL, but this change is not helping to print the exception. There is just a hang.

mauricema:
can you share you config so that I can take a look ?

stalamudupula (approved):
Verified this patch on ADL's gpio config. Works as expected.

Looks good to me.

PR #1487, opened 2022-02-14
[Common] Docker file iasl version update

koktong-ong:
Update the iasl version in Dockerfile to 20200925
Current iasl version in Dockerfile out-dated and
causing build error in Docker


mauricema (approved):
Looks good.

PR #1499, opened 2022-02-22
[ADLS] Fix bad DSO failure test case

sindhuragrandhi:
This patch fixes the test case where when a bad dso is
provided, it will revert back to the default dso settings.

TEST= Ran the test case successfully on ADLS board.


raghavag (approved):
Looks good to me

PR #1500, opened 2022-02-24
Enable Grahpics Console during firmware update

kokweich:
This patch enables graphics console when entering FWU payload.
FWU progress will show on both graphics console and serial port.

Signed-off-by: kokweich <kok.wei.chan@intel.com>


ywho79 (approved):

mauricema (approved):
Looks good to me.

PR #1505, opened 2022-02-25
Update VBT patching by dynamically locating block

### LINKED ISSUE #1504: [TGL] VBT patching error
On TGL platform, VBT patching does not work as expected.
It could not locate the correct VBT_GENERAL2_INFO block data offset.  As a result all fields for  VBT_GENERAL2_INFO are shifted.
For example:  VBT_GENERAL2_INFO.SizeChild_Struct might show up as 0.

mauricema:
Current VBT patching uses a static big structure VBT_TABLE_DATA to
parse the VBT data. It is not reliable because the sub structures
within VBT_TABLE_DATA might change between different revision. A
better solution is to look through each block data using the common
block header info. It can provide better back compatibility.  This
patch implemented this.

It fixed #1504.

### DIFF
--- a/Silicon/CommonSocPkg/Library/IgdOpRegionLib/IgdOpRegionLib.c
+++ b/Silicon/CommonSocPkg/Library/IgdOpRegionLib/IgdOpRegionLib.c
@@ -10,7 +10,7 @@
     VBT:        Video BIOS Table (OEM customizable data)
     IPU:        Image Processing Unit
 
-  Copyright (c) 2021, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2021 - 2022, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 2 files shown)


Silicon/CommonSocPkg/Library/IgdOpRegionLib/IgdOpRegionLib.c
```diff
@@ -138,19 +177,19 @@ UpdateVbt (
   if (VbtPtr != NULL) {
     DEBUG ((DEBUG_INFO, "VBT data found '%.20a'\n", VbtPtr->VbtHeader.Product_String));
     DEBUG((DEBUG_INFO, "Updating GOP VBT fields using GOP Configuration Lib.\n"));
+    VbtGen2Info = (VBT_GENERAL2_INFO *) LocateVbtBlockById ((UINT8 *)VbtPtr, VBT_BLK_ID_GENERAL2_INFO);
+    if (VbtGen2Info != NULL) {
+      for (Index = 0; Index < ChildStruct_MAX; Index++) {
+        ChildStructPtr[Index] = (CHILD_STRUCT *)(((UINT8 *)VbtGen2Info->Child_Struct) + Index * VbtPtr->VbtGen2Info.SizeChild_Struct);
```
gdong1, line 183: "VbtPtr->VbtGen2Info.SizeChild_Struct" should be changed to "VbtGen2Info ->SizeChild_Struct".
Basically we need try to avoid to use VbtPtr to direct access its member if possible since its member structure size might be changed.

mauricema: Yup. good catch!  Let us update it.

gdong1 (approved):

randylintw:
Verified on TGLU RVP.

PR #1507, opened 2022-02-27
Add support for PCI expansion rom resource allocation

mauricema:
Current SBL PCI enumeration does not allocate resource for PCI ROM
bar because SBL does not deal with option ROM at all. However, the
Linux kernel might expect the ROM bar resource to be allocated.
This patch introduces a static build configuration to allow support
PCI resource allocation for PCI ROM bar.

To enable this feature, please add following into the project
BoardConfig.py file:
  self._PCI_ENUM_FLAG_ALLOC_ROM_BAR = 1

By default, it will be disabled to keep the same behavior as before.

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2017 - 2020, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2022, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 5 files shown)


BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -513,6 +517,20 @@ GatherDeviceInfo (
     Offset = PciParseBar (PciIoDevice, Offset, BarIndex);
   }
 
+  //
+  // Find a available PCI bar slot for PCI expansion ROM
+  //
+  if (EnumPolicy->Flag.FlagAllocRomBar == 1) {
+    BarIndex = PCI_MAX_BAR - 1;
+    while (BarIndex > 0) {
+      if (PciIoDevice->PciBar[BarIndex].Length == PciBarTypeUnknown) {
```
gdong1, line 526: do you mean PciIoDevice->PciBar[BarIndex].BarType when comparing with PciBarTypeUnknown?

mauricema: Ahh, yes, you are right.
Let me fix it.

gdong1 (approved):

PR #1512, opened 2022-03-02
[ADL] Remove RTC logic in payload switching

koktong-ong:
Removed RTC logic, it is problematic and should not be used,
all other platforms are not using this logic and
we should remove this from the code


mauricema (approved):
Thanks for doing this.  Looks good to me.

PR #1521, opened 2022-03-16
[EHL] Fix IBECC error injection

koktong-ong:
Only enabled IehMode when IBECC error injection is disabled
Iehmode will cause failure in IBECC error injection test

### DIFF
--- a/Platform/ElkhartlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
+++ b/Platform/ElkhartlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
@@ -875,6 +875,8 @@ RtcInit (
   Bar     = MmioRead32 (MM_PCI_ADDRESS (0, PCI_DEVICE_NUMBER_PCH_PMC, PCI_FUNCTION_NUMBER_PCH_PMC, R_PMC_CFG_BASE)) & ~0x0F;
   PmConf1 = MmioRead8 (Bar + R_PMC_PWRM_GEN_PMCON_B);
 
+  RtcRead (R_RTC_IO_REGA);
+
   if ((PmConf1 & B_PMC_PWRM_GEN_PMCON_B_RTC_PWR_STS) != 0) {
     DEBUG ((DEBUG_INFO, "Initialize RTC with default values\n"));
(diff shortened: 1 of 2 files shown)


Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -1600,7 +1603,11 @@ UpdateFspConfig (
 
   if (mPchSciSupported) {
     Fspscfg->IsFusaSupported = 0x1;
-    Fspscfg->IehMode = 0x1;
+    if (MemCfgData != NULL) {
```
pqueirosint, line 1606: Should null check be more an ASSERT? or can memcfgdata be actually null?

in case if statement is used please consider add an else condition that sets
Fspscfg->IehMode = 0x1;
to keep same behavior as before.

Platform/ElkhartlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -875,6 +875,8 @@ RtcInit (
   Bar     = MmioRead32 (MM_PCI_ADDRESS (0, PCI_DEVICE_NUMBER_PCH_PMC, PCI_FUNCTION_NUMBER_PCH_PMC, R_PMC_CFG_BASE)) & ~0x0F;
   PmConf1 = MmioRead8 (Bar + R_PMC_PWRM_GEN_PMCON_B);
 
+  RtcRead (R_RTC_IO_REGA);
```
finikorg, line 878: Is this code belongs to other patch?

koktong-ong: For this patch

finikorg: I mean that can be put to separate commit with commit message for the change, but as is also OK.

randylintw:
Help KT to fix the checks error. 
Please move to use this: https://github.com/slimbootloader/slimbootloader/pull/1553

PR #1522, opened 2022-03-16
[ADLS] Fix Tcc issue caused due to incorrect UPD setting.

sindhuragrandhi:
Some of the RTCT table entries werent populated due to this
missing UPD setting. Hence, assigning it to the correct value.

TEST= Verified the fix on ADLS board.


Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
```diff
@@ -387,6 +387,10 @@ TccModePostMemConfig (
     FspsUpd->FspsConfig.TccCacheCfgBase = (UINT32)(UINTN)TccCacheconfigBase;
     FspsUpd->FspsConfig.TccCacheCfgSize = TccCacheconfigSize;
     DEBUG ((DEBUG_INFO, "Load Tcc Cache @0x%p, size = 0x%x\n", TccCacheconfigBase, TccCacheconfigSize));
+
+    if (IsPchS ()) {
```
gdong1, line 391: do we really need this check?

sindhuragrandhi: I will remove it later once I can validate on other ADL flavors.

gdong1: Inside FSP it should have such check (so this kind of UPD would not be used if TCC is not supported) from my understanding. OK, let's merge it firstly.

gdong1 (approved):

PR #1527, opened 2022-03-23
[ADL] Add a new platform

gdong1:
Add a new SO DDR5 platform ID 0x31.


kvparikh89:
@gdong1: I think there are more places that need a switch case. Ex: AcpiPlatform.c (in stage2).

gdong1:
> here are more places that need a switch case. Ex: AcpiPlatform.c (in stag
Yes, I believe there are more changes required to fully enable this platform, that could be added in a separate patch since these info is not available yet.

cshur (approved):

PR #1533, opened 2022-03-29
[RPLS] Update big core number default value

gdong1:
The default big core number was set to 8, it has to be overridden if
the CPU number is not 8. Had better set to 0xFF to all active big cores.
Also it mentioned the default value is 0xFF in the help message.


kvparikh89:
Does this resolve the reboot issue?

kvparikh89 (commented):
Looks good to me.

gdong1:
> Does this resolve the reboot issue?
yes, a reset would happen if the value doesn't match with CPU straps.

raghavag (approved):
looks good to me

PR #1542, opened 2022-04-05
Enhance GenContainer.py

gdong1:
During stitch phase, it would stitch failure if the data
is not signed and the data for the subregion is too small,
so enhance the tool to support the corner case.


cshur (approved):
Looks good

PR #1548, opened 2022-04-13
Add BootToShell configuration item

gdong1:
By default OsLoader payload shell is only available in debug build.
In some case customer also wants to have shell in release build.
So add a BootToShell configuration item to enable shell in release build.
By default BootToShell is set to 0. When it is set to 1 the shell behavior
would be same in release and debug build.

NOTE: user could add this line in platform dlt file to set BootToShell.
GEN_CFG_DATA.BootToShell  |1


kvparikh89 (approved):
Looks good. Please also update the documentation for this change.

PR #1554, opened 2022-04-21
Adl update4

gdong1:
minor changes to ADL

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/CpuInfoLib.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/CpuInfoLib.c
@@ -96,7 +96,7 @@ GetCpuSkuInfo (
   }
 
   if (!SkuFound) {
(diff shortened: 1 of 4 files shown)


Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/CpuInfoLib.c
```diff
@@ -95,7 +96,7 @@ GetCpuSkuInfo (
   }
 
   if (!SkuFound) {
-    DEBUG ((DEBUG_ERROR, "Unsupported CPU SKU, Device ID: 0x%02X, CPUID: 0x%08X!\n", CpuDid, CpuFamilyModel));
+    DEBUG ((DEBUG_INFO, "Unsupported CPU SKU, Device ID: 0x%02X, CPUID: 0x%08X!\n", CpuDid, CpuFamilyModel));
```
mauricema, line 99: don't we want this to be "DEBUG_ERROR" instead ?

gdong1: Yes, Better to keep this as ERROR level. will update it.

mauricema (approved):
looks good to me.

PR #1557, opened 2022-04-22
fix the wrong sblopen_dir value in StitchLoader.py and StitchIfwi.py

vincentchchen:
This patch fixed the warning "the imp module is deprecated in
favour of importlib." This patch can reduce the chance of setting
SBL_SOURCE when the SBL root directory is not named as "SblOpen"
or when the scripts are run from the path other than the SBL root
directory. Also note that os.path.abspath() returns the absolute
path relative to the current working directory instead of the real
path of __file__. So in StitchIfwi.py, sblopen_dir was incorrect
since the working directory had been changed before calling stitch().

### DIFF
--- a/Platform/AlderlakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/AlderlakeBoardPkg/Script/StitchIfwi.py
@@ -1,7 +1,7 @@
 ## @ StitchIfwi.py
 #  This is a python stitching script for Slim Bootloader ADL build
 #
-# Copyright (c) 2020 - 2021, Intel Corporation. All rights reserved. <BR>
+# Copyright (c) 2020 - 2022, Intel Corporation. All rights reserved. <BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 ##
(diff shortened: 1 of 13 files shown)


Platform/AlderlakeBoardPkg/Script/StitchIfwi.py
```diff
@@ -29,9 +29,11 @@
 # sign_bin_flag can be set to false to avoid signing process. Applicable for Btg profile 0
 sign_bin_flag = True
 
-sblopen_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../../../../', 'SblOpen')
+sblopen_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../'))
```
mauricema, line 32: Since it is common for all platforms, it might be better to create a function in IfwiUtility.py so that everybody can just call it.  It is easier to maintain in this way.

vincentchchen: This is somewhat a chicken-and-egg problem :), since we could import IfwiUtility module only after the SBL root is identified, where either the user explicitly specifies it with SBL_SOURCE or it is assumed four levels up relative to the path of StitchIfwi.py

mauricema: OK, your point is valid.   I think we can do this for now.  
Maybe later on we can consider a generic StitchIfwi.py for all platforms, and this script can use modules for platform specific customization.  Just similar to how we deal with BuildLoader.py for all platforms.

mauricema (approved):
Looks good to me.

gdong1 (approved):

PR #1565, opened 2022-04-29
[ADL] Configure MTRR to enable full flash region cache

lsubash:
FSPT doesn't allow to enable full flash code cache.
Bootloader configures MTRR in non BTG cases to avoid
performance penalty.

This fixes the ADL boot from BP1 partition.


raghavag (approved):
looks good to me

PR #1566, opened 2022-05-02
[ADLN] Add GPIO configuration from Cfg Data

sindhuragrandhi:
Add GPP_T as part of the base cfg in order to use for
other ADL flavors like ADL PS.


raghavag (approved):
looks good to me

PR #1569, opened 2022-05-09
Support Capsule Update of Existing uCode FW

sean-m-mcginn:
- Adds code to support the update of existing uCode FW via capsule
- Does not add code to add new/remove existing uCode FW


BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -246,7 +257,15 @@ def Main():
...
     microcode_inf = os.path.join(sbl_dir, 'Silicon', silicon_pkg_name, 'Microcode', 'Microcode.inf')
-    CopyBins (ucode_repo_dir, sbl_dir, microcode_inf)
+    microcode_bins = CopyBins (ucode_repo_dir, sbl_dir, microcode_inf)
+
+    # Add microcode padding
+    if microcode_bins != None and len(sys.argv) > 4:
+        slot_size = int(sys.argv)
+        for microcode_bin in microcode_bins:
+            microcode_bin = os.path.join(sbl_dir,microcode_bin[1])
+            DoAddPaddingData (microcode_bin, slot_size)
```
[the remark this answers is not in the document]
sean-m-mcginn: Wouldn't this sort of padding occur as part of the call to align_pad_file() within create_bootloader_image() of BuildLoader.py? This code here only pads each *.mcb file before they get combined into a *.bin file. That final *.bin file also gets padded.

lsubash, line 267: Does DoAddPaddingData gets invoked for slot_size zero cases also?

lsubash (commented):
Update the details of SLOT_SIZE board config in commit details.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -220,17 +221,27 @@ def BuildFspBins (fsp_dir, sbl_dir, fsp_inf, silicon_pkg_name, flag):
...
+    elif len(Binary) < SlotSize:
+      Binary = Binary + b'\xff' * (SlotSize - len(Binary))
+
+  with open(File, "wb") as fd:
+    fd.write(Binary)
+
 def Main():
 
-    if len(sys.argv) < 3:
+    if len(sys.argv) < 4:
```
lsubash, line 238: What is the argument here?

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -246,7 +257,15 @@ def Main():
         CopyBins (fsp_repo_dir, sbl_dir, fsp_inf)
 
     microcode_inf = os.path.join(sbl_dir, 'Silicon', silicon_pkg_name, 'Microcode', 'Microcode.inf')
-    CopyBins (ucode_repo_dir, sbl_dir, microcode_inf)
+    microcode_bins = CopyBins (ucode_repo_dir, sbl_dir, microcode_inf)
+
+    # Add microcode padding
+    if microcode_bins != None and len(sys.argv) > 4:
+        slot_size = int(sys.argv)
```
lsubash, line 264: Looks its reusing the argument. Add a new parameter here.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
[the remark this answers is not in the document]
gdong1: Existing padding has no relation with SLOT_SIZE. Since this patch is to add SLOT concept, I think it should work regarding UserExtensions section exists or not in the INF file.

BuildLoader.py
```diff
@@ -356,12 +355,13 @@ def update_fit_table (self):
             while ucode_offset < len(rom):
                 ucode_hdr = UCODE_HEADER.from_buffer(rom, ucode_offset)
                 if ucode_hdr.header_version == 1:
-                    if ucode_hdr.total_size:
-                        ucode_size = ucode_hdr.total_size
+                    u_code_images.append(ucode_offset)
+                    if (hasattr (self._board, "SLOT_SIZE")):
+                        ucode_offset += self._board.SLOT_SIZE
```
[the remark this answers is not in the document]
sean-m-mcginn: The motivation for adding a slot size member to the board object is because I only want the "slotting" to occur only when explicitly indicated. I worry that this might introduce unintended side effects to legacy platforms like APL where this feature does not need to be introduced. Also, adding such a slot size member mirrors BIOS.
Adding a check of new vs. old slot size in the FW update code is non-trivial. This is because we would have to manually measure each uCode in the new image and the existing image. There already is a check on the sizes of the new uCode region vs. the existing uCode region, so other regions are safe. If the slot sizes are unaligned, it just simply won't boot.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
sean-m-mcginn: Yes, but in such a case, it does not add any padding bytes.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
sean-m-mcginn: Here are the arguments given to this file, in order:
1) python file name for building binaries (e.g. C:\SblPlatform\SblOpen\BootloaderCorePkg\Tools\PrepareBuildComponentBin.py)
2) working directory (e.g. C:\SblPlatform\SblOpen)
3) silicon package name (e.g. TigerlakePkg)
4) build mode (e.g. /r) 
5) ucode slot size (e.g. 114688)

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
sean-m-mcginn: Should now be fixed.

BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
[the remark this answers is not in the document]
sean-m-mcginn: Should now just pad all *.mcb files, regardless of the UserExtensions section.

lsubash (approved):

gdong1 (approved):

PR #1570, opened 2022-05-10
Add support for getting csme boot time perf data

raghavag:
This patch added support to get csme boot time performance
data and display it in perf command and also before booting
to linux.

Introduced a board config option BOOT_PERFORMANCE_MASK to control
PcdBootPerformanceMask, BIT 2 now enables printing CSME boot
performance data.


Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -94,6 +93,48 @@ BuildOsConfigDataHob (
...
+    DEBUG ((DEBUG_ERROR, "Obtaining CSME Boot performance data failed with status: %r\n", Status));
+    return Status;
+  } else {
+    DEBUG ((DEBUG_INFO, "Found CSME Boot performance data!\n"));
+  }
+
+  CsmeBootTimeData->Revision = 1;
+  CsmeBootTimeData->BootDataVersion = EarlyBootDataVersion;
+  CsmeBootTimeData->BootDataLength = EarlyBootDataLength;
+  CopyMem (CsmeBootTimeData->BootPerformanceData, EarlyBootData, EarlyBootDataLength * sizeof (UINT32));
```
gdong1, line 133: The HOB was built with 64 records by default.
So before copying data to the HOB, we had better have a check the hob length to avoid potential buffer overflow.

raghavag: Updated patch to send allocated hob length via CsmeBootTimeData and verify the length after reading data from csme

gdong1 (approved):

PR #1576, opened 2022-05-18
[ADL] Fix HsPhyInit failure

sindhuragrandhi:
HsPhy feature is only applicable for ADLS and not for others.
Adding a condition to apply only for it.


raghavag (approved):
looks good to me

PR #1577, opened 2022-05-18
Leverage EDKII Build System to Perform uCode Slotting

sean-m-mcginn:
This change leverages the --align parameter on GenFw.exe to perform uCode slotting, when configured

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -348,9 +348,9 @@ class Build(object):
         num_fit_entries = 0
         if self._board.UCODE_SIZE > 0:
             ucode_base = self._board.UCODE_BASE
-            ucode_offset = ucode_base - base;
-            if (ucode_offset < 0):
-                raise Exception ('  UCODE %x\n  UCODE address (0x%08X) out of range' % (base, ucode_base))
+            ucode_offset = ucode_base - base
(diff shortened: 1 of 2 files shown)


BuildLoader.py
```diff
@@ -361,6 +361,8 @@ def update_fit_table (self):
                     u_code_images.append(ucode_offset)
                     num_fit_entries += 1
                     if (hasattr (self._board, "UCODE_SLOT_SIZE")):
+                        if ucode_hdr.total_size > self._board.UCODE_SLOT_SIZE:
+                            raise Exception ('UCODE total size (0x%08X) greater than UCODE slot size' % (ucode_hdr.total_size, self._board.UCODE_SLOT_SIZE))
```
gdong1, line 365: ucode_hdr.total_size should be same with self._board.UCODE_SLOT_SIZE, right?

sean-m-mcginn: The uCode header is pulled directly from a uCode file -- we don't have direct control over it. It is the size of the actual uCode file, without factoring in padding.

gdong1: It makes sense if it is not padded yet.

gdong1 (approved):

PR #1581, opened 2022-05-23
[ADL] Update BDF for UFS device

raghavag:
This patch fixed BDF for UFS device, bus number and function
number are improperly ported and this patch fixed the issue.


kvparikh89 (approved):
Looks good.

PR #1585, opened 2022-05-26
[WIP] [ADLPS] Process GPIO from Cfg Data

sindhuragrandhi:
- Updated GPIO table to match BIOS PV ER3 release.
- Disable SCI for D13 and E00.
- Process the GPIO table from dlt file instead
of the hard-coded table. The mGpioTablePostMemAdlPsDdr5Rvp
is only for reference purposes.
- Move DEBUG CODE END to later part of the GPIO function in
order to add GPIO prints when required.


gdong1 (approved):

kvparikh89:
PR is still WIP. Hence reverted

PR #1599, opened 2022-06-20
Improve BlockIo test code

xuewenxi-intel:
Fixed build and running issues.

Signed-off-by: Gavin Xue <gavin.xue@intel.com>

### DIFF
--- a/PayloadPkg/OsLoader/BlockIoTest.c
+++ b/PayloadPkg/OsLoader/BlockIoTest.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2017 - 2018, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2022, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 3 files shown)


PayloadPkg/OsLoader/BlockIoTest.c
```diff
@@ -89,7 +85,7 @@ TestDevBlocks (
   // Test GetInfo, read and write
   for (Index = 0; Index < 7; Index++ ) {
     ZeroMem (&BlockInfo, sizeof (BlockInfo));
-    Status = DevBlockFunc.GetInfo (Index, &BlockInfo);
+    Status = DevBlockFunc.GetInfo (OsBootOption->HwPart, &BlockInfo);
```
gdong1, line 88: From the BlockIO test logical, it is tried to test HW partition from 0 to 7. If change to OsBootOption->HwPart, then only one HW partition is tested and the for loop is not required. In most cases the HW partition is no more than 7.  I think we could keep current logic to test HW partition 0~7 in case OsBootOption->HwPart is not set correctly.

xuewenxi-intel: Hi Guo,
Thanks for you comments. Recently I am trying to check eMMC write function in SBL and found the running error, so I summitted this patch. If this is the code logic, the error should be expect. I will update the patch just fix build issue.

PR #1626, opened 2022-07-15
[EHL] Update StitchIfwiConfig.py

vincentchchen:
- add required flash image region 5=EC for new FIT (15.40.26.2632)
- remove 'sata' stitch option
- add only items whose features are enabled by default in replace_list

### DIFF
--- a/Platform/ElkhartlakeBoardPkg/Script/StitchIfwiConfig.py
+++ b/Platform/ElkhartlakeBoardPkg/Script/StitchIfwiConfig.py
@@ -1,7 +1,7 @@
 ## @ StitchIfwi.py
 #  This is an IFWI stitch config script for Slim Bootloader
 #
-# Copyright (c) 2019 - 2020, Intel Corporation. All rights reserved. <BR>
+# Copyright (c) 2019 - 2022, Intel Corporation. All rights reserved. <BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 ##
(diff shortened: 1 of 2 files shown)


Platform/ElkhartlakeBoardPkg/Script/StitchIfwiConfig.py
```diff
@@ -74,49 +74,29 @@ def get_platform_sku():
...
+      # need to set ENABLE_PRE_OS_CHECKER = 1 in BoardConfig.py
+      ('IFWI/BIOS/NRD/IPFW/POSC', 'IPFW/PreOsChecker.bin',   'dummy',  'KEY_ID_CONTAINER_COMP_RSA3072', 0 ), # Pre-OS Checker
+
+      # need to set ENABLE_PSEFW_LOADING = 1 in BoardConfig.py
+      ('IFWI/BIOS/NRD/IPFW/PSEF', 'IPFW/PseFw.bin',          'lz4',    'KEY_ID_CONTAINER_COMP_RSA3072', 0 ), # OSE FW
+
+      # need to set ENABLE_TSN = 1 in BoardConfig.py
+      # ('IFWI/BIOS/NRD/IPFW/TSIP', 'IPFW/PseTsnIpConfig.bin', 'lz4',    'KEY_ID_CONTAINER_COMP_RSA3072', 0 ), # PSE TSN IP
+      # ('IFWI/BIOS/NRD/IPFW/TSNC', 'IPFW/TsnConfig.bin',      'lz4',    'KEY_ID_CONTAINER_COMP_RSA3072', 0 ), # TSN Config
+      # ('IFWI/BIOS/NRD/IPFW/TMAC', 'IPFW/TsnMacAddr.bin',     'lz4',    'KEY_ID_CONTAINER_COMP_RSA3072', 0 ), # TSN MAC Address
```
gdong1, line 90: Any reason to remove the file exist check and comment these lines out?
SBL supports to build these binaries into SBL image during build time and stitch time without changing stitch script.  I think we could update StitchIfwi tool if you see an issue (e.g. give a warning if the component is not found in SBL image while replacing component)

vincentchchen: Take TCC as an example,
if TCC is enabled during build time, but the component is not found in SBL image, then user can still replace it during stitch time.
If TCC is not enabled during build time, but the component is put in the specified IPFW path, then stitch process will fail.
To avoid this case, I would rather that user can explicitly modify the replace_list and be aware of the ENABLE_TCC flag being set, and then put the binary in the specified path.
Alternative way is to use stitch -o options to control the replace_list, so no stitch script change would happen.

gdong1: How about we enhance StitchIfwi.py to give a warning for the case "If TCC is not enabled during build time, but the component is put in the specified IPFW path, then stitch process will fail"?  StitchIfwi could know if a component exists or not in SBL image, it could skip to replace a component if that component doesn't exist.

vincentchchen: Instead of raise exception, added the code to skip the component replacement if the component file is not existing or if the component is not locatable in SBL image.
Also print the skipped components so user would be aware of.

example output messages:
IFWI Stitching completed successfully !
Boot Guard Profile: FVME
**Skipped replacement components: ['IFWI/BIOS/NRD/IPFW/TSIP', 'IFWI/BIOS/NRD/IPFW/TSNC', 'IFWI/BIOS/NRD/IPFW/TMAC']**
IFWI image: sbl_ifwi_ehl.bin

gdong1 (approved):

PR #1629, opened 2022-07-18
[Tools] Enhance GpioDataConvert to support GPD group

stanleyintel:
The patch adds support of "GPD" group when parsing a .h header file.
It also
  - adds supports for Pch-N,
  - fixes leading whitespaces in h file

Test commands:

      # h to dlt
      $ python Platform/CommonBoardPkg/Tools/GpioDataConvert.py \
          -cf Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py \
          -if example_adls.h -of dlt -p s \
          -o out_adls.dlt

      # h to yaml
      $ python Platform/CommonBoardPkg/Tools/GpioDataConvert.py \
          -cf Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py \
          -if example_adls.h -of yaml -p s \
          -o out_adls.yaml

      # yaml to h
      $ python Platform/CommonBoardPkg/Tools/GpioDataConvert.py \
          -cf Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py \
          -if out_adls.yaml -of h -p s \
          -o out_h_from_yaml.h

      # dlt to h
      $ python Platform/CommonBoardPkg/Tools/GpioDataConvert.py \
          -cf Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py \
          -if out_adls.yaml -of h -p s \
          -o out_h_from_dlt.h
      # check: compare example_adls.h and out_h_from_dlt.h

      # yaml to txt
      $ python Platform/CommonBoardPkg/Tools/GpioDataConvert.py \
          -cf Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py \
          -if out_adls.yaml -of txt -p s \
          -o out_txt_from_yaml.txt

      # dlt to txt
      $ python Platform/CommonBoardPkg/Tools/GpioDataConvert.py \
          -cf Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py \
          -if out_adls.dlt -of txt -p s \
          -o out_txt_from_dlt.txt

### DIFF
--- a/Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py
+++ b/Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py
@@ -12,31 +12,62 @@
 # in a platform specific Gpio library.
 #
 
+# mPchLpGpioGroupInfo in Silicon/AlderlakePkg/Library/GpioSiLib/GpioSiLib.c
 grp_info_lp = {
     # Grp     Index
     'GPP_B' : [ 0x0],
     'GPP_T' : [ 0x1],
     'GPP_A' : [ 0x2],
     'GPP_R' : [ 0x3],
+    # 'SPI' : [ 0x4],
+    'GPD'   : [ 0x5],
     'GPP_S' : [ 0x6],
     'GPP_H' : [ 0x7],
     'GPP_D' : [ 0x8],
     'GPP_U' : [ 0x9],
+    # 'VGPIO' : [ 0xA],
     'GPP_C' : [ 0xB],
     'GPP_F' : [ 0xC],
+    # 'HVCMOS' : [ 0xD],
     'GPP_E' : [ 0xE],
 }
 
+# mPchNGpioGroupInfo in Silicon/AlderlakePkg/Library/GpioSiLib/GpioSiLib.c
+grp_info_n = {
+    # Grp     Index
+    'GPP_B' : [ 0x0],
+    'GPP_T' : [ 0x1],
+    'GPP_A' : [ 0x2],
+    'GPP_R' : [ 0x3],
+    # 'SPI' : [ 0x4],
+    'GPD'   : [ 0x5],
+    'GPP_S' : [ 0x6],
+    'GPP_H' : [ 0x7],
+    'GPP_D' : [ 0x8],
+    'GPP_I' : [ 0x9],
+    # 'VGPIO' : [ 0xA],
+    'GPP_C' : [ 0xB],
+    'GPP_F' : [ 0xC],
+    # 'HVCMOS' : [ 0xD],
+    'GPP_E' : [ 0xE],
+}
+
+# mAdlPchSGpioGroupInfo in ./Silicon/AlderlakePkg/Library/GpioSiLib/GpioSiLib.c
 grp_info_s = {
     # Grp     Index
     'GPP_I' : [ 0x0],
     'GPP_R' : [ 0x1],
     'GPP_J' : [ 0x2],
+    # 'VGPIO' : [ 0x3],
+    # 'VGPIO_0' : [ 0x4],
     'GPP_B' : [ 0x5],
     'GPP_G' : [ 0x6],
     'GPP_H' : [ 0x7],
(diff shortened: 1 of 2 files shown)


Platform/AlderlakeBoardPkg/Script/GpioDataConfig.py
```diff
@@ -12,46 +12,68 @@
...
+# Silicon/AlderlakePkg/Include/Register/PchRegsGpioVer2.h
 grp_info_lp = {
     # Grp     Index
     'GPP_B' : [ 0x0],
     'GPP_T' : [ 0x1],
     'GPP_A' : [ 0x2],
     'GPP_R' : [ 0x3],
-    'GPP_S' : [ 0x6],
+    'GPD'   : [ 0x4],
+    'GPP_S' : [ 0x5],
```
stalamudupula, line 23: Hi, the original commit had this table updated based on Ver4.h, or GpioSiLib.c tables 'mPchLpGpioGroupInfo' and 'mAdlPchSGpioGroupInfo'. Could you please double check whether we have to use Ver4 or Ver2.

If we have to use Ver2, then this patch set is correct, but GpioSiLib.c has to change.
If not, GpioSiLib.c is correct, and  this file should not be touched except for adding a dictionary for ADL-N.

sindhuragrandhi: Code uses Ver2 mapping and 'mPchLpGpioGroupInfo' also uses Ver2 mapping. In this file original patch, 'grp_info_lp' dictionary didnt consider SPI as No register (0xFF), instead it counted as 0x4 and hence GPP_S was counted up as 0x6 which is incorrect. It should actually be 0x5 for GPP_S.

But one thing that doesnt match between mPchLpGpioGroupInfo and PchRegsGpioVer2.h is with **GPP_C** and **GPP_F**. According to mPchLpGpioGroupInfo, if we count, GPP_C should be 0xA and GPP_F should be 0xB. But if we follow PchRegsGpioVer2.h, GPP_C is defined as 0xB while GPP_F is defined as 0xA.

For this tool to work correctly, should we match with mPchLpGpioGroupInfo or the definitions for these groups from the header file? Clearly, they are not completely matching up.

sindhuragrandhi: I syned with Sai again on this topic. It seems like for this tool to work correctly, we need to follow the indexing as per the mPchLpGpioGroupInfo[] from GpioSiLib.c, and not what is defined in PchRegsGpioVer2.h for Lp series. Similarly for S and N series. And SPI should still be considered as one of the indexes to be counted in. Please change these dictionaries according to the GpioGroupInfo[]. If SPI is not needed, skip adding it to the dictionary here but count it as +1.

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
```diff
@@ -595,23 +601,32 @@ def get_sbl_dws_from_txt (parts, pad_name, sbl_dw0, sbl_dw1, gpio_cfg_file):
...
-    if not 'GPP_' in parts[0]:
-        return '', 0x0, 0x0
     splitdata = parts[0].split('_')
     pin_number = splitdata[len(splitdata)-1]
     pin_number = normalize_pad_name (pin_number)
-    pad_name = 'GPP_' + pin_number
+    if 'GPP_' in parts[0]:
+      pad_name = 'GPP_' + pin_number
+    elif 'GPD' in parts[0]:
+      pad_name = pin_number
```
stalamudupula, line 626: pad_name = pin_number --- > pad_name = 'GPD' + pin_number

stalamudupula (approved):
Looks good to me.

Platform/CommonBoardPkg/Tools/GpioDataConvert.py
stalamudupula: makes sense.

gdong1 (approved):

PR #1632, opened 2022-07-19
Move SG1B and UCOD into TS region

sean-m-mcginn:
Allow for SG1B and UCOD to be moved into the TS region
The actual movement of these regions will occur in BoardConfig*.py files in the Intel internal SBL platform repo

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -264,6 +264,7 @@ class BaseBoard(object):
         self.BUILD_ARCH            = ''
         self.KEYH_SVN              = 0
         self.CFGDATA_SVN           = 0
+        self.BUILD_IDENTICAL_TS    = 0
 
         self.RTCM_RSVD_SIZE        = 0xFF000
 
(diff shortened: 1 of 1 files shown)


lsubash (approved):
looks good.

gdong1:
could you add more comments in the commit message how this patch is related with patch title "Move KEYH, SG1B, and UCOD into TS region"?

sean-m-mcginn:
> could you add more comments in the commit message how this patch is related with patch title "Move KEYH, SG1B, and UCOD into TS region"?

Done

sean-m-mcginn:
@gdong1 Is this one good to be merged?

BuildLoader.py
```diff
@@ -264,6 +264,7 @@ def __init__(self, *args, **kwargs):
         self.BUILD_ARCH            = ''
         self.KEYH_SVN              = 0
         self.CFGDATA_SVN           = 0
+        self.BUILD_IDENTICAL_TS    = 0
```
gdong1, line 267: TS is always same for A and B. From its usage maybe we could change "BUILD_IDENTICAL_TS" to "REBASE_STAGE1B" to make this setting more clear?

sean-m-mcginn: Right now, this flag is only to control the rebasing of SG1B. However, in an upcoming patch, this flag will also control the patching of the flash map with the boot partition. I just didn't add this change here because it also requires a *.c code change.

sean-m-mcginn:
@gdong1 Is this okay to be merged?

gdong1 (approved):

PR #1636, opened 2022-07-22
Add shell command to list USB devices

arghyabi:
New shell command 'usbdev' added to enumerate USB bus and list down
all the USB devices that are found on the bus.

Signed-off-by: Biswas Arghya <arghya.biswas@intel.com>


gdong1:
If USB is in the boot option, all the USB devices will be printed while trying to boot from USB.  I am not sure if we really need to add a shell command. Maybe you could give your usage case on why we need it?

gdong1 (approved):

PR #1640, opened 2022-07-25
FWU: refine the reboot timing for different payload combinations

vincentchchen:
- If BIOS update is followed by any payload,
  reboot to ensure the update is completed.
- Before processing CMDI payload, ensure CSME update has
  taken effect to prevent {OEMKEYREVOCATION} command failure.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -307,7 +307,6 @@ ValidThenDecFwuRetryCount (
   }
 }
 
-
 /**
   Get state machine flag from flash.
 
@@ -607,7 +606,7 @@ UpdateStatus (
   //
   for (Count = 0; Count < MAX_FW_COMPONENTS; Count ++) {
     if (FwUpdCompStatus[Count].HardwareInstance == Signature) {
-      DEBUG((DEBUG_VERBOSE, "FOund the component to update status\n"));
+      DEBUG((DEBUG_VERBOSE, "Found the component to update status\n"));
       break;
     }
   }
@@ -914,7 +913,6 @@ ProcessCapsule (
 {
   UINT16                        Count;
   UINT16                        i;
-  UINT16                        TotalPayloadCount;
   UINT16                        PayloadItemCount;
   EFI_STATUS                    Status;
   UINT32                        FwUpdStatusOffset;
@@ -975,6 +973,7 @@ ProcessCapsule (
     FwUpdStatus.Version = FW_UPDATE_STATUS_VERSION;
     FwUpdStatus.Length = sizeof(FW_UPDATE_STATUS);
     FwUpdStatus.RetryCount = InitFwuRetryCount ();
+    FwUpdStatus.CsmeNeedReset = CSME_NEED_RESET_INIT;
 
     //
     // Save the current capsule signature into flash
@@ -1020,15 +1019,13 @@ ProcessCapsule (
     return EFI_NOT_FOUND;
   }
 
-  TotalPayloadCount = (UINT16)(CapHeader->EmbeddedDriverCount + CapHeader->PayloadItemCount);
-  if (CapHeader->PayloadItemCount > MAX_FW_COMPONENTS) {
+  // Get the number of payloads
+  PayloadItemCount = CapHeader->PayloadItemCount;
+  if (PayloadItemCount > MAX_FW_COMPONENTS) {
     DEBUG((DEBUG_ERROR, "# of payloads(%d) in capsule exceeds max(%d). The excess payloads will be skipped\n",
-          TotalPayloadCount, MAX_FW_COMPONENTS));
+          PayloadItemCount, MAX_FW_COMPONENTS));
     PayloadItemCount = MAX_FW_COMPONENTS;
-  } else {
-    PayloadItemCount =  CapHeader->PayloadItemCount;
   }
-
 
   //
   // Zero out memory for component structures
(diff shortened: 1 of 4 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -1133,12 +1132,20 @@ FindImage (
     return EFI_NOT_FOUND;
   }
 
+  // Get the number of payloads
+  PayloadItemCount = CapHeader->PayloadItemCount;
+  if (PayloadItemCount > MAX_FW_COMPONENTS) {
+    DEBUG((DEBUG_ERROR, "# of payloads(%d) in capsule exceeds max(%d). The excess payloads will be ignored\n",
+          PayloadItemCount, MAX_FW_COMPONENTS));
+    PayloadItemCount = MAX_FW_COMPONENTS;
```
stanleyintel, line 1140: FindImage does not need to check for MAX_FW_COMPONENTS. 
The MAX_FW_COMPONENTS is defined for the max # of entries (FW_UPDATE_COMP_STATUS) in RSVD. It has a upper bound because of the size of RSVD. 

However, the main goal of FindImage() to find a payload in capsule. A payload may locate after MAX_FW_COMPONENTs, and it should be still searchable. For example, CSMD may locate after MAX_FW_COMPONENTS, but it should be still valid.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -1330,14 +1382,22 @@ InitFirmwareUpdate (
...
+    DEBUG((DEBUG_ERROR, "# of payloads(%d) in capsule exceeds max(%d). The excess payloads will be skipped\n",
+          PayloadItemCount, MAX_FW_COMPONENTS));
+    PayloadItemCount = MAX_FW_COMPONENTS;
+  }
+
   //
   // Loop through the components to perform update
   //
-  for (Count = 0; Count < MAX_FW_COMPONENTS; Count ++) {
+  for (Count = 0; Count < PayloadItemCount; Count ++) {
```
stanleyintel, line 1397: Is that required?  An unused FwUpdCompStatus[Count].UpdatePending == 0.

Assume MAX_FW_COMPONENTS is 10000, it will save certain cpu time, but for single-digit MAX_FW_COMPONENT, the change in line 1385 ~ 1393 adds more code complexity than its benefit.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -975,6 +974,7 @@ ProcessCapsule (
     FwUpdStatus.Version = FW_UPDATE_STATUS_VERSION;
     FwUpdStatus.Length = sizeof(FW_UPDATE_STATUS);
     FwUpdStatus.RetryCount = InitFwuRetryCount ();
+    FwUpdStatus.CsmeNeedReset = 0xF;
```
stanleyintel, line 977: how about giving it the value (0xf) a name. And why not 0xff?

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -1217,11 +1224,54 @@ ApplyFwImage (
...
+          }
+        }
       }
     }
     break;
   case FW_UPDATE_COMP_CMD_REQUEST:
+    {
+      // If CSME is updated, the system needs a reset for the changes to take effect.
+      // Otherwise, the CMDI {OEMKEYREVOCATION} command might fail.
+      EFI_STATUS  UpdStatus;
```
stanleyintel, line 1251: a separate function will be clear.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
vincentchchen: In ProcessCapsule(), the payloads in a capsule being traversed are limited to the first MAX_FW_COMPONENTS payloads. The payloads beyond are ignored. All the traversed payloads are captured in FwUpdCompStatus[ ]. So there is no way to find a payload located after MAX_FW_COMPONENTS.

stanleyintel: Incorrect. Imagine CSME + CSMD as an example. CSME is within MAX_FW_COMPONENTS, but CSMD is placed outside of MAX_FW_COMPONENTS. This kind of placement is still valid. Unless you want to disallow this placement.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
vincentchchen: My idea is to process only the number of PayloadItemCount payloads.
Consider the situation when there is only one BIOS payload (means PayloadItemCount < MAX_FW_COMPONENTS), then we may need to check the UpdatePending status of the next FwUpdCompStatus[Count] to know if the BIOS payload is the final one to process.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
vincentchchen: Because it is of size UINT8. And you are right if adding more comments to describe the meaning of the magic number.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
stanleyintel: it is not neat. you can think twice about my comment/example left above.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
stanleyintel: Just because it is UINT8, why use 0xF here?

PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
```diff
@@ -1091,6 +1091,80 @@ UpdateSblComponent (
...
+WriteCsmeNeedResetFlag (
+  IN  UINT8  Value
+  )
+{
+  EFI_STATUS  Status;
+  UINT32      FwUpdStatusOffset;
+  UINT8       CurrVal;
+
+  CurrVal = ReadCsmeNeedResetFlag();
+  if (Value > CurrVal) {
```
stanleyintel, line 1149: because CSME_NEED_RESET_<state> only uses the lower 4 bits of FW_UPDATE_STATUS.CsmeNeedReset, it implies the high 4 bits are reserved for future. Thus, it is better to either use a MASK before comparison or redefine the  FW_UPDATE_STATUS.CsmeNeedReset as a bit-field member. Similarly, when update the field, also need to reserve current high 4 bits. 

Note: it is also impossible to reduce the member to 3-bit width for 4 states: 111, 110, 100, 000.

vincentchchen: Thanks for figuring out it. Simply re-defined CSME_NEED_RESET_$state to use 8-bit values (0xFF, FE, FC, F8) to accommodate it.

gdong1 (approved):

PR #1643, opened 2022-07-27
[ADL-N/P/PS] Workaround for S4 resume when S0ix disabled.

bejeanmo:
Remove OS version check in PEPD _STA method. This is a change to the intelpep.sys ASL device (PEPD) for low power idle support. Previously, _STA would always return present/enabled for Win10 even if S0ix config flag was disabled. Changed version always returns status based on S0ix flag. Without this workaround, S4 resume was failing on ADL-N/P/PS.


sindhuragrandhi:
Looks good to me.

gdong1 (approved):

PR #1644, opened 2022-07-27
[ADL-P] Revert USB4 CM mode to FW CM only.

bejeanmo:
Previous setting of SW-first CM mode was causing a hang during S3/S4 entry. Likely this setting requires additional ACPI/driver support which is missing.


sindhuragrandhi:
Looks good to me.

gdong1 (approved):

PR #1648, opened 2022-07-28
[ADL] Fix for THC1 PEP constraints appearing when device is disabled.

bejeanmo:
FSP-S UPD format for THC port assignment was updated, but PEP logic
was using the old format, and causing the PEP constraint to always be
enabled. This issue didn't seem to be causing a problem, but resulted
in noncompliant low power idle constraints.


sindhuragrandhi:
Looks good to me.

gdong1 (approved):

PR #1665, opened 2022-08-18
[ADLPS] CRB PV release update

koktong-ong:
Update BoardID and PlatformID for CRB board
Added ddi config for CRB board
Update FSPM UPD due to common value across all sku

### DIFF
--- a/Platform/AlderlakeBoardPkg/Include/PlatformBoardId.h
+++ b/Platform/AlderlakeBoardPkg/Include/PlatformBoardId.h
@@ -15,7 +15,7 @@ Defines Platform BoardIds
 #define FlavorDesktop                       0x2
 
 // AlderLake-N Boards
-#define BoardIdAdlNDdr5Crb                            0x0A // MiniITX is CRB
+#define BoardIdAdlNDdr5Crb                            0x0B // MiniITX is CRB
 #define PLATFORM_ID_ADL_N_DDR5_CRB                    0x0A
 
 #define BoardIdAdlNLp5Rvp                             0x07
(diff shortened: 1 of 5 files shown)


Platform/AlderlakeBoardPkg/Include/PlatformBoardId.h
```diff
@@ -25,8 +25,8 @@ Defines Platform BoardIds
 #define BoardIdAdlPSDdr5Rvp                           0x0B
 #define PLATFORM_ID_ADL_PS_DDR5_RVP                   0x0B
 
-#define BoardIdAdlPSDdr5Crb                           0x0D // MiniITX is CRB
-#define PLATFORM_ID_ADL_PS_DDR5_CRB                   0x0D
+#define BoardIdAdlPSDdr5Crb                           0x0C // MiniITX is CRB
```
gdong1, line 28: what's the value read from SMBUS for PS board? Often this value would not change once we know the value, not sure  why changes BoardIdAdlPSDdr5Crb here.

koktong-ong: the value read from SMBUS is 0xD but due to the offset workaround in line 125 in EcSupport.c
*BoardId = (SmbusInputPort1Info.InputPort1Fields.BoardId - ADL_SMBUS_BOARDID_OFFSET);
The reason with this workaround is we seeing same conflict (same BoardID) between ADLN CRB and ADLPS RVP
https://github.com/slimbootloader/slimbootloader/commit/0dc5b1da43db0232098b15ecb55a090903b874ee

koktong-ong: Updated the patch as per discussion

koktong-ong:
@gdong1 
Updated the patch as per discussion

gdong1 (approved):

PR #1673, opened 2022-08-30
Tcc flag linx

akshathatc:


Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
```diff
@@ -58,6 +58,7 @@ GetCpuStepping(
   return ((CPU_STEPPING) (Eax.Uint32 & CPUID_FULL_STEPPING));
 }
 
+#if FeaturePcdGet (PcdTccEnabled)
```
sindhuragrandhi, line 61: This condition is a duplicate. Same condition already exists when it is being called.

Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
```diff
@@ -319,6 +319,7 @@ GetMaxCpuPciePortNum (
   }
 }
 
+#if FeaturePcdGet (PcdTccEnabled)
```
sindhuragrandhi, line 322: This condition is already checked when the function is being called? Why do you have to build out completely?

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
akshathatc: It was added to fix build errors caused due to Tcc FSP UPDs which are used in function definition.

Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
akshathatc: It was added to fix build errors caused due to Tcc FSP UPDs which are used in function definition.

sindhuragrandhi (approved):

gdong1 (approved):

PR #1675, opened 2022-08-31
Stop TCO timer in beginning of FWU PLD

sean-m-mcginn:
Stop TCO timer in beginning of FWU PLD
to ensure it is started without error

### DIFF
--- a/BootloaderCorePkg/Stage2/Stage2.c
+++ b/BootloaderCorePkg/Stage2/Stage2.c
@@ -289,8 +289,9 @@ NormalBootPath (
   PrintStackHeapInfo ();
   DEBUG_CODE_END ();
 
-  // Stop timer so boot retries stop
-  if (PcdGetBool (PcdSblResiliencyEnabled)) {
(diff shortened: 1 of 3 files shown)


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -362,8 +363,9 @@ S3ResumePath (
 
   AddMeasurePoint (0x31F0);
 
-  // Stop timer so boot retries stop
-  if (PcdGetBool (PcdSblResiliencyEnabled)) {
+  // FWU payload is the only payload in SBL scope, so stop TCO
+  // timer if another payload is set to be launched
+  if (PcdGetBool (PcdSblResiliencyEnabled) && GetBootMode () != BOOT_ON_FLASH_UPDATE) {
```
gdong1, line 368: in S3 resume path, there is no payload would be run including Osloader and UEFI payload.
in this case, we should StopTcoTimer too I think.

lsubash (approved):

gdong1 (approved):

PR #1683, opened 2022-09-13
GPIO shell command enhancement for ADL platform

Karuppa-samy:
Updated the GPIO shell command to take GPIO group and pin number as inputs.

Signed-off-by: M Karuppasamy <karuppasamy.m@intel.com>
Signed-off-by: Akshatha Thekkade <akshatha.thekkade@intel.com>

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/ShellExtensionLib/CmdGpio.c
+++ b/Platform/AlderlakeBoardPkg/Library/ShellExtensionLib/CmdGpio.c
@@ -8,6 +8,71 @@
 
 #include <Library/ShellLib.h>
 #include <Library/GpioLib.h>
+#include <GpioPinsVer2Lp.h>
+#include <GpioPinsVer4S.h>
+#include <Library/PchInfoLib.h>
+#include <Library/GpioSiLib.h>
+
+
+/*GPIO Group number for ADLN and ADLP in Alphabetic order from A to Z*/
+GPIO_GROUP GpioGrpAdlnpTable[] = {
+  GPIO_VER2_LP_GROUP_GPP_A,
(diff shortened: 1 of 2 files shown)


Platform/AlderlakeBoardPkg/Library/ShellExtensionLib/CmdGpio.c
```diff
@@ -35,6 +100,98 @@ CONST SHELL_COMMAND mShellCommandGpio = {
...
+  OUT UINT32 *GpioGroup)
+{
+  if (IsPchS()) {
+    *GpioGroup = GpioGrpAdlsTable[GrpArrayIndex];
+  }
+  else if (IsPchP() || IsPchN()) {
+    *GpioGroup = GpioGrpAdlnpTable[GrpArrayIndex];
+  }
+
+  if (*GpioGroup == 0) {
```
bejeanmo, line 122: *GpioGroup isn't guaranteed to be initialized to zero in this or the calling function. Can you please add a different check for the failing case?

Karuppa-samy: Will assign *GpioGroup value to zero in this function. After that *GpioGroup is getting assigned to GPIO group value from hardcoded global array corresponding to PCH and GrpArrayIndex. If GrpArrayIndex mapped to value zero in the global array, then that group is not present in ADL platform.

gdong1 (approved):
let's merge this patch firstly. It is expected to have a follow up patch to get the gpio group information table from bootloader (e.g. platform HOB) so that we could make this shell command common used for other platforms.

PR #1690, opened 2022-09-20
Create Python tool for SBL component corruption

sean-m-mcginn:
Create a pre-runtime tool that corrupts SBL components so that the SBL resiliency feature can more easily be tested and demonstrated

### DIFF
--- a/BootloaderCorePkg/Tools/CorruptComponentUtility.py
+++ b/BootloaderCorePkg/Tools/CorruptComponentUtility.py
@@ -1,0 +1,51 @@
+## @ CorruptComponentUtility.py
+#  This is a python IFWI/SBL component corruption utility.
+#
+# Copyright (c) 2022, Intel Corporation. All rights reserved. <BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+from CommonUtility import *
+from IfwiUtility import *
+
+import random
+import argparse
+import sys
+
+def corrupt_component(input_image, output_image, component_path):
+
+    image_bin = bytearray(get_file_data(input_image))
+    comp_bin = bytearray()
+
+    # Extract the component from the IFWI/BIOS
+    status = IFWI_PARSER.extract_component(image_bin, comp_bin, component_path)
+    if status:
+        sys.exit(1)
+
+    # Perturb some random byte in the component, ensuring it's not
+    # padding/null and different from what's currently there
+    comp_offset = random.randrange(0, len(comp_bin))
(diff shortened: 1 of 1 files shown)


gdong1 (approved):

BootloaderCorePkg/Tools/CorruptComponentUtility.py
```diff
@@ -0,0 +1,51 @@
...
+    gen_file_from_object(output_image, image_bin)
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-i', '--input-image', dest='input_image', type=str,
+                    required=True, help='Specify input IFWI/SBL image file path')
+    parser.add_argument('-o', '--output-image', dest='output_image', type=str,
+                    required=True,  help='Specify output IFWI/SBL image file path')
+    parser.add_argument('-p', '--path', dest='component_path', type=str,
+                    required=True,  help='Specify path of component to corrupt in IFWI/SBL image')
```
lsubash, line 48: Sean, could you add details to help section? Provide some description about the path of component.

BootloaderCorePkg/Tools/CorruptComponentUtility.py
```diff
@@ -0,0 +1,51 @@
...
+    status = IFWI_PARSER.extract_component(image_bin, comp_bin, component_path)
+    if status:
+        sys.exit(1)
+
+    # Perturb some random byte in the component, ensuring it's not
+    # padding/null and different from what's currently there
+    comp_offset = random.randrange(0, len(comp_bin))
+    while comp_bin[comp_offset] == 0xff or comp_bin[comp_offset] == 0x00:
+        comp_offset = random.randrange(0, len(comp_bin))
+    comp_bin[comp_offset] = random.choice([i for i in range(0, 0xff + 1) if i != comp_bin[comp_offset]])
```
lsubash, line 31: does 0xFF bytes corruption happens from start of region?

sean-m-mcginn: We only corrupt a single byte and that byte can be anywhere in the region, provided that the existing byte is not null or zero. We corrupt that single byte with a random value that is guaranteed different from the one already there.

BootloaderCorePkg/Tools/CorruptComponentUtility.py
sean-m-mcginn: The component path is found via IfwiUtility.py. An example is IFWI/BIOS/TS0/SG1A. I have added steps to the test team BKM for finding this path. I can open another PR to add this description if you think it's worth?

PR #1697, opened 2022-09-27
update CfgData_Memory.yaml

ssbyna:
provide MemTestOnWarmBoot UPD option into Config Editor

Signed-off-by: Syahirah Sabryna <nur.syahirah.sabryna.mohmad@intel.com>


bejeanmo (changes requested):
Adding this field to the config editor has no effect unless it is also copied to the FspmUpd during boot. 

Additionally, please make sure to always sign off on commits (commit with "-s" option) and provide an informative commit title (first line should be a synopsis of functional changes, not just file name).

PR #1699, opened 2022-09-28
[LiteVariable] Added Data NULL conditional check in InternalGetVariable()

Karuppa-samy:
Argument DataLen of function InternalGetVariable() inside Reclaim() function is not initialized. This uninitialized value is assigned to another variable and compared, resulting in EFI_BUFFER_TOO_SMALL error when Data is NULL. Hence added Data NULL conditional check with DataLen to overcome EFI_BUFFER_TOO_SMALL error when Data is NULL.

Signed-off-by: M Karuppasamy <karuppasamy.m@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Library/LiteVariableLib/LiteVariableLib.c
+++ b/BootloaderCommonPkg/Library/LiteVariableLib/LiteVariableLib.c
@@ -412,7 +412,7 @@ InternalGetVariable (
   DataSizeIn = *DataSize;
   VariableDataLen = FindVarHdrPtr->DataSize - VariableNameLen;
   *DataSize = VariableDataLen;
-  if (DataSizeIn <  VariableDataLen) {
+  if ((Data != NULL) && (DataSizeIn <  VariableDataLen)) {
     return EFI_BUFFER_TOO_SMALL;
   }
 


bejeanmo (commented):
Since Reclaim() calls InternalGetVariable() with Data == NULL, this DataLen param is fine to be uninitialized. While your workaround is ok, the real issue is that InternalGetVariable shouldn't be checking DataLen when Data is NULL, since it's an optional parameter. @gdong1 can you confirm?

spkamat:
In the current implementation DataLen and Data are treated as mutually exclusive variables. Hence it shall fail at the below check inside InternalGetVariable(). 

DataSizeIn = *DataSize;
  VariableDataLen = FindVarHdrPtr->DataSize - VariableNameLen;
  *DataSize = VariableDataLen;
  if (DataSizeIn <  VariableDataLen) {
    return EFI_BUFFER_TOO_SMALL;
  }

One way (probably more appropriate) to fix it would be to make DataLen dependent on Data. But for now, the current fix solves the issue.

bejeanmo:
> One way (probably more appropriate) to fix it would be to make DataLen dependent on Data. But for now, the current fix solves the issue.

@spkamat I don't think we should include a workaround in core code when the appropriate fix here should be simple. I think a simple check for Data != NULL before checking DataSizeIn should be sufficient, shouldn't it?

spkamat:
> > One way (probably more appropriate) to fix it would be to make DataLen dependent on Data. But for now, the current fix solves the issue.
> 
> @spkamat I don't think we should include a workaround in core code when the appropriate fix here should be simple. I think a simple check for Data != NULL before checking DataSizeIn should be sufficient, shouldn't it?

Not really, as DataSize is a local variable of Reclaim() passed uninitialized to InternalGetVariable(). Even if Data is != NULL, this would still be undetermined value and hence fail the check. I don't see DataSize being derived from/based on Data. Or am I missing something very obvious here?

bejeanmo:
What I mean is that in InternalGetVariable, Data is optional and DataSize is mandatory. However, when Data is NULL, DataSize is only used as an output. DataSize doesn't make sense as an input when Data is NULL. There is no output buffer, so it doesn't make sense to validate the output buffer size.

What I'm suggesting is this change in InternalGetVariable:

if (**Data != NULL &&** DataSizeIn < VariableDataLen) {
return EFI_BUFFER_TOO_SMALL;
}

Karuppa-samy:
@bejeanmo added Data NULL conditional check as suggested

PR #1710, opened 2022-10-06
[ICXD] Upstream ICX-D LCC/HCC after PV

cshur:
Tested to boot Yocto and Windows.

Signed-off-by: cshur <cs.hur@intel.com>


bejeanmo (changes requested):
Overall looks good, but many copyright headers need dates updated.

cshur:
Thanks for checking. Regarding dates, ICX team used BCD license from 1st time and if there is any changes from original file, we updated dates.

sindhuragrandhi (approved):

bejeanmo (approved):

gdong1 (commented):
looks good to me.

PR #1712, opened 2022-10-07
Fix erroneous uCode base offset in BP1 SG1A

sean-m-mcginn:
During the build process, SG1A is built
for BP0 and a copy is made for BP1. Then,
the parameters of the BP1 SG1A are adjusted
to reflect that it is in BP1. One of the
parameters that gets adjusted during this
process is the uCode base. It was noticed
that this uCode base does not get updated
at all for platforms using FSP spec 2.2, but 
another parameter gets updated instead.

Specifically, the error is in the FSP-T UPD
struct. The offset of the uCode base in
the FSP-T UPD struct is 0x40 for platforms using
FSP spec 2.2. However, the offset that actually
gets updated for platforms using FSP spec 2.2 is
0x20. This change corrects this offset for the
relevant platforms.


lsubash (approved):
Looks good

PR #1695, opened 2022-09-27
[WIP] Update Universal payload hob for secure boot.

lsubash:
Universal paayload hobs are updated for secure boot and measured boot. Event logs Hobs are created to consume by TCG2Dxe in uefi payload.

Signed-off-by: Subash Lakkimsetti <subash.lakkimsetti@intel.com>

PR #1717, opened 2022-10-11
fix: Check for possible NULL dereference introduced in #1695.

bejeanmo:
PR #1695 possible NULL deregerence was triggering static analyzer failure. Added a NULL check before dereferencing.


kvparikh89 (approved):
Looks good to me

Issue #1714, opened 2022-10-10
Interrupt packet length is maximum 8, not exclusively 8


maldf:
While integrating a HID USB keyboard I found that it had an interrupt packet length of 4.
In
https://github.com/slimbootloader/slimbootloader/blob/107fabbebdbc8ebb60dfd038c0896a99cffd1a58/BootloaderCommonPkg/Library/XhciLib/XhcPeim.c#L920
it requires the interrupt packet length to be only 8.
If it is changed so that:
... (MaximumPacketLength > 8)
similar to all the other tests, the HID USB keyboard works.

randylintw:
Hi, 
Would you like to submit a patch to fix this issue?

maldf:
Patch attached:
[0001-USB-irq-transfer-size.txt](https://github.com/slimbootloader/slimbootloader/files/9764265/0001-USB-irq-transfer-size.txt)

randylintw:
Could you create a pull request for this fix?

randylintw:
Thanks, btw, your patch format is incorrect, could you modify ? 
 * Missing Signed-off-by! (Note: this must be added by the code contributor!)
WARNING - Line 3 of commit message is too long (77 >= 76).

randylintw:
Hi Mald, thanks for your contribute and you can close this issue, thanks.

Issue #1713, opened 2022-10-10
USB descriptor iteration error


maldf:
While iterating over a USB descriptor in GetExpectedDescriptor(), the "Len" variable always uses the 1st byte in the buffer, see
https://github.com/slimbootloader/slimbootloader/blob/107fabbebdbc8ebb60dfd038c0896a99cffd1a58/BootloaderCommonPkg/Library/UsbBusLib/UsbPeim.c#L913
This is only true for the 1st iteration, thereafter the variable should advance as the descriptor advances, it should be:
Len = Buffer[parsed]

randylintw:
Hi, 
Would you like to submit a patch to fix this issue?

maldf:
Patch attached.
[0001-USB-fix-descriptor-parsing.txt](https://github.com/slimbootloader/slimbootloader/files/9764244/0001-USB-fix-descriptor-parsing.txt)

randylintw:
Could you create a pull request for this fix?

PR #1718, opened 2022-10-13
Usb fixes

maldf:
Pull request for issues #1713 and #1714


gdong1 (approved):
Thanks for these fixes. It looks good to me.

PR #1720, opened 2022-10-13
Enhance multiboot support

gdong1:
Currently SBL would return when preferred graphics mode is set in the multiboot image. This patch would continue boot and print the preferred graphics mode information.


bejeanmo (approved):
Guo, this looks fine. The multiboot spec doesn't seem to be very clear on what this flag actually means wrt headless systems.

PR #1727, opened 2022-10-19
[ADL] enable TCO timer by default

gdong1:
TCO timer could be enabled regardless resiliency feature. So just remove the resiliency conditional.


sindhuragrandhi:
This might introduce new issues as we haven't done this so far. It needs to be tested fully before it can be merged. We are in the middle of doing multiple releases on PS and N. Until we are completely done with the releases, we shouldnt merge this in.

PR #1728, opened 2022-10-19
[EHL] support multi VBT

Daran30:
Currently,  Slim Bootloader support for multiple VBT files doesn't work on EHL. Support for multiple VBT files, which works on other platform, does not work correctly for EHL. However, this change can fix it locally by adding the support from another platform into the EHL support once integrated into the public SBL

Signed-off-by: Devathurai Linggeis Daran linggeis.daran.devathurai@intel.com


bejeanmo (commented):
Darren, did you mean to commit the BoardConifg.py and Stage2BoardInitLib.c files instead of the .patch file?

Daran30:
> Darren, did you mean to commit the BoardConifg.py and Stage2BoardInitLib.c files instead of the .patch file?

Hi Bejean, Yes. I meant to commit the BoardConfig.py and Stage2BoardInitLib.c files

PR #1733, opened 2022-10-20
[EHL] support multi VBT

Daran30:
Currently, Slim Bootloader support for multiple VBT files doesn't work on EHL. Support for multiple VBT files, which works on other platform, does not work correctly for EHL. However, this change can fix it locally by adding the support from another platform into the EHL support once integrated into the public SBL

Signed-off-by: ldevathu <linggeis.daran.devathurai@intel.com>


gdong1 (approved):
looks good to me.

PR #1734, opened 2022-10-26
[RPL/ADL] L2 CAT not getting enabled when TCC enabled.

bejeanmo:
FspsUpdUpdateLIb was not enabling FspsUpd->FspsConfig.L2QosEnumerationEn based on PLATFORM_ADLS, which is no longer used. L2QosEnumerationEn is present in ADL-P,S,PS, and RPL-S,P FSP UPD, just not ADL-N, so need to prevent compiling this code for that platform only.

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
+++ b/Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
@@ -468,7 +468,7 @@ TccModePostMemConfig (
 
     if (IsPchS ()) {
       FspsUpd->FspsConfig.TccMode = 1;
-#ifdef PLATFORM_ADLS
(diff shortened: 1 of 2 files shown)


bejeanmo:
As far as I can tell PLATFORM_ADLS is no longer used or defined anywhere. I did a basic build test of all ADL/RPL platforms with this change and ENABLE_TCC=1 , and it doesn't break on this piece of code. I also confirmed L2 CAT now gets enabled correctly on RPL-S with this change. @sindhuragrandhi this change will affect ADL platforms also. Do you see any issue with it?

sindhuragrandhi (approved):

sindhuragrandhi:
I donot see any issues with this change. I believe PLATFORM_ADLS is generated during build time.

bejeanmo:
Thanks @sindhuragrandhi I see the #define now in the build scripts.

randylintw:
Can re-title since it's fixing in adl platform ?

gdong1 (approved):

PR #1740, opened 2022-11-02
[ADLN] Set FIVR transition value

sindhuragrandhi:
For S0ix usecase, set Fivr transition time to zero to resolve a hardware issue.

Signed-off-by: "Grandhi, Sindhura" <sindhura.grandhi@intel.com>


Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
```diff
@@ -867,6 +867,10 @@ UpdateFspConfig (
       FspsConfig->PchLanEnable = 0x0;
       ZeroMem (FspsConfig->PchIshI2cEnable, sizeof (FspsConfig->PchIshI2cEnable));
       ZeroMem (FspsConfig->PchIshGpEnable, sizeof (FspsConfig->PchIshGpEnable));
+      // Apply FIVR workaround
```
kvparikh89, line 870: This is not required on closer look. This is already taken care of on line 1134

kvparikh89 (changes requested):

Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
sindhuragrandhi: Agree

PR #1744, opened 2022-11-07
OsLoader: Add Multiboot-2 support

bachaue1:
- Add (partial) multiboot-2 support as specified in https://www.gnu.org/software/grub/manual/multiboot2/multiboot.html
- header file imported (and adjusted to SBL OsLoader) from https://github.com/NetBSD/src/blob/trunk/sys/arch/i386/include/multiboot2.h

Signed-off-by: Bruno Achauer <bruno.achauer@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Include/Library/MultibootLib.h
+++ b/BootloaderCommonPkg/Include/Library/MultibootLib.h
@@ -234,11 +234,18 @@ typedef struct {
   BOOTPARAMS_IMAGE_DATA   BootParamsData;
 } TRUSTY_IMAGE_DATA;
 
+typedef struct multiboot2_start_tag MULTIBOOT2_START_TAG;  /* opaque type */
+
+typedef struct {
+  MULTIBOOT2_START_TAG   *StartTag;
+} MULTIBOOT2_INFO;
+
 #define MAX_MULTIBOOT_MODULE_NUMBER  16
 typedef struct {
   IMAGE_DATA              BootFile;
   IMAGE_DATA              CmdFile;
   MULTIBOOT_INFO          MbInfo;
+  MULTIBOOT2_INFO         Mb2Info;
   IA32_BOOT_STATE         BootState;
   UINT16                  CmdBufferSize;
   UINT16                  MbModuleNumber;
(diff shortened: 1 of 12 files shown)


BootloaderCommonPkg/Include/Library/MultibootLib.h
```diff
@@ -234,11 +234,16 @@ typedef struct {
   BOOTPARAMS_IMAGE_DATA   BootParamsData;
 } TRUSTY_IMAGE_DATA;
 
+typedef struct {
+  struct multiboot2_start_tag         *start_tag;
+} MULTIBOOT2_INFO;
```
gdong1, line 239: This library header file is an interface for other modules to use Multiboot library.
so this library should not depend on library implementation including structure definitions.
For other modules (e.g. OsLoader), it looks "struct multiboot2_start_tag" is not defined in API file.
and prefer updating the variable name "start_tag" in this file to follow SBL variable style (e.g. StartTag).

PayloadPkg/OsLoader/OsLoader.h
```diff
@@ -170,6 +171,18 @@ DumpMbInfo (
...
+Print out the Multiboot-2 information block.
+
+@param[in]  Mi  The Multiboot-2 information block to be printed.
+
+**/
+struct multiboot2_start_tag;    /* XXX: move to header file */
+VOID
+DumpMb2Info (
+  IN  CONST struct multiboot2_start_tag *Mi
+  );
```
gdong1, line 184: Add DumpMb2Info  to MultibootLib.h if it is expected to be called by other modules.

BootloaderCommonPkg/Include/Library/MultibootLib.h
bachaue1: "struct multiboot2_start_tag" is opaque from a client point of view (for using the library it's only necessary that it's a pointer to a structure but not the structure definition).  I'll change the header file to stress that.

PayloadPkg/OsLoader/OsLoader.h
bachaue1: OsLoader.h already has DumpMbInfo(); should I move that as well?

BootloaderCommonPkg/Include/Library/MultibootLib.h
gdong1: Thanks for the update.

PayloadPkg/OsLoader/OsLoader.h
gdong1: Got it. It is ok to put it together with DumpMbInfo(). Late we could have a separate patch to move them.

gdong1 (approved):

PR #1746, opened 2022-11-07
Stop TCO timer if OS/UEFI payload not found

sean-m-mcginn:
If OS/UEFI payload is not found, stop the TCO timer before halting the CPU as there is no backup of
OS/UEFI payload


lsubash (approved):
looks good.
We can think opf switch to Osloader when UEFI payload is not present.

kvparikh89 (approved):
Looks good to me

gdong1 (approved):

PR #1752, opened 2022-11-14
feat: [RPL-P] Add support for enabling FSP FuSa features.

bejeanmo:
FSP UPD FuSa toggles will be set based on new config data fields where applicable or to predefined values when called for by the FuSa spec. This requires setting PcdFusaSupport at build time in case platform FSP doesn't support FuSa.

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
+++ b/Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
@@ -689,4 +689,12 @@ UpdateFspConfig (
       }
     }
   }
+#if FixedPcdGetBool(PcdFusaSupport)
+  if ((SiCfgData != NULL) && (SiCfgData->FusaConfigEnable)) {
+    DEBUG((DEBUG_INFO, "FuSa enabled. Overriding CpuCrashLog and IBECC settings to Enabled - All requests protected.\n"));
+    Fspmcfg->CpuCrashLogEnable = 1;
+    Fspmcfg->Ibecc = 1;
+    Fspmcfg->IbeccOperationMode = 2;
+  }
+#endif
 }
--- a/Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
+++ b/Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
@@ -1238,6 +1238,28 @@ UpdateFspConfig (
   if (FeaturePcdGet (PcdEnablePciePm)) {
     StoreRpConfig (FspsConfig);
   }
+
+#if FixedPcdGetBool(PcdFusaSupport)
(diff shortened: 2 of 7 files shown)


Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
```diff
@@ -689,4 +689,13 @@ UpdateFspConfig (
       }
     }
   }
+#if FixedPcdGetBool(PcdFusaSupport) == 1
```
gdong1, line 692: "== 1" could be removed since PcdFusaSupport is boolean

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
```diff
@@ -689,4 +689,13 @@ UpdateFspConfig (
...
+  if (SiCfgData->FusaConfigEnable)
```
gdong1, line 693: To avoid potential KW issue, I think it need check if SiCfgData is NULL as below.
if ((SiCfgData != NULL) && (SiCfgData->FusaConfigEnable)) {

Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
```diff
@@ -1238,6 +1238,26 @@ UpdateFspConfig (
   if (FeaturePcdGet (PcdEnablePciePm)) {
     StoreRpConfig (FspsConfig);
   }
+
+#if FixedPcdGetBool(PcdFusaSupport) == 1
+    FspsConfig->FusaConfigEnable = SiCfgData->FusaConfigEnable;
```
gdong1, line 1243: "== 1" could be removed.
To avoid potential KW issue, I think it need check if SiCfgData is NULL as below.
if (SiCfgData != NULL) {

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
bejeanmo: Fixed

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
bejeanmo: Should be fixed now.

Platform/AlderlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
bejeanmo: Should be fixed now.

gdong1 (approved):

PR #1754, opened 2022-11-15
Remove unnecessary check in partition library

gdong1:
In the Ubuntu ISO image, it has 0xF8 partition entries. The partition library could not work as expected since it expects the partition entries are no more than 128.

Potentially it could have more partition entries, so just remove the check.


bejeanmo (approved):
Looks good Guo. I see this 128 Partition limit mentioned in some examples on wikis, but as far as I can tell it isn't mentioned in the spec at all. I wonder where it came from.

PR #1755, opened 2022-11-15
Move TCO Timer Status Check/Start Right Before FSP-M

sean-m-mcginn:
It was discovered that, when BiosRedAssistance = Enabled, CSME actually uses the ACM active timer to time all of SG1A through the beginning of SG1B (up to end of FSP-M), flipping the TS and rebooting if a certain timeout is reached. This change closes the resiliency gap in SG1B while ensuring that there is as little overlap as possible between the TCO timer and the ACM active timer.


gdong1 (approved):
It looks good to me.

PR #1756, opened 2022-11-16
Allow to load capsules signed with RSA3072 from raw partition

arturkow2000:
When loading capsule from raw partition, some check prevent capsules signed by a RSA3072 from being recognized as a valid capsule. Remove obsolete checks which expect RSA2048, signature type is already checked at a later stage from AuthenticateCapsule.


bejeanmo (approved):

bejeanmo (changes requested):
Please add sign-off message to commit.

lsubash (approved):

bejeanmo (approved):

gdong1 (approved):

PR #1770, opened 2022-11-30
Allocate empty uCode slots for future uCode updates

sean-m-mcginn:
The FIT allows for some empty uCode slots to be defined in anticipation of future uCode updates. This change adds as many empty slots as possible after the non-empty slots in the uCode region.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -356,39 +356,34 @@ class Build(object):
 
             # Collect all CPU uCode images
             u_code_images = []
-            while ucode_offset < len(rom):
-                # Extract info from CPU uCode images
-                ucode_hdr = UCODE_HEADER.from_buffer(rom, ucode_offset)
-                if ucode_hdr.header_version == 1:
-                    u_code_images.append(ucode_offset)
-                    num_fit_entries += 1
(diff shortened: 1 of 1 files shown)


gdong1 (approved):
The change looks good to me. But the big uCode region size especially having empty uCode slot would impact boot performance in order to load all the uCode including empty region. Maybe late we could add some comments somewhere (e.g. to the UCODE_SIZE in board config file) on the impact.

BuildLoader.py
```diff
@@ -356,39 +356,34 @@ def update_fit_table (self):
...
-                # uCode Update
-                if len(u_code_images) > 0:
-                    offset = u_code_images.pop(0)
-                    fit_entry.set_values(base + offset, 0, 0x100, 0x1, 0)
-                    print ('  Patching entry %d with 0x%08X - uCode' % (i, fit_entry.address))
-                else:
-                    print ('  Nullifying unused uCode patch entry %d' % i)
-                    fit_entry.type      = 0x7f
-
-            if len(u_code_images) > 0:
```
lsubash, line 388: why this condition is removed?

lsubash (commented):
Please do a check with legacy platforms where ucode SLOT SIZE is not set.

PR #1777, opened 2022-12-06
Update universal payload ID

gdong1:
Update the universal payload ID to align with
the universal payload specification.


kvparikh89 (approved):

bejeanmo:
@gdong1 will there be a corresponding PR in the uefi payload repo match this?

gdong1:
> here be a corresponding PR in the uefi payload repo match thi

The open source EDK2 code already made this change. the internal UEFI payload would rebase edk2 master.

kvparikh89:
@gdong1: Does this affect the branch that we use 202008?

gdong1:
> s this affect the branch that we use 20200
no impact.

gdong1:
the qume test failure was caused by the test case which used a pre-built UEFI payload and it is outdated.
The test case should be updated separately.

PR #1784, opened 2022-12-14
fix: [RPL-S] Fix for TSN build error caused by FSP UPD updates.

bejeanmo:
RPL-S FSP UPD was updated so PchTsnEnable is now a 2-byte array with a separate byte for each port. This is now different from the ADL platforms and needs to be included selectively at build time.


bejeanmo:
Using the build time PLATFORM_RPLS check here isn't ideal here, but I think the only way around this is to duplicate a lot of code for raptorlake.

sindhuragrandhi (approved):

gdong1 (approved):

PR #1796, opened 2023-01-03
[ADL] Skip SioInit when S0ix is enabled.

kvparikh89:
On ECless platforms, Sio is initialized during boot flow. However, when S0ix is enabled, Sio need not be initialized.

Signed-off-by: Kalp Parikh <kalp.parikh@intel.com>


cshur (approved):
Looks good

gdong1 (approved):

PR #1805, opened 2023-01-11
feat/tool: allow to update container component together with BIOS

stanleyintel:
container component such as '-p TMAC:IPFW" should be allowed to update together with BIOS region (-p BIOS)


lsubash (commented):
Stanley, we would like separate out redundant components and non-redundant components.
This change has by @sean-m-mcginn done as part of firmware resiliency. 
What is the reason we would want to combine bios and container components?

stanleyintel:
> Stanley, we would like separate out redundant components and non-redundant components. This change has by @sean-m-mcginn done as part of firmware resiliency. What is the reason we would want to combine bios and container components?

Hi,

Example, TMAC:IPFW does not belong to BIOS region. It should be allowed to update together with BIOS region. This kind of update is common in all platform supporting TCC (e.g., ADL, EHL, TGL)

stanleyintel:
> > Stanley, we would like separate out redundant components and non-redundant components. This change has by @sean-m-mcginn done as part of firmware resiliency. What is the reason we would want to combine bios and container components?
> 
> Hi,
> 
> Example, TMAC:IPFW does not belong to BIOS region. It should be allowed to update together with BIOS region. This kind of update is common in all platform supporting TCC (e.g., ADL, EHL, TGL)

Oops, I am wrong..  IPFW is still in BIOS region

PR #1818, opened 2023-01-31
fix: enumerate all USB instances to search devices

arghyabi:
instance 0 was hardcoded, so not able to read the devices from other instances. 
Now read the instances number from the device table and enumerate all USB instances one by one.

Signed-off-by: Biswas Arghya <arghya.biswas@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Library/ShellLib/CmdUsbDev.c
+++ b/BootloaderCommonPkg/Library/ShellLib/CmdUsbDev.c
@@ -1,7 +1,7 @@
 /** @file
   Shell command `usbdev` to view all usb device list
 
-  Copyright (c) 2022, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2022 - 2023, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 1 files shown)


spkamat (approved):

tsaikevin:
@arghyabi, if my understanding is correct, this function looks for USB device from table mPlatformDevices which is defined in Stage1BBoardInitLib.c.

It has one instance for USB boot device now. do you make any change to it?

BootloaderCommonPkg/Library/ShellLib/CmdUsbDev.c
```diff
@@ -66,55 +66,66 @@ ShellCommandUsbDevFunc (
...
-      //
-      // Checking the NameStr is not combination of spaces and tabs
-      //
-      if (((char)NameStr[StringIndex]) != 32 && ((char)NameStr[StringIndex]) != 9) {
-        CharFound++;
+      StringSize = StrSize (NameStr);
+
+      CharFound = 0;
+      for (StringIndex = 0; StringIndex < StringSize; StringIndex++) {
+        if (NameStr[StringIndex] == 0) {
```
gdong1, line 108: It looks this check is not necessary since it could not be 0 before String Size.

arghyabi: This check is not to validate the length of the string, it is there to validate those devices' names returned as null strings. If the string is null then it will be marked as an 'Unnamed Device'.

tsaikevin (approved):

BootloaderCommonPkg/Library/ShellLib/CmdUsbDev.c
gdong1: If the name string is null, it will be set to "N/A" in line 102 already. so it is not possible to have NULL string here.

PR #1822, opened 2023-02-01
[ADL/RPL] Enhance PSD-related code

sean-m-mcginn:
This change fixes a bound range exception thrown in HECI call for FW capabilities and prioritizes use of HOBs over HECI calls in PSD-related code.


lsubash (approved):

lsubash:
Would you update the platforms tested for this PR.

PR #1828, opened 2023-02-04
ICXD: Fix R_CDF_PCH_GPIO_PCR_GPP_N_PADCFGLOCKTX register offset

buytenh:


cshur:
There are two PADCFGLOCKTX registers. (TxState Register)
PADCFGLOCKTX_WEST2_0 is C4h, PADCFGLOCKTX_WEST1_0 is D4h. GPIO table refer C4h register.
No need to update this code.

buytenh:
This about the EAST side GPIO registers, not the WEST side.

There is indeed a PADCFGLOCKTX register at offset 0xc4, but that is
the PADCFGLOCKTX register for the EAST2 (L bank), while the PADCFGLOCKTX
register for the EAST0 (N bank) is at 0xd4.

cshur (approved):
Looks good to me.

cshur:
I checked with BIOS and yes. You are right.

PR #1833, opened 2023-02-17
Ensure that ELF images are loaded at their preferred address

bachaue1:
Currently, loading an ELF image will relocate the image (to its preferred load address) only if there is at least one ELF segment that is not completely contained in the file loaded (i.e. the memory segment size is larger than the size inside the file. This behavior is generally triggered by the segment that contains the .bss section).

Consequently, images that contain initialized data only (e.g. QNX IFS files) don't get relocated, and will usually crash when jumping to the image entry point.

This change checks whether the ELF image is already loaded at the preferred address, and forces image relocation if not.

### DIFF
--- a/BootloaderCommonPkg/Library/ElfLib/ElfLib.c
+++ b/BootloaderCommonPkg/Library/ElfLib/ElfLib.c
@@ -1,7 +1,7 @@
 /** @file
   ELF library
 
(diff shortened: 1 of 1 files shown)


rprangar (approved):
LGTM

PR #1835, opened 2023-02-27
Fix TPM2 Event Logging

sean-m-mcginn:
These changes ensure that:

- Each TPM2 event type matches the structure of the data associated with it
- Aligns TPM2 events between BIOS and SBL


gdong1:
Please fix the build failures.

lsubash (commented):
Check the case of CSME measured boot.
TpmHashAndExtendPcrEventLog in MeMeasurementlib.c

PR #1839, opened 2023-03-07
[ICXD][LCC] Update Microcode for both LCC and HCC

cshur:
Version update 1.1 to 1.2.
Update Microcode 201 to 230


paaakash (approved):
Looks good to me.

gdong1 (approved):

PR #1841, opened 2023-03-07
fix: [RPL-P] Build error when ENABLE_SOURCE_DEBUG=1.

bejeanmo:
On platforms where only serial console input is enabled, enabling source debug will lead to an infinite loop in ConsoleRead(). Newer compilers detect this and generate a build error.


sindhuragrandhi (approved):

sindhuragrandhi:
@bejeanmo since the change is in the common file, can we remove the [RPL-P] from the title so we all know its a common fix when referring to git history?

gdong1 (approved):

PR #1843, opened 2023-03-10
Fix OsLoader handling of non-container images and remove PcdContainerBootEnabled

atharvalele:
OsLoader should not attempt to boot images that are not packaged as containers.

Earlier, OsLoader would attempt to load and boot a non-container image and fail in
SetupBootImages() as the LoadedImage struct would not be populated correctly.

Now, ParseBootImage() will report a non-container file as unsupported and OsLoader
will move on to the next entry in the boot options.


gdong1 (approved):
It looks good to me

rprangar (approved):
LGTM

lsubash (approved):
Looks good.

PR #1844, opened 2023-03-14
fix: Build stage2 stack and heap from suitable available memory.

bejeanmo:
Previously, Stage1B was building the Stage 2 stack and heap immediately below the FSP Reserved Memory resource descriptor region, assuming that this was a large available memory region. This assumption breaks when FSP resource descriptors move around, leading to unpredictable behavior.

This modification forces stage 1b to select a region for stack/heap from the FSP available memory descriptors that has adequete space for stack, heap, ACPI resources, and Payload reserved mem.

### DIFF
--- a/BootloaderCorePkg/Library/FspSupportLib/FspSupportLib.c
+++ b/BootloaderCorePkg/Library/FspSupportLib/FspSupportLib.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2023, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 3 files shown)


sindhuragrandhi (approved):

BootloaderCorePkg/Library/FspSupportLib/FspSupportLib.c
```diff
@@ -33,6 +33,54 @@ GetFspNvsDataBuffer (
...
+  requested length.
+
+  @param  HobListPtr   A HOB list pointer.
+  @param  Length       A pointer to the requested length. If suitable
+                       descriptor is found, length will be updated with
+                       resource length.
+  @retval              Reserved region start address.  0 if this region does not exist.
+ */
+UINT64
+GetFspAvailableSystemMem(
```
rprangar, line 47: This function tries to find available memory of given length?

Then should it be renamed - as FindLoaderMemory or something like that without the FSP word in it?

BootloaderCorePkg/Library/FspSupportLib/FspSupportLib.c
```diff
@@ -33,6 +33,54 @@ GetFspNvsDataBuffer (
...
+
+  /*
+   * Get the HOB list for processing
+   */
+  Hob.Raw = (VOID *)HobListPtr;
+
+  /*
+   * Collect memory ranges
+   */
+  while (!END_OF_HOB_LIST (Hob)) {
```
rprangar, line 62: Do we need to add any additional checks? Can this function find memory which is less than 1 MB?

BootloaderCorePkg/Library/FspSupportLib/FspSupportLib.c
bejeanmo: It's looking in the FSP-M HOB list resource descriptors for available memory.

BootloaderCorePkg/Library/FspSupportLib/FspSupportLib.c
bejeanmo: Guo and I are still discussing this fix. I think the three of us probably need to discuss this change some more.

bejeanmo:
Updated implementation to use highest address region found for completeness, but this will be closed in favor of a temporary platform solution while we address this in FSP.

PR #1847, opened 2023-03-17
[ADL/RPL] Update PlatformMemorySize default value

gdong1:
SBL config data PlatformMemorySize is used to set FSP UPD PlatformMemorySize which is used by FSP to check available memory range size for FSP reserved memory.
A too small size might cause FSP to use a small available memory range used for other purpose. So change the default value of PlatformMemorySize to 64MB.
Since default value change, remove unnecessary override.

### DIFF
--- a/Platform/AlderlakeBoardPkg/CfgData/CfgDataInt_Adls_Crb_Ddr4.dlt
+++ b/Platform/AlderlakeBoardPkg/CfgData/CfgDataInt_Adls_Crb_Ddr4.dlt
@@ -2,7 +2,7 @@
 #
 #  Platform Configuration Delta File.
 #
-#  Copyright (c) 2020 - 2021, Intel Corporation. All rights reserved.<BR>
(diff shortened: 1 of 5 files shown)


bejeanmo (approved):

Platform/AlderlakeBoardPkg/CfgData/CfgDataInt_Adls_Crb_Ddr4.dlt
kvparikh89, line 1: Please update year.

kvparikh89 (commented):
Please update the year in all files.

Platform/AlderlakeBoardPkg/CfgData/CfgDataInt_Adls_Crb_Ddr4.dlt
gdong1: done

PR #1855, opened 2023-03-24
fix: verify fwupdate capsule only when Verified Boot is enabled

vincentchchen:
Firmware update got failed when verified boot was turned off. While fwupdate capsule is always generated with signing key,
it should be verified only when verified boot is enabled. Otherwise, raise a warning to inform that the capsule is not authenticated.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -1,7 +1,7 @@
 /** @file
 This driver is to update firmware in boot media.
 
-Copyright (c) 2017 - 2021, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2017 - 2023, Intel Corporation. All rights reserved.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -747,7 +747,7 @@ AuthenticateCapsule (
   }
 
   if (Header->ImageOffset >= FwSize || Header->ImageOffset + Header->ImageSize >= FwSize) {
-    DEBUG ((DEBUG_ERROR, "Invalid capsule: ImageOffset=0x%x, ImageOffset=0x%x\n", Header->ImageOffset, Header->ImageOffset));
+    DEBUG ((DEBUG_ERROR, "Invalid capsule: ImageOffset=0x%x, ImageSize=0x%x\n", Header->ImageOffset, Header->ImageSize));
     return EFI_INVALID_PARAMETER;
   }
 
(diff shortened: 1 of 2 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -747,16 +747,21 @@ AuthenticateCapsule (
...
     return EFI_INVALID_PARAMETER;
   }
 
-  PubKeyHdr       = (PUB_KEY_HDR *) (FwImage + Header->PubKeyOffset);
-  SignatureHdr    = (SIGNATURE_HDR *) (FwImage + Header->SignatureOffset);
-  Status = DoRsaVerify (FwImage, Header->SignatureOffset, HASH_USAGE_PUBKEY_FWU, SignatureHdr, PubKeyHdr, PcdGet8(PcdCompSignHashAlg), NULL, NULL);
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "Image verification failed, %r!\n", Status));
-    return EFI_SECURITY_VIOLATION;
+  if (!FeaturePcdGet (PcdVerifiedBootEnabled)) {
```
lsubash, line 754: When verified boot is turned off in SBL, Fwupdate doesn't need verification. 
AuthenticateCapsule should be kept under feature flag PcdVerifiedBootEnabled.

vincentchchen: Yes, I was thinking about this, since the AuthenticateCapsule() includes some file size checks. So added the evaluation on RsaVerify only. But it's good to keep the evaluation on AuthenticateCapsule(), I will move the code over.

lsubash (approved):

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -747,7 +747,7 @@ AuthenticateCapsule (
   }
 
   if (Header->ImageOffset >= FwSize || Header->ImageOffset + Header->ImageSize >= FwSize) {
-    DEBUG ((DEBUG_ERROR, "Invalid capsule: ImageOffset=0x%x, ImageOffset=0x%x\n", Header->ImageOffset, Header->ImageOffset));
+    DEBUG ((DEBUG_ERROR, "Invalid capsule: ImageOffset=0x%x, ImageSize=0x%x\n", Header->ImageOffset, Header->ImageSize));
```
kvparikh89, line 750: Can you please also edit the function description regarding this change?

gdong1 (approved):

PR #1858, opened 2023-03-31
fix: [ADL/RPL] SBL image larger than 16MB hangs after TempRamInit.

bejeanmo:
The PCH decodes MMIO accesses to the top of 4GB to SPI flash for a maximum window size of 16MB. For SBL images larger than 16MB, this PostTempRamInit hook was causing the MTRR to overlap with NEM stack set up by FSP-T, causing a hang.


sindhuragrandhi (approved):

sindhuragrandhi:
@bejeanmo  Please update the year

gdong1 (approved):

PR #1860, opened 2023-04-03
Align CSME update with BIOS/driver

sean-m-mcginn:
This change is to align the CSME update mechanism with BIOS and the CSME update driver.
In SBL, we check if we are updating to the same version of CSME, but stop short of cancelling the update if we are. This check throws an error with some platforms as CSME isn't always on FPT_PARTITION_NAME_FTPR partition. Since there is no equivalent check in BIOS, we just delete this check to resolve the error.
Also the latest version of the CSME update driver contains some error codes that are not in SBL. Therefore, we add these error codes to help with future debugs.


PayloadPkg/FirmwareUpdate/CsmeFwUpdate.c
kvparikh89, line 1: Update year

lsubash (commented):
Sean, does these updates applicable for older platforms as ADL, TGL, EHL?
Its common common module from CFL. Does these headers matches the SBL lib headers in CSME packages?

lsubash:
Lets confirm with CSME folks also for backward compatibility.
Ideally. when CSMD driver gets built, its recommended to update the all the dependencies along with lib.

lsubash (approved):
Looks good.
Validated by vincent on EHL.
Sean validated on MTL, ADL.

gdong1 (approved):

PR #1864, opened 2023-04-05
feat: [ADL/RPL] Update CrashLog support implementation for latest FSP.

bejeanmo:
New CrashLog support design is for FSP to collect CrashLog info and report in HOB. Bootloader responsibility is just to populate ACPI BERT with FSP collected CrashLog Data now. That way bootloader has no Silicon code.


sindhuragrandhi:
Where will the CpuCrashLogDataBufferHob be invoked currently?

gdong1 (approved):

PR #1879, opened 2023-04-21
Add board name option to clean command

tsaikevin:
Build script will not copy FSP, VBT and Microcode bin files from repository if it finds these files existing in taget folder.
Above step keeps them from unintentional update in series of build process. Adding a board name to clean command helps to get latest binaries from repository in next build.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -1439,8 +1439,9 @@ def main():
     for cfgfile in board_cfgs:
         module_name = os.path.basename(os.path.dirname(cfgfile))[:-8] + os.path.basename(cfgfile)[:-3]
         brdcfg = load_source(module_name, cfgfile)
-        board_names.append(brdcfg.Board().BOARD_NAME)
-        module_names.append(brdcfg)
+        if brdcfg.Board().BOARD_NAME:
+            board_names.append(brdcfg.Board().BOARD_NAME)
+            module_names.append(brdcfg)
 
     ap = argparse.ArgumentParser()
     sp = ap.add_subparsers(help='command')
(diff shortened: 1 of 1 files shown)


randylintw:
Will the fsp/ucode also clean in the SblPlatform.zip?

BuildLoader.py
```diff
@@ -1497,6 +1497,56 @@ def cmd_clean(args):
...
+            for index, name in enumerate(board_names):
+                if args.board == name:
+
+                    board  = module_names[index].Board()
+                    if hasattr(board, 'FSP_INF_FILE'):
+                        fsp_inf = board.FSP_INF_FILE
+                    else:
+                        fsp_inf = 'Silicon/%s/FspBin/FspBin.inf' % board.SILICON_PKG_NAME
+
+                    if os.path.exists(fsp_inf):
```
gdong1, line 1532: If the workspace is not under SblOpen, fsp_inf might not be found since it is a relative path to current workspace.
you might check os.path.join (sbl_dir, fsp_inf) and os.path.join (os.environ['PLT_SOURCE'], fsp_inf) if 'PLT_SOURCE' in os.environ. Same for other INF file

tsaikevin: @gdong1, update, pls review. thanks

tsaikevin:
> Will the fsp/ucode also clean in the SblPlatform.zip?

this option doesn't change anything beyond files list added to fspbin.inf and microcode.inf.
it should be fine. thanks

tsaikevin:
> Using "$git clean -fdx folder" could be another simpler implementation, while it will clean anything even beyond file list in inf. But should be able to fulfill this feature if the folder is carefully specified. e.g. $ git clean -fdx Platform\AlderlakeBoardPkg\VbtBin

@vincentchchen thanks for input, I try to avoid git clean cmd as you mention all untrack files will be removed. we have no clue what files are added to fsp, vbt, microcde folders. In development/debug period, it can be a original copy and some test version binaries. With this patch, it reverts the file copying from previous build and keep impact minimal.

gdong1 (approved):

PR #1891, opened 2023-05-01
Update litvariable

gdong1:


BootloaderCommonPkg/Library/LiteVariableLib/LiteVariableLib.inf
```diff
@@ -38,7 +38,7 @@
   HobLib
 
 [Guids]
-
+  gZeroGuid
```
sindhuragrandhi, line 41: Is it mandatory to specify ZeroGuid if NULL?

gdong1: That's the implementation choice to simplify the caller.

BootloaderCommonPkg/Library/LiteVariableLib/LiteVariableLib.c
```diff
@@ -363,17 +365,23 @@ InternalGetVariable (
...
   UINTN                   DataSizeIn;
+  EFI_GUID                *VarGuid;
 
   if ((DataSize == NULL) || (VariableName == NULL)) {
     return EFI_INVALID_PARAMETER;
   }
 
+  VarGuid = VariableGuid;
+  if (VarGuid == NULL) {
+    VarGuid = &gZeroGuid;
```
tsaikevin, line 376: @gdong1 is it suggested to declare GUID somewhere in header as below? although actually I don't find any warning in build log, only find complain in VS code. thanks

extern EFI_GUID gZeroGuid

bejeanmo: EDK2 build should put this extern in the autogen.h since it is listed in the [Guids] section of the .inf file.

gdong1: That's true. the build tool will generate it. I think it is not necessary to have a header file if there is no GUID structure for this guid.

bejeanmo (approved):
Implementation looks good. I implemented some test variable writes of large data chunks and confirmed that repeated SetVariable calls with changing data correctly trigger Reclaim. Reclaim to secondary region works as expected. GetVariable and GetNextVariableName continue to return correct data after reclaim and across reboots.

gdong1:
> riable calls with changing data correctly trigger Reclaim. Reclaim to secondary regi

Thanks for the validation.

PR #1892, opened 2023-05-02
fix: [RPL-P] CRB hang in UEFI payload when PCIe switch present.

bejeanmo:
This change only sets the GFX framebuffer cache type to Write-combining when not booting the UEFI payload. The UEFI payload PCI hostbridge driver will override the WC setting anways for the whole PCI root bridge memory range, so it wasn't gaining the write-combining performance bonus, and was causing a CPU exception in this particular case for RPL-P.

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -458,9 +458,13 @@ BoardInit (
     if (FeaturePcdGet (PcdEnablePciePm)) {
       PciePmConfig ();
     }
-    Status = SetFrameBufferWriteCombining (0, MAX_UINT32);
-    if (EFI_ERROR(Status)) {
-      DEBUG ((DEBUG_INFO, "Failed to set GFX framebuffer as WC\n"));
+    // UEFI Payload will change cache type to UC based on PCI root bridge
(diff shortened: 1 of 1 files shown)


Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -458,9 +458,13 @@ BoardInit (
     if (FeaturePcdGet (PcdEnablePciePm)) {
       PciePmConfig ();
     }
-    Status = SetFrameBufferWriteCombining (0, MAX_UINT32);
-    if (EFI_ERROR(Status)) {
-      DEBUG ((DEBUG_INFO, "Failed to set GFX framebuffer as WC\n"));
+    // UEFI Payload will change cache type to UC based on PCI root bridge
```
sindhuragrandhi, line 461: Is this specific to RPL-P or can this solution be safely applied to all ADL as well?

bejeanmo: The hang has only been seen on RPL-P so far. However, on all ADL the UEFI payload will undo the WriteCombining setting.

sindhuragrandhi: Okay, thanks.

sindhuragrandhi (approved):

gdong1 (approved):

PR #1906, opened 2023-05-20
[ADLN/ASL] Fastboot improvement

tsaikevin:
Saved timeout from reading EC. Reduced up to 300ms. Adjusted Stage 1A & 1B region size for ADLN Fastboot feature.

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/Stage1BBoardInitLib/EcSupport.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage1BBoardInitLib/EcSupport.c
@@ -348,7 +348,10 @@ GetBoardId (
   //
   *PlatformId = 0xFF;
 
+#if !defined(PLATFORM_ADLN) && !defined(PLATFORM_ASL)
   GetBoardIdFromEC(&BoardID);
+#endif
+
   if (BoardID == 0xFF){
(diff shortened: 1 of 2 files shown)


Platform/AlderlakeBoardPkg/Library/Stage1BBoardInitLib/EcSupport.c
```diff
@@ -348,7 +348,10 @@ GetBoardId (
   //
   *PlatformId = 0xFF;
 
+#if !defined(PLATFORM_ADLN) && !defined(PLATFORM_ASL)
```
gdong1, line 351: SBL config data is preferred if it is available. For this case, prefer using SiCfgData->EcAvailable.

tsaikevin: @gdong1 BoardId is required before SBL can read config data, 

at the point SBL reading boardId value via EC, the dlf flag is not available(will read platform ID 0 value)

gdong1: oh, yes.  PlatformId is not got yet.

gdong1 (approved):

PR #1907, opened 2023-05-23
fix: update GPIO base table ID check

atharvalele:
A GPIO YAML can refer to any platform ID's GPIO table as the base GPIO table. Since SBL supports upto 32 platform IDs, the number for the check in the if condition should be 32.

### DIFF
--- a/Silicon/CommonSocPkg/Include/Library/GpioLib.h
+++ b/Silicon/CommonSocPkg/Include/Library/GpioLib.h
@@ -23,6 +23,11 @@
 //
 #define GPIO_PADCFG_DW_REG_NUMBER  4
 
+//
+// Max supported platform ID for GPIO base table
+//
+#define GPIO_MAX_BASE_TABLE_PLATFORM_ID  32
+
 /**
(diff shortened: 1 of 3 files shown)


Silicon/CommonSocPkg/Include/Library/GpioLib.h
```diff
@@ -23,6 +23,11 @@
 //
 #define GPIO_PADCFG_DW_REG_NUMBER  4
 
+//
+// Max supported platform ID for GPIO base table
+//
+#define GPIO_MAX_BASE_TABLE_PLATFORM_ID  32
```
rprangar, line 29: Should this be 32 or 31 and the check use "<=" instead of "<"

atharvalele: Checks for platform ID elsewhere in the source code use ``< 32`` , so I think keeping this as 32 will be better

rprangar (approved):
LGTM

gdong1 (approved):

Issue #1913, opened 2023-05-29
LF/CRLF is not coverted to NUL in Multiboot module string


jiaqingz-intel:
When loading Multiboot container, the LF/CRLF is not coverted to NUL in Multiboot module string when booting

The command I used to build container is below, it was copied from https://github.com/intel/meta-acrn/blob/master/docs/slimbootloader.md

```
sudo python3 GenContainer.py create -cl CMDL:cmdline.txt ACRN:acrn.32.out CMD0:sos-cmdline.txt IMG0:vmlinuz -k ./SblKeys/OS1_TestKey_Priv_RSA3072.pem -a RSA3072_PKCS1_SHA2_384 -t MULTIBOOT -o container.bin
```

Both ACRN and SlimBootloader prints the string of the multiboot module as below, followed by chararcters not belonging to string

```
Linux_bzImage
  LZDMh, 
```

Multiboot Spec (https://www.gnu.org/software/grub/manual/multiboot/multiboot.html#Boot-information-format) says both cmdline and module string are null-terminated string. Is not converting to the NUL termintaor expected?

The cmdline is parsed at

https://github.com/slimbootloader/slimbootloader/blob/a34e54e1754d5dd011786b081760d1d3d286b878/PayloadPkg/OsLoader/OsLoader.c#L444-L467

stanleyintel:
Hi,

  As described in the PR, the patch fixes the issue that a mod string shall be a zero-terminated ASCII string (required by multboot spec), but it does not convert newlines to NULL (not required by multiboot spec).

PR #1916, opened 2023-05-30
fix: multiboot mod string in zero-terminated ASCII format #1913

stanleyintel:
Required by multiboot spec (*1), a mod string is a zero-terminated ASCII string. The patch introduces LoadMultibootModString to load mod string from a IMAGE_DATA (allocating a new buffer when requiring to append zero-terminated char).

The patch does not reuse GetFromConfigFile because GetFromConfigFile was designed to be compatible with legacy format (e.g., EOF signature) and truncates new line chars (which is not required by multiboot mod).

For performance, the patch does not run "isascii" check.

Minor changes:
  - Fix typo error (InitMultibootMmap)
  - Declare FreeImageData in Library/IasImageLib.h
  - Dump mod string for debug build

Reference:
1. https://www.gnu.org/software/grub/manual/multiboot/multiboot.html

Verify: EHL CRB
Acked-by: Jiaqing Zhao <jiaqing.zhao@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Include/Library/IasImageLib.h
+++ b/BootloaderCommonPkg/Include/Library/IasImageLib.h
@@ -129,6 +129,21 @@ IasGetFiles (
   OUT  IMAGE_DATA  *Img
   );
 
+/**
+  Free the allocated memory in an image data
+
+  This function free a memory allocated in IMAGE_DATA according to Allocation Type.
+
+  @param[in]  ImageData       An image data pointer which has allocated memory address,
+                              its size, and allocation type.
+
+**/
+VOID
+EFIAPI
+FreeImageData (
+  IN  IMAGE_DATA    *ImageData
+  );
+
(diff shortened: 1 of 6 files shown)


jiaqingz-intel:
Hi, Stanley

I think we can make this easier by finding the first `\n` or `\r` and replace it with `\0` in `MultiBoot->MbModuleData[ModuleIndex].CmdFile.Addr`.

There is no need to create a copy and reassign the string. The code I linked did so because Slimbootloader may append something to the cmdline in `UpdateOsParameters()`, so allocating a larger buffer for it is necessary.

For the multiboot module string, there is no such need, changing the character inplace is enough.

stanleyintel:
> > 
> 
> Oh I missed one thing, if the cmdline file does not contain `\n`, `\r\n` or `\0`, a larger buffer is needed to append the null terminator. How about make this snippet a function instead of coping and pasting it?

A new patch will be submitted with more fix included.

gdong1 (approved):

PR #1931, opened 2023-06-22
fix: [Common] Fix Coverity issues

tsaikevin:
Checked for Null pointer before dereference (CWE 476)

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/SmbiosInit.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/SmbiosInit.c
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2020 - 2022, Intel Corporation. All rights reserved.<BR>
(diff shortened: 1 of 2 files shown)


bejeanmo:
@tsaikevin is an ASSERT good enough to cover this? We have DEBUG_PROPERTY_DEBUG_ASSERT_ENABLED set in PcdDebugPropertyMask even in release mode, but that may not be the case in a non-default use case. I think it's better to return an error status in this case also.

tsaikevin:
> @tsaikevin is an ASSERT good enough to cover this? We have DEBUG_PROPERTY_DEBUG_ASSERT_ENABLED set in PcdDebugPropertyMask even in release mode, but that may not be the case in a non-default use case. I think it's better to return an error status in this case also.

@bejeanmo update the code with error status returning to caller. pls review it.

Either way should work, as long as avoiding pointer dereference. fyi

bejeanmo (changes requested):
Update copyright dates. Otherwise looks good.

gdong1 (approved):

tsaikevin:
> Update copyright dates. Otherwise looks good.

@bejeanmo done. fyi

PR #1932, opened 2023-06-23
fix: [Common] Verify RSDP Signature before update ACPI

tsaikevin:
RSDP table PCD is valid in normal boot path where AcpiInit is called. The table could move in OS which casues issue in S3 path. Verify signature before updating ACPI to prevent above issue.

### DIFF
--- a/BootloaderCorePkg/Stage2/Stage2Support.c
+++ b/BootloaderCorePkg/Stage2/Stage2Support.c
@@ -81,7 +81,7 @@ BoardNotifyPhase (
     PlatformService = (PLATFORM_SERVICE *) GetServiceBySignature (PLATFORM_SERVICE_SIGNATURE);
     // Add perf data to FPDT - SBL Performance Table
     Rsdp = (EFI_ACPI_5_0_ROOT_SYSTEM_DESCRIPTION_POINTER *)(UINTN)PcdGet32 (PcdAcpiTablesRsdp);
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Stage2/Stage2Support.c
```diff
@@ -81,7 +81,7 @@ BoardNotifyPhase (
     PlatformService = (PLATFORM_SERVICE *) GetServiceBySignature (PLATFORM_SERVICE_SIGNATURE);
     // Add perf data to FPDT - SBL Performance Table
     Rsdp = (EFI_ACPI_5_0_ROOT_SYSTEM_DESCRIPTION_POINTER *)(UINTN)PcdGet32 (PcdAcpiTablesRsdp);
-    if (Rsdp != NULL) {
+    if (Rsdp != NULL && Rsdp->Signature == EFI_ACPI_5_0_ROOT_SYSTEM_DESCRIPTION_POINTER_SIGNATURE) {
```
gdong1, line 84: In S3 path, this notification will also be called but PcdAcpiTablesRsdp is still using the default value. I think we just need update SBL boot performance data in normal boot path.
Since current code only updates ACPI FPDT SBL perf table in board notification, and UEFI payload doesn't call the board notification, I will create a patch to enhance it and fix S3 issue.

PR #1960, opened 2023-07-11
fix: [ADLN] fix SPI and I2C mode in ACPI NVS

tsaikevin:
Update same config value as BIOS code.
Address unused value (CWE 563)


sindhuragrandhi (approved):

sindhuragrandhi:
Trigger CI

bejeanmo (changes requested):
Suggest removing these lines and modifying the values directly in mPchSSerialIoSPIMode and mPchSSerialIoI2CMode. It's unclear to write the values from the table and then immediately overwrite them with different values in code.

bejeanmo (approved):

PR #1964, opened 2023-07-17
fix: [ADL] change GPIO CfgData to use virtual group index

vincentchchen:
Different ADL PCH series group GPIO pins into different sets. The current CfgData uses ADLS's GPIO group sets as the base, and overwrites the GrpIdx fields via DLT files for other PCH series. This results in two issues:
1. "CfgDataTool.py export" command failed to extract DLT files from SBL image, since it cannot identify the proper group index in CfgDataExt.bin from the base table in CfgDataInt.bin
2. When using ConfigEditor.py to configure the PadGroup field of GPIO Payload Selection, it will map to a wrong group index.

This patch changes the GPIO CfgData to use virtual alphabetical group index for all PCH series, and the virtual index will be converted to physical group ID accordingly when system configures the GPIO pins.


sindhuragrandhi:
Sometime ago, I believe we decided not to do the virtual to physical mapping method because the groups in Config Editor doesnt match up to what the EDS spec says and customer mgiht get confused. If I remember, we removed this approach on purpose.

@gdong1 @rprangar Please share your inputs on this PR where the virtual to physical mapping of the groups has been re-introduced.

sindhuragrandhi:
@bejeanmo this patch is changing the base file CfgData_Gpio.yaml as well as some of the common files. It might effect RPL GPIO dlt files as well if this PR gets merged.

vincentchchen:
> 

Yes, there is the corresponding changes needed applying on RPL.
https://github.com/intel-innersource/firmware.boot.bootloader.sbl.sblplatform/pull/848

vincentchchen:
> Sometime ago, I believe we decided not to do virtual to physical mapping method because groups in Config Editor doesnt match up to what the EDS spec says and customer might get confused. If I remember, we removed this approach on purpose.
> 
> @gdong1 @rprangar Please share your inputs on this PR where the virtual to physical mapping of the groups has been re-introduced.
> 
> If this approach makes it better, I dont mind it but just want to bring it up so we are aware of the background.

Didn't know about the history; hoping to get more information of the background :)

It seems to me that the sequence, or the GrpIdx, of the GPIO groups is defined in GpioSiLib.c according to the mPchLpGpioGroupInfo[ ] array. I didn't get how the array was mapped to PCH EDS. For example, in ADLP EDS, the GPP_U is not mentioned, but it is still listed in the array occupying the slot [9]. So customer might not be easy to figure out the right group ID from EDS; only with the help from CfgData_GpioPadGroups.yaml. But, ADLP and ADLS have totally different mapping tables.

So, if to keep the physical IDs, maybe the yaml syntax needs to have the capability of !INCLUDE files according to some conditions (e.g. PlatformName).

gdong1:
> So, if to keep the physical IDs, maybe the yaml syntax needs to have the capability of !INCLUDE files according to some conditions (e.g. PlatformName).
Combination of the GPIO settings from different SKU makes the GPIO more complicated in both config tool and code. The initial thought is to have a single SBL image to support different SKUs. Gavin different SKU have different FSP which makes it impossible to have single SBL to support different SKUs. I think it is time to separate the GPIO settings to let each SKU only have its own GPIO settings.

gdong1:
For the group physical mapping, it should be transparent for developer and end user once the initial GPIO table is created. Will also check this.

vincentchchen:
Thanks for the inputs. Will figure out a different design to have only physical tables capable of mapping for different PCH skus.
Let me close the PR first.

Silicon/AlderlakePkg/Library/GpioSiLib/GpioSiLib.c
```diff
@@ -188,6 +188,28 @@ typedef struct {
...
 
+// GPIO group table to convert from virtual alphabetical group index to physical group ID
+CONST UINT8 mPchGpioGroup[3][26] = {
+  { // PCH-N based on GPIO_GROUP_INFO mPchNGpioGroupInfo
+    /*A     B     C     D     E     F     G     H     I     J     K     L     M      */
+    0x02, 0x00, 0x0B, 0x08, 0x0E, 0x0C, 0xFF, 0x07, 0x09, 0xFF, 0xFF, 0xFF, 0xFF,
+    /*N     O     P     Q     R     S     T     U     V     W     X     Y     Z(GPD) */
+    0xFF, 0xFF, 0xFF, 0xFF, 0x03, 0x06, 0x01, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x05
+  },
+  { // PCH-P based on GPIO_GROUP_INFO mPchNGpioGroupInfo
```
vincentchchen, line 199: mPchLpGpioGroupInfo

PR #1984, opened 2023-08-02
fix: [PCIe] Fix assert check

spkamat:
Assert check was one off causing problem for corner cases. Upated the check.

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
@@ -1639,7 +1639,7 @@ PciProgramResources (
       Root->PciBar[BarType - 1].BaseAddress = Address;
       if (Root->PciBar[BarType - 1].Length > 0) {
         ResBase[Index] += Root->PciBar[BarType - 1].Length;
-        ASSERT (ResBase[Index] <= ResLimit[Index]);
+        ASSERT (ResBase[Index] <= (ResLimit[Index] + 1));
       }
 
       ProgramResource (Root, BarType);


BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -1639,7 +1639,7 @@ PciProgramResources (
       Root->PciBar[BarType - 1].BaseAddress = Address;
       if (Root->PciBar[BarType - 1].Length > 0) {
         ResBase[Index] += Root->PciBar[BarType - 1].Length;
-        ASSERT (ResBase[Index] <= ResLimit[Index]);
+        ASSERT (ResBase[Index] <= (ResLimit[Index] + 1));
```
tsaikevin, line 1642: @spkamat does this change to avoid the condition of ResBase[Index] == ResLimit[Index]? can just remove the equal sign like below?
ASSERT (ResBase[Index] < ResLimit[Index]);

spkamat: When base is 0x1000, length is 1000 and limit is 0x1FFF (our current scenario), this assert check is not valid as the address range would still be from 0x1000 to 0x1FFF (1000 locations) but equation would check for 0x1000 <= 0x2000 and fail.

cshur: Yes. The limit address ends at FFFF and base address starts with 4K alignment. This ASSERT means base address should be equal or bigger than previous limit address.   Chirag PR code should be right.  Anyway  in (ResLimit[Index] + 1), no need ().

gdong1 (approved):

tsaikevin (approved):

PR #1995, opened 2023-08-17
MultibootLib: Fix unable to load multiboot image with modules

jiaqingz-intel:
LoadMultibootModString() checks the module index against MultiBoot-> MbModuleNumber, which value has not been assigned when called, causing failure when loading multiboot image with modules. Remove the check to fix it.

Fixes c9d70e74dcfe ("fix: multiboot mod string in zero-terminated ASCII format #1913")


gdong1 (approved):
It looks good to me.

PR #1997, opened 2023-08-18
fix:[RPL-PS] add board id to support RVP

kdbarnes-intel:
Add Platform Flag for RPL-PS to add ADL-PS RVP board


bejeanmo (changes requested):
Update copyright date.

bejeanmo (approved):

PR #1998, opened 2023-08-20
enhance CfgData to support board extension yaml

vincentchchen:


sindhuragrandhi (approved):

BootloaderCorePkg/Tools/ConfigEditor.py
```diff
@@ -793,6 +795,44 @@ def load_from_delta(self):
...
+            'AdlSSdm5' : 'adls',
+            # RPL series
+            'RPLS S14' : 'rpls',
+            'RPLS S17' : 'rpls',
+            'RPLP CRB' : 'rplp',
+            'RPLPDDR5' : 'rplp',
+            'RPLP LP5' : 'rplp',
+            'RplPsCrb' : 'rplps',
+            'RplPsRvp' : 'rplps'
+        }
```
gdong1, line 820: ConfigEditor.py is a common tool. It is not expected to add platform name info. We need think about other solution.

BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -2,7 +2,7 @@
 ## @ BuildUtility.py
 # Build bootloader main script
 #
```
gdong1, line 4: Given we have different FSP for different SKU, it is not possible to use a single image to support different silicon SKU.
So there is no point to have a combine all the GPIO for different SKUs into one GPIO table.
Maybe we could let each SKU has a CfgDataDef.yaml (e.g. CfgDataDefRpls.yaml CfgDataDefRplp.yaml, etc.). This way we could make GPIO config simple.

BootloaderCorePkg/Tools/ConfigEditor.py
vincentchchen: Got it, this patch is specific to the use case when user opens the "raw" yaml file under Platform\AlderlakeBoardPkg\CfgData. We may expect user to open the "cooked" yaml file under Build\BootloaderCorePkg\DEBUG_VS2015x86\FV, which is also the one included in our release package, then user won't encounter the issue.
If we have separate CfgDataDef.yamls for different platforms as your suggestion, then could also prevent from this issue :).
Will create another pull request soon.

BootloaderCorePkg/Tools/BuildUtility.py
vincentchchen: With this PR, the outcome CfgDataDef.yaml is specific for the platform of the specified board name. That is, this PR reuse the commom part of CfgDataDef.yaml, and include the GPIO part for specified board during build time. While it's true that having separate CfgDataDef.yamls will be a simpler design, also can avoid the issue on ConfigEditor.
We may need to have separate CfgDataDef.yaml as well as separate CfgData_PayloadSelection.yaml
Will create another pull request soon based on the idea :)

vincentchchen: Consider to have only three CfgDataDef.yaml for ADLN, ADLP, ADLS
e.g. ADLPS, RPLP could be able to share the same yaml with ADLP

PR #2005, opened 2023-08-22
Support booting multiple container images

jiaqingz-intel:
- [x] Load image from separate partition
- [x] Add new container type for pure data
- [x] Update multiboot info struct with extra images
- [x] Shell support

Known issue:
* ~~Break RTCM compatibility as following lines are removed~~ [Workaround]

Tested on QEMU


PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -299,7 +311,9 @@ GetBootImageFromFs (
     goto Done;
   }
 
-  if (LoadedImage->LoadImageType == LoadImageTypeExtra0) {
+  // TODO: RTCM compatibility
+  // if (LoadedImage->LoadImageType == LoadImageTypeExtra0) {
+  if (0) {
```
gdong1, line 316: TCC binaries (including RTCM) is dropped from ADL/RPL. To keep it works for old platforms, let's keep it until TCC is removed from all platforms. It looks all the RTCM file name is "/boot/sbl_rtcm", maybe we could check if file name is rtcm and extra0, we could keep it as-is.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -807,22 +823,26 @@ EFIAPI
...
   CONTAINER_IMAGE           *ContainerImage;
+  EFI_HANDLE                 FsHandle;
 
   ASSERT (OsBootOption != NULL);
 
   BootImage = OsBootOption->Image;
   BootFlags = OsBootOption->BootFlags;
+  BootFromFs = (OsBootOption->DevType != OsBootDeviceSpi) &&
+               (OsBootOption->DevType != OsBootDeviceMemory) &&
+               (OsBootOption->FsType < EnumFileSystemMax);
```
gdong1, line 845: It looks the default FsType is 0 (EnumFileSystemTypeFat), so we could not use FsType to get BootFromFs.

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -857,8 +877,8 @@ LoadBootImages (
     ContainerImage = &BootImage[Index].ContainerImage;
     if ((ContainerImage->Indicate == '!') && (ContainerImage->BackSlash == '/')) {
       Status = GetBootImageFromIfwiContainer (OsBootOption, LoadedImage);
-    } else if (FsHandle != NULL) {
-      Status = GetBootImageFromFs (FsHandle, OsBootOption, LoadedImage);
+    } else if (BootFromFs) {
```
gdong1, line 880: Maybe we could use filename as below to know if it is boot from file system. 
} else if ((AsciiCharToUpper(OsBootOption->Image[Index].FileImage.FileName[0]) >= 'A') && AsciiCharToUpper(OsBootOption->Image[Index].FileImage.FileName[0]) <= 'Z') {

PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -1217,20 +1190,20 @@ BootOsImage (
     }
 
     //
-    // Init File System
+    // Parse A/B Slot Info
     //
     if (!EFI_ERROR (Status)) {
-      Status = InitBootFileSystem (OsBootOption, HwPartHandle, &FsHandle);
+      Status = ParseAbSlot (OsBootOption, HwPartHandle);
```
gdong1, line 1196: Why we add A/B slot info here? If we need update A/B support, we could create a separate PR.
BTW, in the boot option, Image[2] is for A/B support, please check if you could use existing logic.

PayloadPkg/OsLoader/LoadImage.c
jiaqingz-intel: Fixed

PayloadPkg/OsLoader/OsLoader.c
gdong1: Got it. This PR just moves A/B support out of InitBootFileSystem(). It is expected change so that it could work for all kind of image loading type.  Current SBL just provides the method to support OS A/B, it could be updated in actual use case late.

PayloadPkg/OsLoader/LoadImage.c
jiaqingz-intel: Here `OsBootOption` is `OS_BOOT_OPTION`, its `FsType` comes directly from boot option. `FILE_IMAGE_LOCATION->FsType` is not used here.

jiaqingz-intel:
Added more changes, please help review.

PayloadPkg/OsLoader/LoadImage.c
gdong1: In one boot option, SBL already supports to load raw image from different partition. With this change, SBL could support to load image by file name from different partition. So it is possible in in boot option some images are loaded by file name, and other image are loaded by LBA address. SBL might read each image differently, so the FsType and SwPart from Os_BOOT_OPTION should not be used, or it is used only for the normal boot image.

jiaqingz-intel: Got it.

PayloadPkg/OsLoader/LoadImage.c
jiaqingz-intel: Here checking `OsBootOption->Image[Index].LbaImage.Valid == 1` might be better. `\1` cannot be the first char in filename.

jiaqingz-intel:
Is it ok to print boot option with extra images like this? 

image

jiaqingz-intel:
Linux ADLPS_X86_DEBUG fails with `the required fv image size 0x574d0 exceeds the set fv image size 0x57000`. @gdong1 Can you help on this?

gdong1:
> Is it ok to print boot option with extra images like this?
> 
> ```
> Idx|ImgType|DevType|DevNum|Flags|HwPart|FsType|SwPart|File/Lbaoffset
>   0|      0|  SATA |    0 |   8 |   FF | AUTO |    3 | acrn.container *Current
>                                 Extra0 | AUTO |    2 | sbl_vmlinuz
>                                 Extra1 | RAW  |    0 | 0x800
>   1|      0|  NVME |    0 |   0 |    0 | AUTO |    0 | container.bin
>   2|      0|   USB |    0 |   0 |    0 | AUTO |    0 | container.bin
> ```

It looks good to me.

gdong1:
> Linux ADLPS_X86_DEBUG fails with `the required fv image size 0x574d0 exceeds the set fv image size 0x57000`. @gdong1 Can you help on this?

Yes, you could add below 2 lines to Platform\AlderlakeBoardPkg\BoardConfig.py
        self.OS_LOADER_FD_SIZE    = 0x00058000
        self.OS_LOADER_FD_NUMBLK  = self.OS_LOADER_FD_SIZE // self.FLASH_BLOCK_SIZE

gdong1 (commented):
The change looks good to me in general with the build failure fix.

gdong1 (approved):

PR #2007, opened 2023-08-24
feat: [ADL] separate CfgDataDef.yaml for different PCH series

vincentchchen:
ADL groups GPIO pins into different sets for different PCH series. The current CfgData uses ADLS's GPIO group sets as the base, and overwrites the GrpIdx fields via DLT files for other PCH series. This results in two issues:
1. "CfgDataTool.py export" command failed to extract DLT files from SBL image, since it cannot identify the proper group index in CfgDataExt.bin from the base table in CfgDataInt.bin
2. When using ConfigEditor.py to configure the PadGroup field of GPIO Payload Selection, it will map to a wrong group index.

This patch separates CfgDataDef.yaml for each PCH series
- add board extension yaml files for Adln, Adlp, Adls
  * CfgDataDef*.yaml
  * CfgData_Gpio_*.yaml
  * CfgData_PayloadSelection_*.yaml
- assign an invalid/unique GrpIdx for the unused GPIO group
  * e.g. 1F, 1E, ..
- adjust GrpIdx of CfgDataExt_Upx12.dlt based on Adlp
- revise the payload selection information in DLT files

This patch also fixes the issue:
- When ConfigEditor opens CfgDataDef.yaml more than once, the config changes will not be caught by "Save Config Changes to Delta File" It is because the "info" argument of build_cfg_list() in GenCfgData.py will retain its value in the lifetime of ConfigEditor, and the offset will overflow. So need to reset the offset for a fresh load_yaml().


BootloaderCorePkg/Tools/BuildUtility.py
```diff
@@ -536,6 +536,20 @@ def copy_expanded_file (src, dst):
...
+        # ADL series
+        'adln'  : 'Adln',
+        'adlp'  : 'Adlp',
+        'adlps' : 'Adlp',
+        'adls'  : 'Adls',
+        # RPL series
+        'rplp'  : 'Adlp',
+        'rplps' : 'Adlp',
+        'rpls'  : 'Adls',
+    }
```
gdong1, line 551: BuildUtility.py is a common tool. I think platform specific settings should come from BoardConfig.py.
how about add below line in the BoardConfig.py?
        self.CFG_DATA_MAIN_FILE   = 'Platform/AlderlakeBoardPkg/CfgData/CfgDataDef_rpls.inf'

vincentchchen: Will do accordingly. Always thanks for pointing out the design concept

PR #2009, opened 2023-08-25
fix: [ADL] Check PcdTccEnabled at compile time

arghyabi:
Update the PcdTccEnabled check from run time to compile time. 
And move the PcdTccEnabled pcd to fixedPcd as declared in BootloaderCorePkg.dsc file.
This change is needed for those platforms where TCC is not supported and CfgData_Tcc.yaml file is not used in CfgDataDef.yaml.

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
@@ -271,7 +271,9 @@ PlatformUpdateAcpiTable (
   VOID                        *FspHobList;
   PLATFORM_DATA               *PlatformData;
   FEATURES_CFG_DATA           *FeaturesCfgData;
+#if FixedPcdGet8 (PcdTccEnabled)
   TCC_CFG_DATA                *TccCfgData;
+#endif
   EFI_STATUS                   Status;
   EFI_ACPI_6_3_FIXED_ACPI_DESCRIPTION_TABLE *FadtPointer;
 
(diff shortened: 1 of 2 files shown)


Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
```diff
@@ -271,7 +271,9 @@ PlatformUpdateAcpiTable (
   VOID                        *FspHobList;
   PLATFORM_DATA               *PlatformData;
   FEATURES_CFG_DATA           *FeaturesCfgData;
+#if FixedPcdGet8 (PcdTccEnabled)
   TCC_CFG_DATA                *TccCfgData;
+#endif
```
gdong1, line 276: If CfgData_Tcc.yaml file is not used in CfgDataDef.yaml, then when PcdTccEnabled is set to TRUE it could cause build failure.  Will you remove TCC_ENABLE from boardconfig.py?

arghyabi: Yes, we will remove TCC_ENABLE from that platform-specific boardconfig.py where TCC is not supported.

gdong1 (approved):
ok.

Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
gdong1: got it.

PR #2020, opened 2023-09-05
[ICX-D] Fix wrong PCR event digest for ACM Authority PCR Event

### LINKED ISSUE #2013: [ICX-D] Not being able to reproduce PCR 7 value using Authority PCR event log
Hi,

I noticed an issue when enabling authority data measurement (PCR 7) and Boot Guard P3 on an ICX-D HCC system. The PCR 7 value cannot be reproduced using the digest value logged in debug output or TCG event log in memory.

After modifying `CaculateAuthorityPCRExtendValue`, the digest value is logged correctly, and PCR 7 value can be reproduced using that value.
1. Getting ACM key hash from MSR 0x20::0x23.
2. Padding BP key hash to 384 bit.

Thanks.

cans-arista:
Previously, the digest calculated by Slimboot cannot be used to reproduce the PCR 7 value. There're two issues:
1. ACM public key hash should be read from MSRs (0x20::0x23).
2. BP key hash is padded to 384-bit in ACM.

This fix is tested on an ICX-D HCC system. (Fix #2013)


cans-arista:
I was using ACM 1.2.2 and 1.2.3 and the results were the same.
You mentioned "PCRs". Which ones are you referring to? Is the PCR 7 value reproducible on your system before the change?

cans-arista:
Hi @sean-m-mcginn ,

I'm using Slimboot to boot into a Linux-based bootloader.

cans-arista:
It should not be a Windows-related issue.
The value extended to PCR 7 does not match the Slimboot logs. I've tried to read the SPI data in/out TPM, and after patching the fix, I'm able to get the actual digest value that goes into TPM.

cans-arista:
Hi @sean-m-mcginn,

Thanks for working on the fix. Yeah, it looks good to me. I will try to test the patch when it is ready.

PR #2021, opened 2023-09-05
[MTL] Fix SMM hang

sean-m-mcginn:
This change fixes an SMM hang that can occur in MTL and RPL universal payloads

A future patch will be created which re-enables the board notify in the universal payload

### DIFF
--- a/BootloaderCorePkg/Stage2/Stage2.c
+++ b/BootloaderCorePkg/Stage2/Stage2.c
@@ -252,10 +252,14 @@ NormalBootPath (
 
   PayloadId = GetPayloadId ();
 
-  // Force board notify to false here to avoid going
-  // into a low power state that prevents SMM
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -254,19 +251,12 @@ NormalBootPath (
...
-    // will be used to indicate if it will handle FSP notifications.
-    CallBoardNotify = FALSE;
-  } else {
-    CallBoardNotify = TRUE;
-  }
+
+  // Force board notify to false here to avoid going
+  // into a low power state that prevents SMM
+  // initialization in MTL and RPL universal payloads
+  CallBoardNotify = FALSE;
```
lsubash, line 258: i believe we should have a condition for other payloads apart from osloader and UEFIpayload.

PR #2023, opened 2023-09-06
Tool: Update the build Tools

gdong1:
To search config data main file, search from os.environ['PLT_SOURCE'] firstly, then search from os.environ['SBL_SOURCE'] to make sure PLT_SOURCE has high priority.


gdong1:
it looks this is not required.

PR #2027, opened 2023-09-15
Fix authority measure calculation for PCR[7]

sean-m-mcginn:
This change corrects the authority measure calculation in the common BootGuardTpmEventLogLib

If BtG KM is signed using RSA 3K key, a SHA384 hash should be added to the authority PCR structure

If BtG KM is signed using RSA 2K key, a SHA256 hash (padded to 384 bits) should be added to the authority PCR structure

The entire authority PCR structure should be initialized to zero before being populated

### DIFF
--- a/Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
+++ b/Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
@@ -58,11 +58,10 @@ typedef struct {
 //
 // ACM definition
 //
-
-#define ACM_KEY_HASH_MMIO_ADDR_0  0xFED30400
-#define ACM_KEY_HASH_MMIO_ADDR_1  (ACM_KEY_HASH_MMIO_ADDR_0 + 8)
-#define ACM_KEY_HASH_MMIO_ADDR_2  (ACM_KEY_HASH_MMIO_ADDR_0 + 16)
-#define ACM_KEY_HASH_MMIO_ADDR_3  (ACM_KEY_HASH_MMIO_ADDR_0 + 24)
+#define ACM_KEY_HASH_MSR_ADDR_0  0x20
+#define ACM_KEY_HASH_MSR_ADDR_1  0x21
+#define ACM_KEY_HASH_MSR_ADDR_2  0x22
+#define ACM_KEY_HASH_MSR_ADDR_3  0x23
 #define ACM_PKCS_1_5_RSA_SIGNATURE_SIZE  256
 #define ACM_MODULE_TYPE_CHIPSET_ACM      2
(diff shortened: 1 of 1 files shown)


cans-arista:
Thanks for the fix!
I tested this patch on an ICX-D platform with BtG profile 3 and Authority Data Measurement enabled. PCR 7 value can be reproduced using the log generated by Slimboot.
Tested the changed event type for Authority Data Measurement event with tpm2_eventlog (included in tpm2-tools (5.2-1build1)). The PCR 7 event TCG log can be decoded with the tool. (Previously it would fail with event data not correct)
(Fix #2013)

Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
```diff
@@ -432,7 +431,7 @@ typedef struct {
 typedef struct {
   UINT64 AcmPolicySts;
   UINT16 AcmSvn;                               // ACM_SVN from ACM Header
-  UINT8  AcmKeyHash[SHA384_DIGEST_SIZE];       // The hash of the key used to verify the ACM (SHAxxx)
+  UINT8  AcmKeyHash[SHA256_DIGEST_SIZE];       // The hash of the key used to verify the ACM (SHAxxx)
```
lsubash, line 434: note, newer client platforms has ACH keys hash SHA384

Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
```diff
@@ -1166,35 +1178,41 @@ CreateDetailPcrEvent (
...
-      }
+    NewEventHdr.PCRIndex  = 7;
+    NewEventHdr.EventType = EV_PLATFORM_CONFIG_FLAGS;
+    CaculateAuthorityPCRExtendValue (Sha256);
+
+    ZeroMem (Digests, sizeof(TPML_DIGEST_VALUES));
+    if ((ActivePcrBanks & HASH_ALG_SHA256) != 0) {
+      Digests->digests[Digests->count].hashAlg = TPM_ALG_SHA256;
+      CopyMem (Digests->digests[Digests->count].digest.sha256, Sha256, SHA256_DIGEST_SIZE);
+      Digests->count++;
```
lsubash, line 1205: We have to support for SHA384 and SM3 also.

Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
lsubash: Yes, this lib should not have issue.

gdong1 (approved):

PR #2030, opened 2023-09-20
(WIP) Check the openssl location

gdong1:
Recently windows VM updated openssl which caused SBL build failure. This patch will check possible openssl location and version.


gdong1:
The openssl comes back. this pr is not required.

PR #2037, opened 2023-09-29
feat: Added FIT4 Entry in FIT table

arghyabi:
FIT4 Entry is needed for PCH-less Server platforms to execute CBnT and/or PFR. By default, FIT4 Entry will not come into the scenario. It needs to enable from BoardConfig.py for specific platforms.


spkamat (approved):

cshur (approved):
LGTM

PR #2040, opened 2023-10-06
fix: [ADLN] program memory range above max address uncacheable

tsaikevin:
Windows memory test failed with write back cacheable setting for memory address above maximum usable address.
The fix programs one MTRR register for memory range above max usable address to override default MTRR register setting after temp ram exit.


sindhuragrandhi:
Is this applicable for all projects?

tsaikevin:
> Is this applicable for all projects?

@sindhuragrandhi this issue is related to windows memory test and only reported to ADLN board at the moment.

The change will not benefit to project doesn't report this issue since it changes cache setting of memory space that doesn't exist. So it's added per need basis now.

gdong1 (approved):

PR #2042, opened 2023-10-09
fix: [MTL] Correction on UPD naming of PlatformDebugOption

ssbyna:
The UPD name of PlatformDebugConsent has been changed to PlatformDebugOption and by default is disabled.


koktong-ong (approved):
looks good to me

bejeanmo (approved):

PR #2044, opened 2023-10-10
feat: Addition of GPIO V2 library support

Karuppa-samy:
GPIO V2 library uses sideband access to read and write a particular GPIO also adds support for Virtual GPIO (These GPIO can not have external connections,These are mainly used to integrate discrete GPIO devices present on board)

### DIFF
--- a/Silicon/CommonSocPkg/Include/GpioV2Config.h
+++ b/Silicon/CommonSocPkg/Include/GpioV2Config.h
@@ -1,0 +1,101 @@
+/** @file
+  Header file for GpioConfig structure used by GPIOV2 library.
+
+  Copyright (c) 2023, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+#ifndef _GPIOV2_CONFIG_H_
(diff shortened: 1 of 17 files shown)


ywho79:
Are these changes verified on Meteor Lake?

Karuppa-samy:
@ywho79 Yes @jinjhuli has tested the patch in Meteor Lake

sindhuragrandhi:
@Karuppa-samy in all of the reads and write in the P2sblib, Im not  seeing case P2SbMsgAccess, can you please add this case as well. This will be used for other use-cases as well as other MTL SKUs.

jinjhuli:
> @ywho79 Yes @jinjhuli has tested the patch in Meteor Lake

Will submit patch to amend if any changes is require in future.

Karuppa-samy:
> @Karuppa-samy in all of the reads and write in the P2sblib, Im not seeing case P2SbMsgAccess, can you please add this case as well. This will be used for other use-cases as well as other MTL SKUs.

@sindhuragrandhi  P2SbMsgAccess can be added on top of this patch when need arise in future.. there won't be any conflict

sindhuragrandhi (approved):

PR #2045, opened 2023-10-10
Fix: [Common] Support Firmware Update(FWU) with multiple device instance

Karuppa-samy:
This patch supports reading of capsule image from a configured device instance, During FWU process incase of multiple device instances.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
+++ b/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
@@ -1,7 +1,7 @@
 /** @file
   This file contains the implementation of FirmwareUpdateLib library.
 
-  Copyright (c) 2017 - 2022, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2017 - 2023, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -406,6 +406,7 @@ GetCapsuleImage (
   UINT8                   HwPart;
   UINT8                   StartPart;
   UINT8                   EndPart;
+  UINT8                   DevInstanace_configured;
 
   Status = EFI_UNSUPPORTED;
   CapsuleInfo = NULL;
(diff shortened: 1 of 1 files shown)


PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -427,41 +428,48 @@ GetCapsuleImage (
     DEBUG((DEBUG_ERROR, " CapsuleInfo not found \n"));
     return EFI_NOT_FOUND;
   }
+  DevInstanace_configured = CapsuleInfo->DevInstance;
+  //DevInstance is enumerated upto configured DevInstance in "CfgData_CapsuleInformation.yaml"
+  for (CapsuleInfo->DevInstance = 0; CapsuleInfo->DevInstance <= DevInstanace_configured; CapsuleInfo->DevInstance ++)
```
gdong1, line 433: From this pacth, it would firstly to try instance 0 to get capsule image even device instance 1 is configured for capsule image, is it expected?

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -427,41 +428,48 @@ GetCapsuleImage (
...
+  {
+    Status = InitBootDevice(CapsuleInfo);
+    if (EFI_ERROR(Status)) {
+      return Status;
```
gdong1, line 437: And if we supports multiple device, should try next one if there is error?

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
Karuppa-samy: Yes, That's the expected behavior.

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
Karuppa-samy: No, It will increment only the instance of configured device and not multiple device

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
gdong1: For example, user configured to read capsule image from NVMe 1, but it read a capsule from NVMe 0 if both NVMe have capsule image on it. I think we should try to read capsule image from the specified/configured device first.

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
gdong1: For example, if there are 2 NVMe instances, it means the system could have 2 NVMe devices installed. When it tried the first instance/device failed, it should try another instance/device, right?

gdong1: should we close this PR if it is not required?

PR #2047, opened 2023-10-11
Update FPDT Table

gdong1:
Update the FPDT header length to include the length of boot records. This way we could easily get all the boot records from dumped ACPI FPDT table.


ywho79 (approved):
LGTM

randylintw:
LGTM ~

PR #2051, opened 2023-10-17
fix: [AZB] fix GPIO definition issue from linux

arghyabi:
_LP_SUPPORT was disabled as a result,  GPIO programming was faulty from the Linux side. Enable the _LP_SUPPORT to select the proper ACPI to fix the issue.


spkamat (approved):

ywho79 (approved):
LGTM

PR #2055, opened 2023-10-19
[ADLP] Update FSP/VBT for MR5 release

vincentchchen:
- update FSP version to IoT ADL-P MR5 (0C.01.B0.50)
- update VBT version to 251
- Update platform version to 1.5


vincentchchen:
The QEMU check failed due to the added VbtAdlP.json contains Tab characters.
The VbtAdlP.json file is copied as is, so it is not modified against the errors.

Platform/AlderlakeBoardPkg/VbtBin/VbtAdlP.dat
gdong1, line 1: Is this file same with the one from here https://github.com/intel/FSP/tree/master/AlderLakeFspBinPkg/IoT/AlderLakeP/VBT?
Maybe we should get from there if it is same. (Same as FSP),

vincentchchen: Yes, they are the same. I have updated this PR to copy the JSON file from FSP repo.
Previously, sometimes, the FSP repo didn't update VBT files along with FSP files, so we may not get the latest one, especially  not easy to manage for different ADL SKUs.
Anyway, we can take extra care of this VBT part when FSP files are being updated for new MRs.

gdong1: ok. looks good to me.

gdong1 (approved):

PR #2057, opened 2023-10-20
[ICX-D] Configuration knob for flash descriptor verification

pqueirosint:
Adding a dedicated stitching option to create
IFWI images with flash descriptor verification enabled. and signed accordingly.

### DIFF
--- a/Platform/IdavilleBoardPkg/Script/StitchIfwiConfig.py
+++ b/Platform/IdavilleBoardPkg/Script/StitchIfwiConfig.py
@@ -159,6 +159,12 @@ class StitchIfwiConfig(StitchIfwiConfigBase):
                     ('./Debug/DelayedAuthenticationModeConfiguration/DelayedAuthMode', value),
                 ])
 
+            if 'fd0v' in plt_params_list:
+                xml_change_list.extend([
+                    ('./PlatformProtection/FD0Verification/FD0VEnabled', "0xF"),
+                    ('./UepConfiguration/IntelFpfTable/FDV_EN_LLDD', "0xF"),
+                ])
+
         return xml_change_list
 
 def get_stitch_ifwi_config_instance():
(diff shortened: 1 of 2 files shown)


Platform/IdavilleBoardPkg/Script/StitchIfwiConfig.py
```diff
@@ -159,6 +159,12 @@ def get_xml_change_list (self, platform, plt_params_list):
                     ('./Debug/DelayedAuthenticationModeConfiguration/DelayedAuthMode', value),
                 ])
 
+            if 'fd0v' in plt_params_list:
+                xml_change_list.extend([
+                    ('./PlatformProtection/FD0Verification/FD0VEnabled', "0xF"),
+                    ('./UepConfiguration/IntelFpfTable/FDV_EN_LLDD', "0xF"),
+                ])
+
```
gdong1, line 167: The change looks good to me.  
User might not know about this stitch parameter without help info. It would be great to add it to align with other platforms (refer check_parameter in other script).

gdong1 (approved):

Platform/IdavilleBoardPkg/Script/StitchIfwiConfig.py
pqueirosint: Hi @gdong1

Thanks for taking a look. Agree with the lack of documentation for the currently implemented options.
Either user knows what they are not they need to guess from the fit options being impacted.

IdavilleBoardPkg does not have  check_parameter implemented at all. Probably idea was not introduced.
I guess I can copy past and introduce check_parameter for Idaville as well. Would go a bit more beyond the scope i had intended but I can do if that was your idea.

Just let me know what you would like to have.
Cheers.
/pq

gdong1: yes. It would be great if you could add check_parameter as a separate PR when you have time.

PR #2062, opened 2023-10-23
feat: Enhance firmware Update

gdong1:
If the component in SPI flash could not be memory-mapped, it need access using SPI library.
If the component is registered using memory address, it means it need use SPI library to access. It would read required info using SPI library.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
@@ -709,15 +709,18 @@ UpdateContainerComp (
   IN EFI_FW_MGMT_CAP_IMAGE_HEADER  *ImageHdr
   )
 {
-  EFI_STATUS        Status;
-  UINT32            ContainerName;
-  UINT32            ComponentName;
-  UINT32            ComponentBase;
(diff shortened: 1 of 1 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
```diff
@@ -736,14 +739,28 @@ UpdateContainerComp (
   // Update the component
   //
   ContainerHdr = (CONTAINER_HDR *)(UINTN)ContainerEntryPtr->HeaderCache;
-  //
-  // Component base = Container base + data offset from container base + offset of component inside container
-  //
-  ComponentBase = ContainerEntryPtr->Base + ContainerHdr->DataOffset + ComponentEntryPtr->Offset;
+
+  if (ContainerEntryPtr->Base >= 0xF0000000) {
```
bejeanmo, line 743: As a future enhancement, we could get this MMIO flash limit more dynamically and avoid hard coding it.

gdong1: 0xF0000000 (256MB flash space) should be safe. RAM memory will not use this range.

bejeanmo (approved):

PR #2087, opened 2023-12-15
fix: [MTL] Update EC ACPI table

jinjhuli:
Update latest EC ACPI table reference
from BIOS.


Platform/MeteorlakeBoardPkg/AcpiTables/EcSsdt/Devices/Bat1Real.asl
```diff
@@ -151,6 +152,24 @@ Device (BAT1) {
       }
     }
   }
+
```
ywho79, line 155: If this part is not needed let's not include it

jinjhuli: done

ywho79 (approved):
LGTM

PR #2091, opened 2023-12-19
[MTL] Add Optional BtG Authority Measure

sean-m-mcginn:
This change accomplishes the following:
- Logs BtG authority measure events in TPM event log whenever enabled
- Cleans up the equivalent code for earlier platforms

### DIFF
--- a/Silicon/MeteorlakePkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
+++ b/Silicon/MeteorlakePkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
@@ -48,6 +48,11 @@ typedef struct {
 //
 // ACM definition
 //
+#define ACM_KEY_HASH_MSR_ADDR_0                         0x20
+#define ACM_KEY_HASH_MSR_ADDR_1                         0x21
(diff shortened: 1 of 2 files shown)


lsubash (approved):

Silicon/MeteorlakePkg/Library/BootGuardLibCBnT/BootGuardTpmEventLogLib.c
```diff
@@ -1018,6 +1041,182 @@ CreatePolicyDataMeasurementEvent (
...
+  UINT8                                    *AuthorityPcrDataPtr;
+  UINT32                                   AuthorityPcrDataSize;
+  UINT8                                    *CurrPos;
+  UINT16                                   KeyModulusSize;
+  UINT8                                    MaxModulusExpo[RSA_KEY_SIZE_3K/8 + 4];
+  UINT16                                   Index;
+
+  Acm = FindAcm ();
+  ASSERT (Acm != NULL);
+  if (Acm == NULL) return FALSE;
```
gdong1, line 1073: Following coding style, it could be changed to:
if (Acm == NULL) {
  return FALSE;
}
ASSERT could be removed since there is a check in line 1073. you could add debug info message if you want to capture error message. Same comments to similar code in this function.

PR #2093, opened 2023-12-21
[UP7000] Enable UP7000 ADLN50 basic boot

ssbyna:
Add support for UP SQUARED PRO 7000 EDGE ADLN based board. The PCIe M.2 slot CN2 on the board is able to detect NVMe SSD. Debug output is enabled at header CN14 on the board(e.g. UART1) Tested to boot with OS loader payload and UEFI payload.

Build with following command:

python BuildLoader.py build adln50

To stitch the SlimBootloader.bin with IFWI uses StitchLoader.py script with '-p' as given below:

python Platform/AlderlakeBoardPkg/Script/StitchLoader.py -i <BIOS_IMAGE_NAME> -s Outputs/adln50/SlimBootloader.bin -o sbl_up7000adln50_ifwi.bin -p 0xAA000106


ywho79 (approved):
Looks good to me

Platform/AlderlakeBoardPkg/BoardConfigAdlN50.py
```diff
@@ -0,0 +1,490 @@
...
+##
+# Import Modules
+#
+import os
+import sys
+import time
+
+sys.dont_write_bytecode = True
+sys.path.append (os.path.join('..', '..'))
+from BuildLoader import *
```
gdong1, line 18: SBL supports board overridden. It would be good to use that to give another example on how to support ADLN50 board based on ADLN to provide necessary override instead providing a full config file.

ssbyna: hi guo, 

this boardconfigadln50.py file is newly created because the bios size for adln50 is relatively small compared to the normal adln, so the size is adjusted to match the bios size (around 10KB). 

if using the boardconfigadln.py, later during the stitching part would get error as we can't auto-adjusted the bios region size to fit in the binary.

gdong1: I mean you could to use boardconfigadln50.py, and let boardconfigadln50.py based on boardconfigadln.py to only have overridden items to save hundreds of line code.

ssbyna: hi @gdong1 , amended the boardconfigadln50.py to be configured based on boardconfigadln.py, indeed saving lots of lines, thank you for you input. Tested and verified the latest changes is bootable to both ubuntu & windows.

gdong1: Thanks for the update. Please fix the patch check failures.
From failure logs as below, it looks there are spaces in empty line
Trailing whitespace found
   File: Platform/AlderlakeBoardPkg/BoardConfigAdlN50.py
   Line:         
 * Trailing whitespace found
   File: Platform/AlderlakeBoardPkg/BoardConfigAdlN50.py
   Line:

gdong1 (approved):

PR #2101, opened 2024-01-04
feat: Validate BMP headers to protect against LogoFAIL.

bejeanmo:
The SBL logo is verified as part of Stage2 verification, so untrusted logos won't be parsed, but it's still good to prevent dereferencing BMP header pointers that may fall outside of BMP file.


lsubash (approved):
looks good

gdong1 (approved):

PR #2112, opened 2024-01-16
fix: [TPM] correcting cfg data and key hash

aborwank:
issue: https://jira.devtools.intel.com/browse/IOTGSECSW-2031

### DIFF
--- a/BootloaderCorePkg/Stage1B/Stage1B.c
+++ b/BootloaderCorePkg/Stage1B/Stage1B.c
@@ -676,8 +676,6 @@ ContinueFunc (
   TPMI_ALG_HASH               MbTmpAlgHash;
   HASH_STORE_TABLE            *KeyHashBlob;
   CDATA_BLOB                  *CfgDataBlob;
-  EFI_PLATFORM_FIRMWARE_BLOB  KeyHashFwBlob;
-  EFI_PLATFORM_FIRMWARE_BLOB  CfgDataFwBlob;
 
   Stage1bParam   = (STAGE1B_PARAM *)Context1;
   OldLdrGlobal = (LOADER_GLOBAL_DATA *)Context2;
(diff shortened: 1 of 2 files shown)


BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -731,8 +730,8 @@ ContinueFunc (
                     MbTmpAlgHash,
                     Stage1bParam->ConfigDataHash,
                     EV_PLATFORM_CONFIG_FLAGS,
-                    sizeof(CfgDataFwBlob),
-                    (UINT8 *)&CfgDataFwBlob);
+                    (UINT32)CfgDataFwBlob.BlobLength,
```
gdong1, line 733: How about directly use CfgDataBlob->UsedLength and CfgDataFwBlob could be removed?

BootloaderCorePkg/Stage1B/Stage1B.c
```diff
@@ -744,8 +743,8 @@ ContinueFunc (
                     MbTmpAlgHash,
                     Stage1bParam->KeyHashManifestHash,
                     EV_EFI_PLATFORM_FIRMWARE_BLOB,
-                    sizeof(KeyHashFwBlob),
-                    (UINT8 *)&KeyHashFwBlob);
+                    (UINT32)KeyHashFwBlob.BlobLength,
```
gdong1, line 746: How about directly use KeyHashBlob->UsedLength and remove variable KeyHashFwBlob?

gdong1:
Good catch the issue and have a fix. It looks there is another place (https://github.com/slimbootloader/slimbootloader/blob/2648eb75f839d9c88607117c2565c04ad0ae510f/BootloaderCommonPkg/Library/TpmLib/TpmLib.c#L1015) that has similar issue.  Could you help fix it in this PR?

BootloaderCorePkg/Stage1B/Stage1B.c
aborwank: done

BootloaderCorePkg/Stage1B/Stage1B.c
aborwank: done

aborwank:
> Good catch the issue and have a fix. It looks there is another place (
> 
> https://github.com/slimbootloader/slimbootloader/blob/2648eb75f839d9c88607117c2565c04ad0ae510f/BootloaderCommonPkg/Library/TpmLib/TpmLib.c#L1015
> 
> ) that has similar issue. Could you help fix it in this PR?

done

lsubash (commented):
Physical SPI address and blob length has to be captured in the TPM event log which would be per the flash map. Please check the TPM event log on Linux for the firmware blobs.
From the container, CbInfo would contain the Ram address.

PR #2114, opened 2024-01-18
[ADL-LP] GPIO: Add missing GPIO defines

jjdalynh:
Correct 2 GPP_Hx and fill out remaining GPIO definitions for ADL-LP.


ywho79 (approved):
LGTM. Please consolidate all similar changes in a single PR instead of splitting into two

jjdalynh:
will do, thanks.  noticed a fix needed for the first one and then realized how much was actually missing that our new Amston Lake platform will actually need so i just did all of them.

PR #2117, opened 2024-01-19
[MTL Test SKU] Resolve MPInit hang

sindhuragrandhi:
Resolve a hard hang seen during boot when X2APIC is enabled.


cshur (approved):
Great! Sindhura.  2 Weeks ago, I met this issue at OSloader booting on GNR-D after I set ENABLE_SMM_REBASE 0 to 2. I verified it on GNR-D CRB. Works fine. Thanks.

gdong1 (approved):

Issue #2118, opened 2024-01-22
SBL fails to reboot on TigerLake


fridgemagnet3:
SBL fails to reboot TGL (using UP XTREME) after initial programming, instead it sits indefinitely at _FSP Requested Reboot_. A manual power cycle is required then in order to recover, at which point everything is ok.

This used to work and I think this was triggered as a result of [this change](https://github.com/slimbootloader/slimbootloader/pull/1825) which passes the FSP Reset status through _FspResetHandler_ and ultimately through to _ResetSystem_. However in the default _ResetSystemLib_, there is no handler for _EfiResetPlatformSpecific_, hence the board does nothing.

stanleyintel:
Thanks for reporting the issue. 

Correct, it is caused by no handler for EfiResetPlatformSpecific. May you cherry pick https://github.com/slimbootloader/slimbootloader/pull/2120/files and check if the patch can resolve your issue?

stanleyintel:
Hi @fridgemagnet3 

   Can you help to check if the patch can solve the issue? If not, please attach the boot console log.

fridgemagnet3:
Hi @stanleyintel 

Sorry I've been tied up with a few other things, it's on my list of things to do tomorrow. Hopefully I'll report back then.

Thanks. Jon.

stanleyintel:
Thank you. Will move the PR to review.

randylintw:
PR merged. Close it due to a prolonged period of inactivity.

PR #2120, opened 2024-01-23
feat: [TGL] add ResetSystemLib (#2118)

stanleyintel:
The patch installs reset handlers for FSP requested reboot.

The implementation of ResetShutdown and ResetPlatformSpecific are different from the common lib (i.e., BootloaderCommonPkg/Library/ResetSystemLib), so it results in a platform-specific implementation.

Verified with Issue #2118.

### DIFF
--- a/Platform/TigerlakeBoardPkg/BoardConfig.py
+++ b/Platform/TigerlakeBoardPkg/BoardConfig.py
@@ -1,7 +1,7 @@
 ## @file
 # This file is used to provide board specific image information.
 #
-#  Copyright (c) 2018 - 2023, Intel Corporation. All rights reserved.<BR>
+#  Copyright (c) 2018 - 2024, Intel Corporation. All rights reserved.<BR>
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 #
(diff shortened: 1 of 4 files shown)


Platform/TigerlakeBoardPkg/BoardConfig.py
```diff
@@ -277,6 +277,7 @@ def GetPlatformDsc (self):
             'PchP2sbLib|Silicon/$(PCH_PKG_NAME)/Library/PchP2sbLib/PchP2sbLib.inf',
             'PchSbiAccessLib|Silicon/CommonSocPkg/Library/PchSbiAccessLib/PchSbiAccessLib.inf',
             'PlatformHookLib|Silicon/$(SILICON_PKG_NAME)/Library/PlatformHookLib/PlatformHookLib.inf',
+            'ResetSystemLib|Platform/$(BOARD_PKG_NAME)/Library/ResetSystemLib/ResetSystemLib.inf',
```
marcjones-syspro, line 280: I see that every other platform has this code, but why not use the BootloaderCommonPkg? 
 ResetSystemLib|BootloaderCommonPkg/Library/ResetSystemLib/ResetSystemLib.inf

stanleyintel: It is a good suggestion and I will rework the patch:
   1. update the common lib to implement for platform-specific reset.
   2. if a platform has a different implementation requirement for the "platform-specific reset", it will need to implement its own ResetSystemLib

stanleyintel: Summary: No rework will be made for the suggestion, but still thanks for the comment.

Detail:
I just re-examined the differences between common ResetSystemLib and TGL one. Unfortunately, TGL still needs to implement its own ResetSystemLib because of shutdown process.

And it may be not a good idea to add a "EfiResetPlatformSpecific" handler in this commit, because it will mix up 2 different changes in one commit. Let's only focus on TGL change for now. Thanks.

gdong1 (approved):

Platform/TigerlakeBoardPkg/BoardConfig.py
gdong1: It would be great to give info in the commit why the common reset lib could not meet the requirement and hance add this platform specific reset lib.

stanleyintel: > It would be great to give info in the commit why the common reset lib could not meet the requirement and hance add this platform specific reset lib.

Fixed with new commit message.

PR #2125, opened 2024-01-29
-fix: [MTL] Fix S0ix on MTL-PS CRB

kokweich:
1) Added WA in pep.asl to put blocking NVME RP into standby/wakeup 
2) Added PcdMTLPSSupport to support platform specific register value

### DIFF
--- a/Platform/MeteorlakeBoardPkg/AcpiTables/AcpiTables.inf
+++ b/Platform/MeteorlakeBoardPkg/AcpiTables/AcpiTables.inf
@@ -63,6 +63,8 @@
   EcSsdt/EcSsdt.asl
   Ecdt/Ecdt.act
 
+[FixedPcd]
+  gPlatformMeteorLakeTokenSpaceGuid.PcdMtlPSSupport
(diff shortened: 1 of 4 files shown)


Platform/MeteorlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
ywho79, line 1: This change does not seem to be related to the s0ix fix. Suggest to raise a different PR just for this

ywho79 (changes requested):

Platform/MeteorlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
kokweich: Updated to only include S0ix fix, will submit NVS changes in different PR

ywho79 (approved):
LGTM

PR #2126, opened 2024-01-29
fix: [MTL] Sync MTL NVS value with BIOS

kokweich:
Sync NVS value of PCIE devices in Stage2BoardInitLib to match MTLPS BIOS. Also added switch case statement for platform-specific GPIO configuration


ywho79 (approved):
LGTM

PR #2127, opened 2024-01-29
feat:[Common] boot from an image from memory

arghyabi:
Add a new entry in CfgData_BootOption.yaml to boot from memory. The boot device must be 7 (MEMORY), and the address of the container boot image (CONTAINER_BOOT_SIGNATURE) can be placed at an address specified by LBA format (i.e., use HEX prefix. EX: 0x12345678).

Example entry of CfgData_BootOption.yaml:
!expand { BOOT_OPTION_TMPL : [ 0,  0,  0,  7,  '-',  '-',  '-',  '-', '0x479E8000'] }

If the address is not predefined then it can be updated in run time from SBL shell boot option table.

Validated by loading a container image to the memory by using 'fs load' command in SBL shell and update the load image address in boot option table.


PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -1297,6 +1312,41 @@ BootOsImage (
     goto Exit;
   }
 
+  if (MediaType == OsBootDeviceMemory) {
```
gdong1, line 1315: Instead of add a special handling here, did you try to use Memory Device to load image from memory? Memory device could be found here https://github.com/slimbootloader/slimbootloader/blob/master/BootloaderCommonPkg/Library/MemoryDeviceBlockIoLib/MemoryDeviceBlockIoLib.c.
Memory is similar with memory mapped SPI device. It would be more consistent to use MemoryDeviceBlockIoLib.

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -117,6 +118,9 @@ FillBootOptionListFromCfgData (
       // LBA should be defined as 64bit.
       // Will remove the typecast when the structure is fixed.
       BootOption->Image[ImageIdx].LbaImage.LbaAddr = (UINT32)Lba;
+    } else if ((StrPtr[0] == '0') && ((StrPtr[1] == 'x') || (StrPtr[1] == 'X')) &&
+               (AsciiStrHexToUint64S (StrPtr, NULL, &RamAddress) == RETURN_SUCCESS)) {
```
stanleyintel, line 122: It is okay to add a new format "0x<hex>" for BootImage. If so, you may also need to update Platform/CommonBoardPkg/CfgData/Template_BootOption.yaml to tell users that "0x<hex>" is for memory address.

Another idea is to reuse the format: "#0x<HEX>" ? (line 114 to 119 ?) According to my test, the format "#0x<hex>" already works for OsBootDeviceMemory, although its usage was designed for the for LBA  (logic block address) of block device. Maybe the only thing we need to do is to reword the term "LBA"  to a more generic term.

PayloadPkg/OsLoader/OsLoader.c
stanleyintel: Just echo Guo: according to my test, the MemoryDeviceBlockIoLib already works well for OsBootDeviceMemory when calling from GetBootImageFromRawPartition().

However, one thing to note is about the MemoryDeviceGetMediaInfo(), it sets DevBlockInfo->BlockNum  = 0x100000000 without checking the real physical memory size. Maybe we should modify MemoryDeviceGetMediaInfo() to reveal the real physical memory size.

BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h
```diff
@@ -186,6 +186,8 @@ typedef union {
   ///
   ///
   CONTAINER_IMAGE      ContainerImage;
+  /// RAM raw address
+  UINT32               RamAddress;
```
stanleyintel, line 190: It will create a corner error in LoadBootImages (PayloadPkg/OsLoader/LoadImage.c) for the case "BootImage[Index].LbaImage.Valid == 1", because RamAddress and LbaImage are two fields of the union structure. 

For other fields, it is unlikely to have BootImage[Index].LbaImage.Valid == 1, because the first byte of FileName and FileImage will not be 0x1. However, it is possible for the first byte of RamAddress to be 0x1 (although it will not aligned memory).

PayloadPkg/OsLoader/OsLoader.c
arghyabi: Updated

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
arghyabi: Updated

BootloaderCommonPkg/Include/Guid/OsBootOptionGuid.h
arghyabi: Updated

PayloadPkg/OsLoader/OsLoader.c
arghyabi: DevBlockInfo->BlockNum is not used in the boot from memory case.

stanleyintel:
A different perspective on the "boot from memory" concept. If an image is already available in memory and its start address is known, why not boot it directly? This would mean that the GetBootImageFromRawPartition() wouldn't need to call MediaReadBlocks() to read anything again. Consequently, BootloaderCommonPkg/Library/MediaAccessLib/MediaAccessLib.c would be unnecessary for this feature.

stanleyintel (approved):

stanleyintel (approved):

stanleyintel (approved):

stanleyintel (approved):

stanleyintel (approved):

spkamat (approved):

Platform/CommonBoardPkg/Library/BoardSupportLib/BoardSupportLib.c
```diff
@@ -117,6 +118,11 @@ FillBootOptionListFromCfgData (
       // LBA should be defined as 64bit.
       // Will remove the typecast when the structure is fixed.
       BootOption->Image[ImageIdx].LbaImage.LbaAddr = (UINT32)Lba;
+    } else if ((StrPtr[0] == '0') && ((StrPtr[1] == 'x') || (StrPtr[1] == 'X')) &&
+               (AsciiStrHexToUint64S (StrPtr, NULL, &MemoryAddress) == RETURN_SUCCESS)) {
+      // For Boot from Memory case the container image address must be like 0x01234567.
+      // Here the LbaAddress is reused as container image address.
+      BootOption->Image[ImageIdx].LbaImage.LbaAddr = (UINT32)MemoryAddress;
```
gdong1, line 125: A hex value could be used as a file name.  Line 114 asked LBA address to start with "#" to reduce conflict risk. Since we already defined it, could we re-use it instead of adding this new condition?

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -881,7 +887,7 @@ LoadBootImages (
     ContainerImage = &BootImage[Index].ContainerImage;
     if ((ContainerImage->Indicate == '!') && (ContainerImage->BackSlash == '/')) {
       Status = GetBootImageFromIfwiContainer (OsBootOption, LoadedImage);
-    } else if (BootImage[Index].LbaImage.Valid == 1) {
+    } else if (BootImage[Index].LbaImage.Valid == 1 || OsBootOption->DevType == OsBootDeviceMemory) {
```
gdong1, line 890: We don't need take memory device as a special device. Valid bit should be set if booting using Lba.

arghyabi: Updated

PR #2129, opened 2024-02-01
fix: [MTL] fix coverity issues

kokweich:
Size of PepSerialIoI2c array is 6 and NHLT table is 1 
Hence, removing cases where its possible to access out of bound index


ywho79 (approved):
LGTM

PR #2130, opened 2024-02-01
OsLoader: Mark multiboot module extra image as ran after appended

jiaqingz-intel:
A multiboot module extra image is appended to boot image and never runs. So mark it as ran after appended. This elimates the "Extra image not loaded, or need pass it to OS in boot parameter" warning.

### DIFF
--- a/PayloadPkg/OsLoader/OsLoader.c
+++ b/PayloadPkg/OsLoader/OsLoader.c
@@ -1076,6 +1076,7 @@ SetupBootImages (
           (LoadedExtraImage->Flags & LOADED_IMAGE_MBMODULE)) {
         DEBUG ((DEBUG_INFO, "SetupBootImage Append ImageType-%d\n", Type));
         AppendMultibootModules (LoadedImage, LoadedExtraImage);
+        LoadedExtraImage->Flags |= LOADED_IMAGE_RUN_EXTRA;
       }
     }
   }


PayloadPkg/OsLoader/OsLoader.c
```diff
@@ -1076,6 +1076,7 @@ SetupBootImages (
           (LoadedExtraImage->Flags & LOADED_IMAGE_MBMODULE)) {
         DEBUG ((DEBUG_INFO, "SetupBootImage Append ImageType-%d\n", Type));
         AppendMultibootModules (LoadedImage, LoadedExtraImage);
+        LoadedExtraImage->Flags |= LOADED_IMAGE_RUN_EXTRA;
```
gdong1, line 1079: Given it is a valid case (extra image loaded, and don't need pass the loaded image to OS), prefer removing the warning message instead adding this flag.

jiaqingz-intel: This log is for the extra image not in MBMODULE format. Other format images seems will be run after loaded. I'd like to keep the current behavior.

gdong1: LOADED_IMAGE_RUN_EXTRA imply there is an extra image will be loaded and run before booting normal OS.
If you don't want to run extra image before normal OS, please remove "BOOT_FLAGS_EXTRA" from boot option to have a try.

jiaqingz-intel: For the MBMODULE type of extra image, instead of running it before the final image, it is appended to the "main" image when it is loaded, and run with the main image together. In this case, if it is appended, it should be marked as ran.

ywho79:
@jiaqingz-intel , is this PR still required? Else we should close this

PR #2132, opened 2024-02-01
fix: [Stage1B] Extending correct cfg data and hash length in tpm

aborwank:
length of external config data and hash keys extended to the TPM was incorrect. This commit fixes this issue.


cans-arista:
Looks good to me! Tested on an ICX-D system

Thanks,
Can

gdong1 (approved):

lsubash (approved):

PR #2133, opened 2024-02-02
feat: [MTLPS] Enable PCH binding region for secure flash

lsubash:
Secure flash uses PCH binding region for storing signature. Used Region 13 for secure flash.


jinjhuli (approved):

ywho79 (approved):
LGTM

PR #2140, opened 2024-02-15
fix: Remove check for physical root bridge device before scanning.

bejeanmo:
Not all root bridges have a physical device at dev 0 func 0 of the root bus, and this requirement prevents some platforms from being fully scanned. From this point, PciEnumerationLib doesn't access the root bridge configuration space at all, so this is an unnecessary check.

Instead, root bus scanning will continue without dev 0 func 0 present. If no devices are found on the root bus, scanning will continue to the next root bridge.

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
@@ -1664,7 +1664,6 @@ PciScanRootBridges (
   OUT       UINT8                  *RootBridgeCount
   )
 {
-  UINT32                            Address;
   UINT16                            Bus;
   UINT8                             SubBusNumber;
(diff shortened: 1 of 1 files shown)


BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c
```diff
@@ -1730,27 +1729,24 @@ PciScanRootBridges (
       Bus = Index;
     }
 
-    Address = PCI_EXPRESS_LIB_ADDRESS (Bus, 0, 0, 0);
-    if (PciExpressRead16 (Address) != 0xFFFF) {
```
bejeanmo, line 1734: This diff looks complicated, but really it's only removal of these two lines, the closing brace, and the Address declaration. The rest is just indent removal.

sindhuragrandhi (approved):

PR #2141, opened 2024-02-16
feat: copying actm binary to build folder

aborwank:
For platforms with seperate actm binary added code to copy actm binary in build folder


lsubash (commented):
Which component will load ACTM binary?
Can this be added as container IPFW if this can be loaded FSP?

aborwank:
> Which component will load ACTM binary? Can this be added as container IPFW if this can be loaded FSP?

fsp will load ACTM. ACTM binary address has to be given to fsp. It has to be 256K aligned. On adding it in IPFW container some header bits are getting added and hence the alignment becomes wrong.

lsubash:
A

> > Which component will load ACTM binary? Can this be added as container IPFW if this can be loaded FSP?
> 
> fsp will load ACTM. ACTM binary address has to be given to fsp. It has to be 256K aligned. On adding it in IPFW container some header bits are getting added and hence the alignment becomes wrong.

Would you dump first 256 bytes and check with actual binary? With container, ACTM will be a container component and respective component has to be extracted.

cshur (approved):
LKGM

PR #2146, opened 2024-03-04
Support bootloader a/b slot for Firmwareupdate Payload

ceiba1985:
Firmwareupadte Payload needs support bootloader a/b slots. The SBL update package is put to bootloader partition and will be updated during the OTA.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
+++ b/PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
@@ -21,7 +21,35 @@
 #include <Library/FirmwareUpdateLib.h>
 #include <Library/ConfigDataLib.h>
 #include <Library/BootOptionLib.h>
+#include <Library/AbSupportLib.h>
 #include <ConfigDataCommonStruct.h>
+
+EFI_STATUS
+GetSwPartByName (
+  IN  EFI_HANDLE            PartHandle,
+  IN  CHAR16*               PartName,
+  OUT UINT32*               SwPart
+  )
+{
+  PART_BLOCK_DEVICE   *PartBlockDev;
+  UINT32 Index;
+
+  // Validate parameters
+  PartBlockDev = (PART_BLOCK_DEVICE *)PartHandle;
+  if ( (PartBlockDev == NULL) || \
+       (PartBlockDev->Signature != PART_INFO_SIGNATURE) ) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  for (Index = 0; Index < PartBlockDev->BlockDeviceCount; Index++) {
+      if (StrCmp(PartName, PartBlockDev->BlockDevice[Index].PartitionName) == 0) {
+        DEBUG ((DEBUG_INFO, "Found Partition %s at SwPart %d\n", PartName, Index));
+        *SwPart = Index;
+        return EFI_SUCCESS;
+      }
+    }
+  return EFI_NOT_FOUND;
+}
 
 /**
   Initialize Boot Device (Media)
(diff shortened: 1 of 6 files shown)


PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -269,12 +300,25 @@ LoadCapsuleImage (
   FileHandle        = NULL;
   HwPartHandle      = NULL;
   FsHandle          = NULL;
+  SwPart		    = 0;
+  BootSlot = 0;
```
bejeanmo, line 304: Align with previous variable initializations, and use spaces instead of tab characters.

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -269,12 +300,25 @@ LoadCapsuleImage (
...
 
   Status = FindBootPartition (CapsuleInfo, &HwPartHandle);
   if (EFI_ERROR (Status)) {
     goto Done;
   }
 
+  Status = GetSwPartByName (HwPartHandle, L"misc", &SwPart);
+  if (EFI_ERROR (Status)) {
+	DEBUG ((DEBUG_INFO, "Unable to find misc partition, Status = %r\n", Status));
```
bejeanmo, line 313: Use spaces instead of tabs.

bejeanmo (changes requested):

gdong1 (approved):
It looks good for me in general. Not sure the impact to change the default values for config data. I could merge it if there is no concern from others.

marcjones-syspro:
It is hard to understand why this patch is needed based on the description. Is there any documentation on A/B support? I couldn't find it. 

It seems that this patch is needed if A/B support is used and the firmware update image is written to the active "slot" and not to the specific SW partition number as indicated in CAPSULE_INFO_CFG_DATA. https://slimbootloader.github.io/security/firmware-update.html#capsule-location

ceiba1985:
Suppose capsule is configured to be put in bootloader_a partition in CAPSULE_INFO_CFG_DATA.  And for a device, if it run in the A slot currently,  the old capsule is in bootloader_a partition.  But after OTA, the device will run in B slot, and the updated capsule is in bootloader_b partition. Without this patch, sbl will still read capsule from bootloader_a partition, whitch is an old one.

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -269,12 +300,25 @@ LoadCapsuleImage (
...
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_INFO, "Unable to find misc partition, Status = %r\n", Status));
+  }
+
+  BootOption.BootFlags = BOOT_FLAGS_MISC | BOOT_FLAGS_MENDER;
+  BootOption.Image[LoadImageTypeMisc].LbaImage.SwPart = (UINT8)SwPart;
+  BootOption.Image[LoadImageTypeMisc].LbaImage.LbaAddr= 0;
+  BootOption.HwPart = CapsuleInfo->HwPart;
+  BootSlot = GetBootSlot (&BootOption, HwPartHandle);
+
```
gdong1, line 321: SBL capsule update image location is decided by the CapsuleInfo.
From the patch, it looks the expected SBL capsule image should be read from PART_A when mender OS is boot from PART_A, and read from PART_B when mender OS is boot from PART_B.
This is a new feature request, and the changes should not impact existing usage. (e.g. should not try to read boot slot info in normal capsule image loading).
I prefer adding a flag in CAPSULE_INFO_CFG_DATA to support this new feature. With this new flag, the change in this file looks like as below:
BootSlot = 0;
if (CapsuleInfo->CapsuleFlag & CAPSULE_FLAGS_MENDER !=0) {
  // update BootSlot as above
}

gdong1:
close this PR as it is not needed.

PR #2150, opened 2024-03-08
[ADL] Make MRC data cache-able to improve performence

liyi77:
Boot time of FSP-M will reduce 90~100ms.

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
@@ -161,8 +161,9 @@ BoardInit (
     PlatformHookSerialPortInitialize ();
     SerialPortInitialize ();
 
-    // Enlarge the code cache region to cover full flash for non-BootGuard case only
-    if ((AsmReadMsr64(MSR_BOOT_GUARD_SACM_INFO) & B_BOOT_GUARD_SACM_INFO_NEM_ENABLED) == 0) {
+    // Enlarge the code cache region to cover full flash for non-BootGuard case or fast boot case
+    if ((AsmReadMsr64(MSR_BOOT_GUARD_SACM_INFO) & B_BOOT_GUARD_SACM_INFO_NEM_ENABLED) == 0
+        || PcdGetBool (PcdFastBootEnabled)) {
       // FSP-T does not allow to enable full flash code cache due to cache size restriction.
       // Here, MTRR is patched to enable full flash region cache to avoid performance penalty.
       // However, the SBL code flow should ensure only limited flash regions will be accessed
--- a/Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.inf
+++ b/Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.inf
@@ -59,6 +59,7 @@
   gPlatformModuleTokenSpaceGuid.PcdRedundantRegionSize
   gPlatformModuleTokenSpaceGuid.PcdStage1BSize
   gPlatformCommonLibTokenSpaceGuid.PcdDebugPortNumber
+  gPlatformModuleTokenSpaceGuid.PcdFastBootEnabled
 
 [FixedPcd]
   gPlatformAlderLakeTokenSpaceGuid.PcdAdlLpSupport


Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
```diff
@@ -119,6 +121,86 @@ EarlyPlatformDataCheck (
...
+      }
+    }
+  }
+  // Append MTRR to make sure MRCD cache-able, which can improve performance of FSP-M.
+  // Note: Mtrr base address MUST align to size (And size must be 2 to the Nth power and larger than 4KB),
+  // unfortunately SBL didnt promise this. So it must be mapped with multiple MTRRs.
+  MemoryAddress = (EFI_PHYSICAL_ADDRESS) ALIGN_DOWN (PcdGet32 (PcdMrcDataBase), MemoryLength);
+  while (MemoryAddress < (PcdGet32 (PcdMrcDataBase) + PcdGet32 (PcdMrcDataSize))) {
+    if (UsedMtrr >= VariableMtrrCount) {
+      break;
```
bejeanmo, line 191: There is a case here where it requires multiple MTRRs to cover the MRCD region, and the available MTRRs is at least one but insufficient to satisfy it. In this case, the code will exit with MTRRs only partially covering the MRC Data region. This may be acceptable, but I wanted to make you aware of it.

liyi77: Yes, I believe this is acceptable.
I tested this case, even if only part of the MRCD is covered, it will improve the performance of FSPM - the effect depends on the proportion of coverage, and there will be no impact on functionality.

jwu55: Yi,  i agree with Bejean,  We should avoid the case only map part of MRCD.  There's risk even result looks "positive" in your local.

liyi77: I think a possible risk is that my local test cannot cover all strange MRCD base and length pairs. Ok, I will update patch to check  valid MTRR number.

jwu55 (approved):

liyi77:
Hi @bejeanmo , previous comments have been fixed, would you like review this patch again?

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
```diff
@@ -119,6 +121,95 @@ EarlyPlatformDataCheck (
...
 }
 
+/**
+  CAR created by ACM covers only Initial code region.
+  This function will expand CAR to cover MRC data and improve performance of MRC code.
+
+**/
+VOID
+EFIAPI
+ExpandCacheAsRam (
```
gdong1, line 131: From the function logic, it just sets MRC data in flash to be catchable.  I didn't see it has any relation with CacheAsRam (NEM). So prefer not using it.

lsubash:
@liyi77 Is this PR is validated with BTG5 profile enabled?

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
```diff
@@ -181,6 +272,10 @@ BoardInit (
       }
       AsmWriteMsr64(MsrIdx, (SIZE_4GB - AdjLen) | CACHE_WRITEPROTECTED);
       AsmWriteMsr64(MsrIdx + 1, (MskLen - AdjLen) | B_CACHE_MTRR_VALID);
+    } else {
+      if (PcdGetBool (PcdFastBootEnabled)) {
+        ExpandCacheAsRam ();
```
gdong1, line 277: This new function is quite complicated and it might take several MTRRs for MRC data and I didn't see it is freed.
Since other BIOS region also need be set to cacheable, could you have a try to set a free MTRR MSR to cover whole BIOS region (similar with BtG disabled case)? Hopefully it could also solve the boot performance issue, and use less MSR and benefit other bios component.

liyi77:
> @liyi77 Is this PR is validated with BTG5 profile enabled?

Yes.

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
liyi77: Thanks for review, maybe ExpandTempRam ()? Then it can also be used if there are other features that need to set var MTRR.

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
liyi77: 1. TempRamExit will free all used var MTRRs after PreMem stage. Usually there are 10 var MTRRs, and 2-3 will be used when ACM initializes NEM (based on the size of IBB). I didn’t see var MTRR used elsewhere in Stage1A.
2. This is just to make the code more general and avoid not having enough LLC to cover the entire BIOS region on some SKU. I don't have enough resources to test all SKUs so I don't know if they have enough LLC.

@jwu55 Hi Ji, do you see any risk about 1 single MTRR?

gdong1: ok. Then it is not an issue to use multiple MTRRs.  
I know there are some limitations on the NEM size (e.g. NEM size = stack data region + code data region; Effective cache size = NEM size + reserved reserved region; both stack data region and code region should be continuous region, etc.).  I don't know the exact reasons for these limitations. I think it should work since a) it works in BtG disabled  case, b) MRC data region is a new code region which is not a continuous region with IBB, it already breaks a limitation.

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
liyi77: Sounds good, patch updated.

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
liyi77: I think you are right. Since it works fine with BTG disabled, it should work the same with BtG enabled. My worries about Cache size are unnecessary.
I updated patch please check.
MTRR value of BtG enabled case as below:
1. Before change:
MTRR: ID(0) BASE(FEF00006) Mask(FFF00800) 
MTRR: ID(1) BASE(FFE00005) Mask(FFF00800) (Stage1A + ACM0)
MTRR: ID(2) BASE(FFFE0005) Mask(FFFE0800) (Stage1B)

2. After
MTRR: ID(0) BASE(FEF00006) Mask(FFF00800) 
MTRR: ID(1) BASE(FF000005) Mask(FF000800) (Full flash region same as BtG disabled case)

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
```diff
@@ -119,6 +121,73 @@ EarlyPlatformDataCheck (
   }
 }
 
+/**
+  CAR created by ACM covers only Initial code region.
+  This function will expand CAR to cover full flash to improve performance.
```
gdong1, line 126: Maybe update this function description. This function enables code cache for BIOS region in flash (PcdFlashSize is BIOS region size). CAR  (Cache As Ram) means a part of cache is used as RAM which is readable and writable. This function is not to set or expand a CAR region (It is already setup by FSPT).

Platform/AlderlakeBoardPkg/Library/Stage1ABoardInitLib/Stage1ABoardInitLib.c
gdong1: with this new change, did you see any boot performance impact comparing with your previous change (only add MRC data region)? 
BTW, to make code simple, without clearing other MTRRs should be OK. The hardware should be able to support the overlap.

liyi77: > with this new change, did you see any boot performance impact comparing with your previous change (only add MRC data region)?
No, performance is same as previous change (FSPM reduced 100ms and no other impact).

> BTW, to make code simple, without clearing other MTRRs should be OK. The hardware should be able to support the overlap.
Sounds good, then we just need to add fastboot check and no need to add new function. I tested and it works fine, patch updated.

gdong1 (approved):

PR #2163, opened 2024-04-10
Fit support

gdong1:


BootloaderCommonPkg/Include/Library/FitLib.h
```diff
@@ -0,0 +1,62 @@
+/** @file
+  FIT Load Image Support
+Copyright (c) 2023, Intel Corporation. All rights reserved.<BR>
```
sindhuragrandhi, line 3: should we change the header to 2024?

gdong1: This file comes from https://github.com/tianocore/edk2/blob/master/UefiPayloadPkg/PayloadLoaderPeim/FitLib.h with minor update. Changing copy right year is optional.

PR #2164, opened 2024-04-11
[MTL]: Update frame buffer HOB

gdong1:
Frame buffer address might be changed during PCI enumeration. So need make sure frame buffer info are correct in the HOB.


sindhuragrandhi (approved):

ywho79 (approved):

gdong1:
close this pr as MTL framebuffer is not same with the BAR1 address.

PR #2168, opened 2024-04-15
fix: [MTL] CRB hang in UEFI payload when PCIe switch present.

randylintw:
Porting from https://github.com/slimbootloader/slimbootloader/pull/1892

Required for https://github.com/slimbootloader/slimbootloader/pull/2164 to avoid exception on uefipayload.


bejeanmo (approved):

randylintw:
close it since #2164 isn't plan to merge.

PR #2171, opened 2024-04-22
fix: FSP 2.4 issues

bejeanmo:
1) Some FSP 2.4 implementations are non-conforming and use FSPM_ARCH_UPD instead of FSPM_ARCH2_UPD as indicated by the specification. Logic is changed to check FSPM UPD header revision for structure version instead of FSP spec revision.

2) MultiPhase FSP PhaseIndex is one-based so loop comparisons need to take this into account. Side effect is that the last phase may be missed. No current platform FSP is utilizing this.


bejeanmo:
Waiting to merge this until some platforms with anomalous FSP can implement workarounds.

gdong1 (approved):

PR #2177, opened 2024-05-10
[EHL] trigger SMI handler in S3 resume

stanleyintel:
This commit resolves the issue where the SMI handler was not being triggered during S3 resume. The problem was due to the functions RestoreS3RegInfo and TriggerPayloadSwSmi not being called.

In addition, the commit also:
       1. unset the PME_B0_EN as UEFI Payload does not have its handler
       2. remove the ClearSmbuStatus() because
          - the HSTS.B_SMBUS_IO_SMBALERT_STS in SMBUS (B0:D31:F4) should be
            handled and cleared by device driver or an appropriate SMI handler.
          - the ClearSmi() will clear GPE0_STS_127_96.SMB_WAK_STS if it is set
       3. unset (FSPS) PeiGraphicsPeimInit and GraphicsConfigPtr during S3 resume
       4. narrow the var scope of mSmmBaseInfo and mS3SaveReg
       5. add required bitfield declaration

Verified with:
       1. UEFI Payload + Ubuntu on EHL CRB (release build)
       2. OSLoader (release build)
      when FEATURES_CFG_DATA.Features.S0ix = 0

### DIFF
--- a/Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
+++ b/Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
@@ -105,11 +105,11 @@ typedef struct {
   UINT32    Rsvd2:3;
 } GPIO_CFG_DATA_DW1;
 
-SMMBASE_INFO mSmmBaseInfo = {
+STATIC SMMBASE_INFO mSmmBaseInfo = {
   { BL_PLD_COMM_SIG, SMMBASE_INFO_COMM_ID, 0, 0 }
 };
 
-S3_SAVE_REG mS3SaveReg = {
+STATIC S3_SAVE_REG mS3SaveReg = {
   { BL_PLD_COMM_SIG, S3_SAVE_REG_COMM_ID, 1, 0 },
   { { REG_TYPE_IO, WIDE32, { 0, 0}, (ACPI_BASE_ADDRESS + R_ACPI_IO_SMI_EN), 0x00000000 } }
 };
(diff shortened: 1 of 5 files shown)


Platform/ElkhartlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -780,7 +798,23 @@ BoardInit (
         DEBUG ((DEBUG_WARN, "SCI device has boot issue\n"));
       }
     }
-    ClearSmbusStatus ();
```
gdong1, line 783: why remove this one?  It isn't mentioned in the commit message.

stanleyintel: The ClearSmbuStatus was designed to clear GPE0_STS_127_96.SMB_WAK_STS and HSTS.B_SMBUS_IO_SMBALERT_STS for SMBUS (B0:D31:F4). However, 
  1. the HSTS.B_SMBUS_IO_SMBALERT_STS should be detected, handled, and cleared by device driver (or proper SMI handler)
  2. the ClearSmi() will clear GPE0_STS_127_96.SMB_WAK_STS if it is set
As a result, ClearSmbuStatus() is an unwanted implementation.

A new commit is submitted to
  1. remove ClearSmbuStatus() code
  2. enhance commit message

gdong1: If I remember correctly, ClearSmBusStatus was added because there are 2 instances to use Smbus to get data in SBL. The second instance doesn't work because Smbus STS is set.  Maybe we should check if this could be handled in SmbusLib if it happens again.

gdong1 (approved):

PR #2186, opened 2024-05-31
feat: Improvements in Stage remapping on 64-bit builds.

bejeanmo:
Recent RemapStage() changes enabled remapping Stage1B when ENABLE_FAST_BOOT is set, even if STAGE1B_XIP is set. This introduced limitations with 64-bit builds because in 64-bit mode, Page Directory Entries (PDEs) cover 2 MiB instead of 4 MiB in 32-bit mode. This will cause Stage1B remapping to fail if Stage1B can't be contained by one 2 MiB aligned PDE. Fortunately there is enough space in the existing page table to use two sets of PDEs/PTEs in this case, doubling the remappable size of Stage1B.

Additionally, RemapStage() should not fail when MapMemoryRange() fails but STAGE1B_XIP is set, since executing Stage1B in place is the expected behavior. Allowing boot to continue without remapping Stage1B will cause a performance loss instead of a failure.

### DIFF
--- a/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
+++ b/BootloaderCommonPkg/Library/PagingLib/PagingMap.c
@@ -85,22 +85,33 @@ MapMemoryRange (
       // 64 bit mode
       PteOffset = 6 * EFI_PAGE_SIZE;
       PdeOffset = 2 * EFI_PAGE_SIZE;
-      EntryNum  = 512;
     } else {
       // 32 bit mode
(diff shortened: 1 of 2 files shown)


bejeanmo:
This PR gives a slight improvement in remappable size, but remapping will still have problems at bigger sizes. Closing this for now since this has been addressed by separating remapping from XIP and Fastboot dependencies. This can be revisited later if there is a need.

PR #2189, opened 2024-06-03
feat: Add BoardConfig field to control Stage1B remap

bejeanmo:
Adding new Board Config item REMAP_STAGE1B to control remapping Stage1B into permanent memory after FSP-M. Decouple this from ENABLE_FAST_BOOT and STAGE1B_XIP. This makes it simpler to enable this remap when the slight performance boost may be needed, and leave it disabled by default.

### DIFF
--- a/BootloaderCorePkg/Library/StageLib/StageLib.c
+++ b/BootloaderCorePkg/Library/StageLib/StageLib.c
@@ -104,12 +104,12 @@ RemapStage (
   UINTN                Cr0;
 
   Is64Bit = IS_X64;
-  if (!Is64Bit && FeaturePcdGet (PcdStage1BXip) && !PcdGetBool (PcdFastBootEnabled)) {
+  if (!Is64Bit && !FeaturePcdGet (PcdRemapStage1B)) {
     return;
   }
 
   PageTblSize = GetPageTablesMemorySize (Is64Bit);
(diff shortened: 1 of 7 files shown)


BootloaderCorePkg/Library/StageLib/StageLib.c
```diff
@@ -121,8 +121,8 @@ RemapStage (
   DEBUG ((DEBUG_INFO, "Load page table from memory @ 0x%08X\n", (UINT32)(UINTN)PageBuffer));
   Status = Create4GbPageTables (PageBuffer, Is64Bit);
 
-  if (!EFI_ERROR(Status) && (!FeaturePcdGet (PcdStage1BXip) || PcdGetBool (PcdFastBootEnabled))) {
-    if (PcdGetBool (PcdFastBootEnabled)) {
+  if (!EFI_ERROR(Status) && FeaturePcdGet (PcdRemapStage1B)) {
+    if (FeaturePcdGet (PcdRemapStage1B)) {
```
gdong1, line 125: Line 125 is not required since it is checked in line 124.

gdong1:
To keep behavior same, it looks we need add "self.REMAP_STAGE1B        = 1" in qemu board config file.

gdong1 (approved):

PR #2208, opened 2024-06-26
feat: [ADL] Separate Fsp version by project

randylintw:
When the fsp.git commit id is updated,
it will update all platform fsp versions without testing. 
Also adjust the fd size in adln50 project to avoid build break.

### DIFF
--- a/Platform/AlderlakeBoardPkg/BoardConfigAdlP.py
+++ b/Platform/AlderlakeBoardPkg/BoardConfigAdlP.py
@@ -27,6 +27,7 @@ class Board(AlderlakeBoardConfig.Board):
         self._EXTRA_INC_PATH      = ['Silicon/AlderlakePkg/Adlp/Include']
         self._FSP_PATH_NAME       = 'Silicon/AlderlakePkg/Adlp/FspBin'
         self.MICROCODE_INF_FILE   = 'Silicon/AlderlakePkg/Microcode/Microcode.inf'
+        self.FSP_INF_FILE         = 'Silicon/AlderlakePkg/FspBin/FspBinIpu.inf'
         self.ACPI_TABLE_INF_FILE  = 'Platform/AlderlakeBoardPkg/AcpiTables/AcpiTablesP.inf'
         self.ENABLE_TCC           = 0
         self._generated_cfg_file_prefix = ''
--- a/Platform/AlderlakeBoardPkg/BoardConfigAzb.py
+++ b/Platform/AlderlakeBoardPkg/BoardConfigAzb.py
@@ -38,6 +38,7 @@ class Board(BaseBoard):
         self._EXTRA_INC_PATH                  = ['Silicon/AlderlakePkg/Azb/Include']
         self._FSP_PATH_NAME                   = 'Silicon/AlderlakePkg/Azb/FspBin'
         self.MICROCODE_INF_FILE               = 'Silicon/AlderlakePkg/Microcode/MicrocodeAzb.inf'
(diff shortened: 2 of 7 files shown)


Platform/AlderlakeBoardPkg/BoardConfigAdlP.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, *args, **kwargs):
         self._EXTRA_INC_PATH      = ['Silicon/AlderlakePkg/Adlp/Include']
         self._FSP_PATH_NAME       = 'Silicon/AlderlakePkg/Adlp/FspBin'
         self.MICROCODE_INF_FILE   = 'Silicon/AlderlakePkg/Microcode/Microcode.inf'
+        self.FSP_INF_FILE         = 'Silicon/AlderlakePkg/FspBin/FspBinIPU.inf'
```
gdong1, line 30: what "IPU" means? do we need change to FspBinIpu.inf to align with the naming conversion?

Platform/AlderlakeBoardPkg/BoardConfigAzb.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, *args, **kwargs):
         self._EXTRA_INC_PATH                  = ['Silicon/AlderlakePkg/Azb/Include']
         self._FSP_PATH_NAME                   = 'Silicon/AlderlakePkg/Azb/FspBin'
         self.MICROCODE_INF_FILE               = 'Silicon/AlderlakePkg/Microcode/MicrocodeAzb.inf'
+        self.FSP_INF_FILE                     = 'Silicon/AlderlakePkg/FspBin/FspBinAzb.inf'
```
gdong1, line 41: The name "Azb" doesn't match with the one in actual file name "azb". maybe need update the actual one to avoid build failure.

Platform/AlderlakeBoardPkg/BoardConfigAdlN.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, *args, **kwargs):
         self.FSP_IMAGE_ID         = 'ADLI-FSP'
         self._EXTRA_INC_PATH      = ['Silicon/AlderlakePkg/Adln/Include']
         self._FSP_PATH_NAME       = 'Silicon/AlderlakePkg/Adln/FspBin'
+        self.FSP_INF_FILE         = 'Silicon/AlderlakePkg/FspBin/FspBin.inf'
```
gdong1, line 37: This is the default path. it is optional to add it.

Platform/AlderlakeBoardPkg/BoardConfigAdlP.py
randylintw: Intel platform update ~

Platform/AlderlakeBoardPkg/BoardConfigAzb.py
randylintw: Strange, my local file name file name case is correct, will move to Linux to check.

Platform/AlderlakeBoardPkg/BoardConfigAdlN.py
randylintw: ok, removed.

gdong1 (approved):

Platform/AlderlakeBoardPkg/BoardConfigAdlP.py
gdong1: Just wondering how it is related with ADL-P.

PR #2214, opened 2024-07-02
[UPX i12] Enable UPX i12 RPL-P

tsaikevin:
Add support for Up Xtreme i12 RPL-P based board.
The PCIe M.2 slot CN11 on the board is able to detect NVMe SSD. Debug output is enabled at header CN9 on the board(e.g. UART1) Tested to boot with OS loader payload and UEFI payload.

To stitch the SlimBootloader.bin with IFWI uses StitchLoader.py script with '-p' as given below:

python Platform/AlderlakeBoardPkg/Script/StitchLoader.py -i <BIOS_IMAGE_NAME> -s Outputs/rplp/SlimBootloader.bin -o sbl_upx12rp_ifwi.bin -p 0xAA000114


ywho79 (approved):
LGTM

PR #2224, opened 2024-07-09
fix: Reload segment selectors after GDT Load in Stage1a

bejeanmo:
If ResetVector GDT is different from that loaded in Stage1A SecStartup() then segment selectors need to be reloaded after GDT is updated.

### DIFF
--- a/BootloaderCorePkg/Stage1A/Ia32/UpdateSelectors.nasm
+++ b/BootloaderCorePkg/Stage1A/Ia32/UpdateSelectors.nasm
@@ -1,0 +1,38 @@
+;------------------------------------------------------------------------------
+;
+; Copyright (c) 2024, Intel Corporation. All rights reserved.<BR>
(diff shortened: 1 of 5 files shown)


BootloaderCorePkg/Stage1A/Ia32/UpdateSelectors.nasm
```diff
@@ -0,0 +1,38 @@
...
+; Abstract:
+;
+;  This code reloads the segment selectors after GDT update
+;
+;------------------------------------------------------------------------------
+DEFAULT REL
+SECTION .text
+BITS 32
+
+; LINEAR_CODE_SEL 0x10 for x64 CS
```
jasonlouyun, line 20: hi, "0x10" should be for x32 CS, right?

bejeanmo: Correct. I will fix the typo.

gdong1 (approved):

PR #2226, opened 2024-07-10
fix : [ARL] Added deprecation warning for ia32 Build

aborwank:
Deprecation warning added in the build log for
'arls' and 'arl' ia32 build.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -1329,6 +1329,9 @@ class Build(object):
         rebuild_basetools ()
 
     def build(self):
+        if (self._board.BOARD_NAME in [ 'arl', 'arls' ]) & (self._board.BUILD_ARCH == 'IA32') :
+            WARNING = '\033[93m'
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -1329,6 +1329,9 @@ def early_build_init(self):
         rebuild_basetools ()
 
     def build(self):
+        if (self._board.BOARD_NAME in [ 'arl', 'arls' ]) & (self._board.BUILD_ARCH == 'IA32') :
+            WARNING = '\033[93m'
+            print(f"{WARNING}\n\n!!! DEPRECATED !!! Arrowlake ia32 build is obsolete. !!! DEPRECATED !!!\n\n{WARNING}")
```
gdong1, line 1334: It is not expected to add platform specific check in common script.
IA32 is supported for ARL even by default it build using X64. I think it is not necessary to add this kind of warning.

aborwank:
@lsubash Can you please review.

PR #2244, opened 2024-08-01
fix:[ARLS] moving Bootgaurd lib to commonsoc lib

aborwank:
Moved ARL-S code to common soc library.


Silicon/CommonSocPkg/Library/BootGuardLibCBnT/BootGuardLibrary.c
```diff
@@ -85,8 +90,18 @@ GetBootGuardInfo (
     /// Read the ME FWS Register
     ///
     if (HeciBaseAddress != 0) {
+      MeFwSts = MmioRead32 (HeciBaseAddress + R_ME_HFS);
```
lsubash, line 93: Are the new registers getting used in status check  apart from MeFwSts4?

aborwank: MeFwSts4 is only used. In ARLS boot guard lib we are printing all MeFwSts. It is not required.

aborwank: I will remove the code and remove the ARLS btg lib.

PR #2249, opened 2024-08-08
feat: [ARLH] Enable UPDs to follow SiConfigData

ssbyna:
PchLockDownBiosLock & PchLockDownGlobalSmi follow board SiCfgData

### DIFF
--- a/Platform/ArrowlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
+++ b/Platform/ArrowlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
@@ -592,16 +592,6 @@ UpdateFspConfig (
     FspsConfig->PchLockDownGlobalSmi = 0x1;
   }
 
-  if (MtlIsSocM ()) {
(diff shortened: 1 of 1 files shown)


Platform/ArrowlakeBoardPkg/Library/FspsUpdUpdateLib/FspsUpdUpdateLib.c
```diff
@@ -631,6 +621,18 @@ UpdateFspConfig (
   FspsConfig->AmtEnabled = 0x1;
   FspsConfig->EnableTcoTimer = 0x1;
 
+  if (MtlIsSocM ()) {
```
ywho79, line 624: Is ARL and MTL sharing the same function? Would be good if this function is named ArlIsSocM instead but if can't then it's fine

ssbyna: currently the same tagging MtlIsSocM is used for both MTL & ARL in bios source code, so we didnt change it yet to avoid any mismatch

ywho79 (approved):

PR #2250, opened 2024-08-09
fix: GenContainer.py not setting default auth type correctly

bejeanmo:
When no auth parameter is specified on the command line, gen_layout chooses auth type based on provided key, but this does not get correctly applied to the monolithic signing component because of an order of operations issue.

### DIFF
--- a/BootloaderCorePkg/Tools/GenContainer.py
+++ b/BootloaderCorePkg/Tools/GenContainer.py
@@ -683,12 +683,14 @@ def adjust_auth_type (auth_type_str, key_path):
     return auth_type_str
 
 def gen_layout (comp_list, img_type, auth_type_str, svn, out_file, key_dir, key_file):
-    hash_type = CONTAINER._auth_to_hashalg_str[auth_type_str] if auth_type_str else ''
+    if auth_type_str in ['SHA2_256','SHA2_384']:
+        raise Exception ("Invalid 'auth' parameter '%s'. Monolithic signing requires a signature-type auth." % (auth_type_str))
     auth_type = auth_type_str
     key_path  = os.path.join(key_dir, key_file)
     auth_type = adjust_auth_type (auth_type, key_path)
-    if auth_type == '':
-        raise Exception ("'auth' parameter is expected  !")
+    hash_type = CONTAINER._auth_to_hashalg_str[auth_type]
+    if auth_type_str == '':
+        print ("No 'auth' parameter specified. Defaulting to '%s' based on key type." % (auth_type))
 
     # prepare the layout from individual components from '-cl'
     if img_type not in CONTAINER_HDR._image_type.keys():


BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -683,12 +683,17 @@ def adjust_auth_type (auth_type_str, key_path):
     return auth_type_str
 
 def gen_layout (comp_list, img_type, auth_type_str, svn, out_file, key_dir, key_file):
+    if auth_type_str in ['SHA2_256','SHA2_384']:
+        raise Exception ("Invalid 'auth' parameter '%s'. Monolithic signing requires a signature-type auth." % (auth_type_str))
```
lsubash, line 687: I think we should allow only hash auth also,
This is common function for any type of container

bejeanmo: This wasn't working previously for monolithic signing, I just added the more clear explanation.

BootloaderCorePkg/Tools/GenContainer.py
```diff
@@ -683,12 +683,17 @@ def adjust_auth_type (auth_type_str, key_path):
...
     auth_type = auth_type_str
     key_path  = os.path.join(key_dir, key_file)
     auth_type = adjust_auth_type (auth_type, key_path)
-    if auth_type == '':
-        raise Exception ("'auth' parameter is expected  !")
+    # If auth type was adjusted and not empty, update hash type accordingly
+    if (auth_type != '') and (auth_type != auth_type_str):
+        hash_type = CONTAINER._auth_to_hashalg_str[auth_type]
+    if auth_type_str == '':
+        print ("No 'auth' parameter specified. Defaulting to '%s' based on key type." % (auth_type))
```
gdong1, line 696: Would it be simple to replace line 688 ~ 694 as below:
    key_path  = os.path.join(key_dir, key_file)
    auth_type = adjust_auth_type (auth_type_str, key_path)
    hash_type = CONTAINER._auth_to_hashalg_str[auth_type]

bejeanmo: This is essentially what I had initially in the PR but I wanted to make it clear that the cases of auth type = SHA2_256/SHA2_384 are not valid for monolithic signing.

gdong1: I didn't check the impact for the newly added lines 686~687. My suggested change doesn't include it.

bejeanmo: I get your meaning now. I believe I was covering a corner case that isn't actually valid. I will try your simplification.

gdong1 (approved):

lsubash (approved):

PR #2275, opened 2024-08-26
feat: [common] Enable PCIe resizable BAR support

pastorcx:
This commit adds support for PCIe resizable BARsupport. 
The feature can be enabled by setting PcdResizableBarSupport for the board build script and its disabled by default.

### DIFF
--- a/BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
+++ b/BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
@@ -61,6 +61,11 @@ typedef enum {
   BusScanTypeInvalid  = 0xFF
 } BUS_SCAN_TYPE;
 
+typedef enum {
+  PciResizableBarMin = 0x00,
+  PciResizableBarMax = 0xFF
+} PCI_RESIZABLE_BAR_OPERATION;
+
 typedef struct {
   UINT16            Io32            : 1;
   UINT16            Mem64           : 1;
@@ -152,6 +157,7 @@ struct _PCI_IO_DEVICE {
   BOOLEAN                                   IsPciExp;
   UINT8                                     PciExpressCapabilityOffset;
   UINT32                                    AriCapabilityOffset;
+
   //
   // SR-IOV
   //
@@ -160,6 +166,15 @@ struct _PCI_IO_DEVICE {
   UINT32                                    SystemPageSize;
   UINT16                                    InitialVFs;
   UINT16                                    ReservedBusNum;
+
+  //
+  // Per PCI to PCI Bridge spec, I/O window is 4K aligned,
+  // but some chipsets support non-standard I/O window alignments less than 4K.
+  // This field is used to support this case.
+  //
+  UINT16                                    BridgeIoAlignment;
+  UINT32                                    ResizableBarOffset;
+  UINT32                                    ResizableBarNumber;
 
   //
   // The bridge device this pci device is subject to
@@ -192,6 +207,17 @@ BarExisted (
   IN  UINTN         Offset,
   OUT UINT32        *BarLengthValue,
   OUT UINT32        *OriginalBarValue
+  );
+
+/**
+  Initialize Resizable BAR
+
+  @param[in,out]  PciIoDevice     Pointer to instance of PCI_IO_DEVICE.
+
+**/
+VOID
+InitializeResizeBar (
+  IN OUT  PCI_IO_DEVICE *PciIoDevice
   );
 
 #endif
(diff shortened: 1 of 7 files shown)


bejeanmo (changes requested):
Resizable BARs will also affect BAR size calculation in PciParseBar() will it not? The PCIE spec says that if BAR size is changed through this mechanism that BAR resource allocation should use this extended capability to check BAR size instead of the old method of writing all 0xF's to the BAR.

bejeanmo:
> Resizable BARs will also affect BAR size calculation in PciParseBar() will it not? The PCIE spec says that if BAR size is changed through this mechanism that BAR resource allocation should use this extended capability to check BAR size instead of the old method of writing all 0xF's to the BAR.

I discussed my concern with the code author, and we believe the PCIE note to which I referred is discussing a different edge case that is not applicable here. Since resizable BAR programming happens prior to resource allocation in this implementation, the size should be correctly reflected in the BAR read-only bits.

BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
```diff
@@ -160,6 +165,14 @@ struct _PCI_IO_DEVICE {
...
   UINT16                                    InitialVFs;
   UINT16                                    ReservedBusNum;
+  //
+  // Per PCI to PCI Bridge spec, I/O window is 4K aligned,
+  // but some chipsets support non-standard I/O window alignments less than 4K.
+  // This field is used to support this case.
+  //
+  UINT16                                       BridgeIoAlignment;
+  UINT32                                       ResizableBarOffset;
+  UINT32                                       ResizableBarNumber;
```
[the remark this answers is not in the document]
bejeanmo: what I mean is that the fields in line 168 and up, as well as line 182 and down start at column 45, while the fields in the newly added lines start at column 48. They should all start at 45.

bejeanmo (changes requested):

BootloaderCorePkg/Library/PciEnumerationLib/InternalPciEnumerationLib.h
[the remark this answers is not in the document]
pastorcx: done, fields adjusted to start at column 45, thanks for the info

BootloaderCorePkg/Library/PciEnumerationLib/PciResizableBar.h
```diff
@@ -0,0 +1,20 @@
...
+#define __PCI_RESIZABLE_BAR_H__
+
+VOID
+EFIAPI
+InitializeResizeBar (
+  IN OUT  PCI_IO_DEVICE              *PciIoDevice,
+  IN      UINT8                       Bus,
+  IN      UINT8                       Device,
+  IN      UINT8                       Func
+  );
```
gdong1, line 18: It looks this is just an internal API of PciEnumerationLib, so better put it into InternalPciEnumerationLib.h instead of create a new header file.

BootloaderCorePkg/Library/PciEnumerationLib/PciResizableBar.c
```diff
@@ -0,0 +1,128 @@
...
+  @param[in]      Func            Device func NO.
+
+**/
+VOID
+EFIAPI
+InitializeResizeBar (
+  IN OUT  PCI_IO_DEVICE              *PciIoDevice,
+  IN      UINT8                       Bus,
+  IN      UINT8                       Device,
+  IN      UINT8                       Func
```
gdong1, line 106: Bus, device and Func are not used in this internal function, better remove them.

BootloaderCorePkg/Library/PciEnumerationLib/PciResizableBar.h
pastorcx: done, moved into the InternalPciEnumerationLib.h and deleted the PciResizableBar header file

BootloaderCorePkg/Library/PciEnumerationLib/PciResizableBar.c
pastorcx: It is used in the [PciEnumerationLib.c](https://github.com/slimbootloader/slimbootloader/blob/0a9d7749e384370162c70b22368ac1d6661c32db/BootloaderCorePkg/Library/PciEnumerationLib/PciEnumerationLib.c#L446) when enable "SUPPORT_RESIZABLE_BAR" through the board build script. 
Removing the bus,device,func generates conflict with previous declaration of "InitializeResizeBar" from InternalPciEnumerationLib.h

gdong1: This is newly added internal function, not sure why add BDF since they are not used.  If remove BDF, of course remove it from header file and caller at same time in this PR.
BTW, this function is not library API function, it is not necessary to add EFIAPI.

pastorcx: ok, removed the BDF from function

gdong1 (approved):

bejeanmo (approved):

PR #2280, opened 2024-08-30
feat: [Common] TCC tools related code clean up

vincentchchen:
Removed the source code related to TCC tools
including DSO switch, SWSRAM switch, TCC subregions TCC Error Log switch, RTCM, RTCT
The code change applies on TGL/EHL/ADL/RPL/MTL


randylintw:
Thanks @vincentchchen , 
BTW, suggest to separate the on-going release project and non release project to avoid TCs fail problem.

PR #2269, opened 2024-08-23
feat:[ADL-P] Add i2c library

costel-ignat:
Ported I2c driver for UEFI.

PR #2282, opened 2024-09-02
feat: [RPL-A] Add support for automotive RVP board.

costel-ignat:
- Added new CPU Id and Plaform ID (MBL RVP).
- Add PreMem GPIO table.
- Add DDI configuration table.
- Add new SPD entry to accommodate the new memory configuration.
- Add support for PCM3168A Audio Codec.
- Implement board detection over I2c.
- Update Irq number for Device D21F3 in the ACPI tables to fix the booting issue in FSP debug mode.
- Configure GPIO pins to be used as display HPD in the ACPI DSDT table.


costel-ignat:
Also added i2c library as a build dependency , it will be removed after the PR #2269  will be merged.

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
```diff
@@ -266,33 +269,54 @@ UpdateFspConfig (
...
+    Fspmcfg->MemorySpdPtr031  = SpdData[MemCfgData->SpdDataSel031];
+    Fspmcfg->MemorySpdPtr100  = SpdData[MemCfgData->SpdDataSel100];
+    Fspmcfg->MemorySpdPtr101  = SpdData[MemCfgData->SpdDataSel101];
+    Fspmcfg->MemorySpdPtr110  = SpdData[MemCfgData->SpdDataSel110];
+    Fspmcfg->MemorySpdPtr111  = SpdData[MemCfgData->SpdDataSel111];
+    Fspmcfg->MemorySpdPtr120  = SpdData[MemCfgData->SpdDataSel120];
+    Fspmcfg->MemorySpdPtr121  = SpdData[MemCfgData->SpdDataSel121];
+    Fspmcfg->MemorySpdPtr130  = SpdData[MemCfgData->SpdDataSel130];
+    Fspmcfg->MemorySpdPtr131  = SpdData[MemCfgData->SpdDataSel131];
+  }
```
gdong1, line 319: It is preferred to use SBL config data for SPD data. It doesn't need change code when changing the memory, and it doesn't impact other platforms.
For your use case, often it needs different PlatformId for different memory configuration.  Given the only difference is the memory, I think it is ok to use a single PlatformId and update MemCfgData->SpdDataSel dynamically using MemId. This way you just need minimal code change here.

stefange-amr: @gdong1 there are two things here that are not matching:

1. The order of the SpdData arrays in SpdTableRplpLp5Auto is different then the way Sbl loads the SpdData.bin files (other order, other memories supported) so even though we update the MemCfgData->SpdDataSelXXX with MemId this would not have the same effect. 
2. Our automotive configuration could have same PlatformId but different memory configurations e.g. 16GB/32GB/64Gb - and this is configured by Manufacturing in the FRU chip.

gdong1: I didn't get your point on the matching or order issue.
It looks you have 9 SPD data blobs for this auto platform, please check if this flow works.
1. In save the SPD data into Files
     e.g.: save mRplpLp5AutoMicron4GMT62F1G32D2DSSpd into Spd_Lp5_Auto_Micron_4G.bin
2. Update auto platform dlt file, override MEMORY_CFG_DATA.MemorySpdPtrx as below
    MEMORY_CFG_DATA.MemorySpdPtr0  | {FILE: Spd_Lp5_Auto_Micron_4G.bin}
    MEMORY_CFG_DATA.MemorySpdPtr1  | {FILE: Spd_Lp5_Auto_Micron_8G.bin}
    MEMORY_CFG_DATA.MemorySpdPtr2  | {FILE: Spd_Lp5_Auto_Micron_16G.bin}
    ...
    MEMORY_CFG_DATA.MemorySpdPtr8  | {FILE: Spd_Lp5_Auto_Samsung_16G.bin}

  # override the default value
  MEMORY_CFG_DATA.SpdDataSel101            | 0

3. Just need add the logic as below in this file 
 
  #if defined(PLATFORM_RPLA)
   MemId = PcdGet8 (PcdMemConfigId);
    Fspmcfg->MemorySpdPtr000  = MemId;
    Fspmcfg->MemorySpdPtr010  = MemId;
    Fspmcfg->MemorySpdPtr020  = MemId;
    Fspmcfg->MemorySpdPtr030  = MemId;
    Fspmcfg->MemorySpdPtr100  = MemId;
    Fspmcfg->MemorySpdPtr110  = MemId;
    Fspmcfg->MemorySpdPtr120  = MemId;
    Fspmcfg->MemorySpdPtr130  = MemId;
#endif

Please specify the exact place if have issue with above update.

stefange-amr: Hi @gdong1 ! thanks a lot for the suggestion! However, wouldn't the proposed solution break the current implementation, considering the fact that we will overwrite the SpdPtr? e.g.

MEMORY_CFG_DATA.MemorySpdPtr0 | {FILE: Spd_Lp5_Auto_Micron_4G.bin}
MEMORY_CFG_DATA.MemorySpdPtr1 | {FILE: Spd_Lp5_Auto_Micron_8G.bin}
MEMORY_CFG_DATA.MemorySpdPtr2 | {FILE: Spd_Lp5_Auto_Micron_16G.bin}
...
MEMORY_CFG_DATA.MemorySpdPtr8 | {FILE: Spd_Lp5_Auto_Samsung_16G.bin}

This will break, at least in theory, the existing support for the already existing memory configurations that Rplp supports. However, since we don't need any of those memory configurations, and since we are maintaining our own SblInternal/Platform/RplAuto/CfgData/*** i guess i wouldn't be a problem to remove all the other memories and simply add support for our own.

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
```diff
@@ -266,33 +269,54 @@ UpdateFspConfig (
...
+    SpdData[0] = 0;
+    SpdData[1] = (UINT32)(UINTN) (((MEM_SPD0_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD0_TAG))->MemorySpdPtr0);
+    SpdData[2] = (UINT32)(UINTN) (((MEM_SPD1_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD1_TAG))->MemorySpdPtr1);
+    SpdData[3] = (UINT32)(UINTN) (((MEM_SPD2_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD2_TAG))->MemorySpdPtr2);
+    SpdData[4] = (UINT32)(UINTN) (((MEM_SPD3_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD3_TAG))->MemorySpdPtr3);
+    SpdData[5] = (UINT32)(UINTN) (((MEM_SPD4_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD4_TAG))->MemorySpdPtr4);
+    SpdData[6] = (UINT32)(UINTN) (((MEM_SPD5_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD5_TAG))->MemorySpdPtr5);
+    SpdData[7] = (UINT32)(UINTN) (((MEM_SPD6_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD6_TAG))->MemorySpdPtr6);
+    SpdData[8] = (UINT32)(UINTN) (((MEM_SPD7_CFG_DATA *)FindConfigDataByTag (CDATA_MEM_SPD7_TAG))->MemorySpdPtr7);
+#if defined(PLATFORM_RPLA)
```
bachaue1, line 300: I think this #ifdef PLATFORM_RPLA ... #endif is wrong because the baseline code sets SpdData[9] unconditionally (cf. original line 279).

Platform/AlderlakeBoardPkg/Include/PlatformBoardId.h
```diff
@@ -75,4 +75,16 @@ Defines Platform BoardIds
 
 #define PLATFORM_ID_RPL_P_DDR5_CRB                    0x08
 
+// RPL-P Auto Malibou Lake Boards
+#define BoardIdRplPLp5AutoRvp                         0x05  // Standard 6 bits Board ID used when reading from IO Expander
+#define ExtendedBoardIdRplPLp5AutoRvp                 0x45  // MBL RVP Platform ID when read from FRU EEPROM
+#define BoardIdRplPAutoLp5Crb                         0x46  // MBL CRB Platform ID when read from FRU EEPROM
+
+#define PLATFORM_ID_RPLP_LP5_AUTO_RVP                 0x05  // Internal Platform Id for Malibou Lake RVP Fab. A and B.
+#define PLATFORM_ID_RPLP_LP5_AUTO_CRB                 0x06  // Internal Platform Id for Malibou Lake CRB.
```
bachaue1, line 84: This conflicts with PLATFORM_ID_ADL_N_UP7EN50 (cf. line 22, above).
Any chance to get a unique value? (0x1..3, 0xc, 0xe..f, 0x15..16, 0x1a and 0x1d..1f seem to be available).

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
gdong1: All the changes in DLT file will only apply to the specific platform, there is no impact to other platforms.  If you find it doesn't work, it should be a bug.

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
costel-ignat: Done.

Platform/AlderlakeBoardPkg/Include/PlatformBoardId.h
costel-ignat: Done.

Platform/AlderlakeBoardPkg/Library/FspmUpdUpdateLib/FspmUpdUpdateLib.c
costel-ignat: Done.
Thanks for the suggestion.

gdong1 (approved):

PR #2283, opened 2024-09-04
fix: Source Level Debug breaks in X64.

bejeanmo:
DebugAgentLib was using its own IDT Entry structure definition that didn't adjust for x64. Removed this definition and switched to built in definition from ProcessorBind.h which does adjust for architecture correctly.


randylintw:
Hi Bejean, will it related to this one?
https://groups.io/g/slimbootloader/message/3744

bejeanmo:
> Hi Bejean, will it related to this one? https://groups.io/g/slimbootloader/message/3744

I saw this on another platform, but yes this looks like the same behavior.

ywho79 (approved):

PR #2301, opened 2024-10-08
ARLP: Fix ARL-P/H/U boot issue

gdong1:
After enabled GPIOV2, TopSwap library used new P2sbLib. For ARL-P, it looks P2SB SBI method doesn't work for P2sb registers, so changed to use MMIO method.
With this change, it could boot to shell using legacy UART (0xFF) or SOC UART (0).


bejeanmo:
Any idea why SBI doesn't work here?

gdong1:
> Any idea why SBI doesn't work here?

Because P2SB_PID was defined but it was not used. so for ARL-P it used a wrong PID.

ywho79 (approved):

PR #2305, opened 2024-10-15
fix: osloader can't boot on new yocto build

randylintw:
Ext2Fs->Ext2FsFeaturesIncompat report EXT2F_INCOMPAT_CSUM_SEED so that fail into not support.

boot tested with old/new yocto image, ubuntu.


randylintw:
fixed patch checker error for tab usage.

gdong1 (approved):

PR #2316, opened 2024-10-23
fix:[MeChipsetLib] Fix the issue of HECI read timeout

aborwank:
Setting MeHeciTimeoutsEnabled to TRUE so that Heci packet read timeout is enabled.


lsubash (approved):
Check the other usecases using these API.

gdong1 (approved):

PR #2319, opened 2024-10-30
fix: [MTL-P] EC timeout issues/ACPI error on RVP.

randylintw:
Windows system log will report Event 13, ACPI error several times and cause windows slow booting issue.

This patch contains:
1. EC data can't be read/updated.
2. fixed ACPI error in Linux system
3. Include NVS def and sync NVS definition. Without this, the EC ACPI device is disappear in windows.


gdong1 (approved):
looks good to me.

Issue #2324, opened 2024-11-06
AlderLake and RaptorLake Platforms define EnableItbm in two locations


marcjones-syspro:
On AlderLake and RaptorLake Platforms,  `EnableItbm` is defined in[ CfgData_Power.yaml](https://github.com/slimbootloader/slimbootloader/blob/b86a51ce98c09772d44cc398cb4cdab0442f7e88/Platform/AlderlakeBoardPkg/CfgData/CfgData_Power.yaml#L674) and [CfgData_Silicon.yaml](https://github.com/slimbootloader/slimbootloader/blob/b86a51ce98c09772d44cc398cb4cdab0442f7e88/Platform/AlderlakeBoardPkg/CfgData/CfgData_Silicon.yaml#L47) and they have opposite defaults. One is enabled and the other isn't. 

CfgData_Power .yaml:
```yaml
  - EnableItbm   :
      name         : Intel Turbo Boost Max Technology 3.0
      type         : Combo
      option       : $EN_DIS
      help         : >
                     Intel Turbo Boost Max Technology 3.0. 0- Disabled; <b>1- Enabled</b>
      length       : 0x01
      value        : 0x01
```

cfgData_Silicon.yaml:
```yaml
  - EnableItbm   :
      condition    : $(COND_S0IX_DIS)
      name         : Intel Turbo Boost Max Technology 3.0
      type         : Combo
      option       : $EN_DIS
      help         : >
                     Intel Turbo Boost Max Technology 3.0. 0- Disabled; <b>1- Enabled</b>
      length       : 0x01
      value        : 0x00
```

A number of .DLT files set [SILICON_CFG_DATA.EnableItbm | 0x1](https://github.com/slimbootloader/slimbootloader/blob/b86a51ce98c09772d44cc398cb4cdab0442f7e88/Platform/AlderlakeBoardPkg/CfgData/CfgDataInt_Adln_Crb_Ddr5.dlt#L122) so you would assume that setting it to 0 would work,
[...]

PR #2325, opened 2024-11-07
fix: [ADL, RPL, MTL] definition for EnableItbm #2324

stanleyintel:
On ADL, RPL and MTL, the definition of EnableItbm had been moved from SILICON_CFG_DATA to POWER_CFG_DATA. The commit removes the unwanted definition in CfgData_Silicon and correct it in DLT files.


marcjones-syspro (approved):
LGTM

PR #2326, opened 2024-11-11
fix:[AZB] Fix USB drive detection issue on CRB

arghyabi:
Resolved hardware issue where USB instance 0, mapped to eMMC, needs initialization and deinitialization to detect a USB flash drive.

### DIFF
--- a/Platform/AlderlakeBoardPkg/CfgData/CfgData_BootOption_Azb.yaml
+++ b/Platform/AlderlakeBoardPkg/CfgData/CfgData_BootOption_Azb.yaml
@@ -11,10 +11,9 @@
 - $ACTION      :
     page         : OS
 # usb
-- !expand { BOOT_OPTION_TMPL : [ 0 ,   0       ,  0 ,    5   ,   0   ,    0xFF  ,     1 ,     1 , '/boot/sbl_os' ] }
-- !expand { BOOT_OPTION_TMPL : [ 1 ,   0       ,  0 ,    5   ,   1   ,    0xFF  ,     1 ,     1 , '/boot/sbl_os' ] }
+- !expand { BOOT_OPTION_TMPL : [ 0 ,   0       ,  0 ,    5   ,   1   ,    0xFF  ,     1 ,     1 , '/boot/sbl_os' ] }
 # nvme
-- !expand { BOOT_OPTION_TMPL : [ 2 ,   0       ,  0 ,    6   ,   0   ,    0x0   ,     1 ,     1 , '/boot/sbl_os' ] }
-- !expand { BOOT_OPTION_TMPL : [ 3 ,   0       ,  0 ,    6   ,   1   ,    0x0   ,     1 ,     1 , '/boot/sbl_os' ] }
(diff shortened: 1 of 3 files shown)


spkamat (approved):

Platform/AlderlakeBoardPkg/CfgData/CfgData_BootOption_Azb.yaml
```diff
@@ -11,10 +11,9 @@
 - $ACTION      :
     page         : OS
 # usb
-- !expand { BOOT_OPTION_TMPL : [ 0 ,   0       ,  0 ,    5   ,   0   ,    0xFF  ,     1 ,     1 , '/boot/sbl_os' ] }
-- !expand { BOOT_OPTION_TMPL : [ 1 ,   0       ,  0 ,    5   ,   1   ,    0xFF  ,     1 ,     1 , '/boot/sbl_os' ] }
+- !expand { BOOT_OPTION_TMPL : [ 0 ,   0       ,  0 ,    5   ,   1   ,    0xFF  ,     1 ,     1 , '/boot/sbl_os' ] }
```
marcjones-syspro, line 14: While less common, I think that this will miss USB2 flash drives and the USB eMMC as a boot device.

marcjones-syspro (commented):
I don't completely understand the issue, but I also noticed the the USB3 keys wouldn't get an insert notification/present detect until after init/deinit of the USB2 controller. Are there some details on the issue? Is this AZB silicon specific or something about the mainboard?

Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -571,6 +624,12 @@ BoardInit (
 
     //Initialize and setup MeMeasureboot
     MeMeasuredBootInit();
+#if FixedPcdGetBool(PcdAzbSupport)
+    // Workaround for specific hardware design of CRB platform:
+    // USB instance 0 is mapped to eMMC, so it must be initialized
+    // and deinitialized to enable USB flash drive detection.
+    UsbDevInstanceInitDeinit();
```
gdong1, line 631: If USB boot option is not in the boot option list or not the first boot option, it is not necessary to apply this option. It would impact boot performance. 
Not sure if we could narrow down which settings (registers) are necessary to WA the HW issue instead of calling USB init and de-init.

Xx3cd1 (approved):

Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
gdong1: I am ok if this is just a temp fix to unblock users.

gdong1 (approved):

Platform/AlderlakeBoardPkg/CfgData/CfgData_BootOption_Azb.yaml
gdong1: From the author comment, it is to WA a board issue. This is just a temp fix before a formal PR.

PR #2340, opened 2024-12-02
fix: [Common/GpioLib] Don't let invalid pin stop GPIO initialization

bachaue1:
If the init table passed to GPIO initialization contains an invalid pin number, debug builds of the init library would terminate silently (i.e. no entries after the offending one would be processed at all, and common config registers like HOSTSW_OWN and unlock data might not get updated). Note that release builds would not detect (and thus simply skip) invalid entries.

This change:
- aligns the debug to the release build behavior: handle all valid GPIO pins passed to the initialization routine;
- complain about invalid pins (debug builds only).
Signed-off-by: Bruno Achauer <bruno.achauer@intel.com>


gdong1 (approved):

gdong1:
Previously there is CI build error which was fixed by https://github.com/slimbootloader/slimbootloader/pull/2343. Not sure why the CI still uses old configuration for this PR. Please rebase to master to push again.

bachaue1:
> Previously there is CI build error which was fixed by #2343. Not sure why the CI still uses old configuration for this PR. Please rebase to master to push again.

Done.

PR #2345, opened 2024-12-03
CI: test the CI build

gdong1:
don't merge this PR


gdong1:
only for test purpose. it is not needed

PR #2357, opened 2024-12-30
feat: [MTL] Update FSP for MR2

samihahkasim:
BIOS version is NEX MTL-UH_MTL-PS MR2 (4122_56) FSP
FSP version is 0DE0CC30


randylintw (approved):
LGTM

PR #2366, opened 2025-01-17
fix: [TGL] increased EPAYLOAD size for UefiPld

vincentchchen:
EPAYLOAD_SIZE increased to accommodate UEFI payload built with different toolchain


randylintw (approved):
LGTM.

PR #2367, opened 2025-01-22
feat: [ARL] Upstream ARL Config Data files

kokweich:
Upstream ARL config data files to open source


Platform/ArrowlakeBoardPkg/CfgData/CfgData_BootOption.yaml
```diff
@@ -0,0 +1,20 @@
...
+#  Copyright (c) 2020 - 2023, Intel Corporation. All rights reserved.<BR>
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+
+- $ACTION      :
+    page         : OS
+- !expand { BOOT_OPTION_TMPL : [ 0 ,   0       ,  0 ,    5   ,   0   ,    0xFF  ,     1 ,     1 , '/boot/sbl_os' ] }
+- !expand { BOOT_OPTION_TMPL : [ 1 ,   0x1F    ,  0 ,    5   ,   0   ,    0     ,     1 ,     1 , '/boot/sbl_rtcm' ] }
```
gdong1, line 14: I think RTCM is no longer available anymore, and SBL would drop its support. please check if it should be removed from the boot option.

kokweich: Thanks Guo Dong , have updated the code to remove RTCM support from boot option, please help to review again

PR #2369, opened 2025-01-22
feat: [ARL] Upstream ARL board config files

kokweich:
Upstream board config files for ARL-H, ARL-S and ARL-U to open source


gdong1:
Please check if we could share same config file for U/H.

randylintw:
pls also enable ci build for arls/arlh, thx.

kokweich:
> Please check if we could share same config file for U/H.

There is some ACM differences hence needed a different config file for U, however I have modified the boardConfig to reuse similar configs from ARLH

PR #2387, opened 2025-02-10
fix: remove clear FSP hob from Uefi payload boot path

tsaikevin:
Remove clear FSP hob from Uefi payload boot path because Fsp NotifyPhase needs FSP hob to work. Fixed 32bit address FspHotList in FSP INFO. UEFI payload consumes ths info to clear FSP hob.


Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -672,6 +670,7 @@ BoardInit (
   case ReadyToBoot:
     if ((GetBootMode() != BOOT_ON_FLASH_UPDATE) && (GetPayloadId() == 0)) {
       ProgramSecuritySetting ();
+      ClearFspHob ();
     }
```
gdong1, line 674: Better moving to EndOfFirmware callback since there is last call to FSP here.

tsaikevin: @gdong1 updated. fyi

gdong1 (approved):

PR #2390, opened 2025-02-12
fix:[QEMU] fix MicroPython authentication issue

arghyabi:
Fix the MicroPython container authentication failure issue due to security violation.

Ref: https://github.com/slimbootloader/slimbootloader/commit/CFF9A54C183F77B8


bejeanmo:
CFGD will still cause a security violation after setup payload changes are saved. Instead, the SBL SETUP payload feature requires verified boot to be disabled. I.e. it is not designed for production firmware, only as a development convenience feature.

PR #2401, opened 2025-02-24
fix: Add error message for firmware update on unsupported redundant regions

pastorcx:
Currently, Firmware Update is not allowed for the following redundant regions on their own: SG02, PYLD, FWUP, EMTY! 
Add an error message when performing a firmware update for unsupported regions.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
@@ -1183,7 +1183,16 @@ UpdateSblComponent (
   //
   if (IsRedundantComponent(ImageHdr->UpdateHardwareInstance)) {
     DEBUG ((DEBUG_INFO, "Redundant component update requested! \n"));
-    Status = UpdateSystemFirmware(ImageHdr, FwPolicy);
+    if (((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_PAYLOAD ||
+        ((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_FWUPDATE ||
+        ((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_STAGE2 ||
+        ((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_EMPTY) {
+        DEBUG((DEBUG_ERROR, "FirmwareUpdate is not allowed for the following Redundant regions on their own: SG02, PYLD, FWUP, EMTY!\n"));
+        Status = EFI_UNSUPPORTED;
+        return Status;
+    } else {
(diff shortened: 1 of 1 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
```diff
@@ -1183,7 +1183,16 @@ UpdateSblComponent (
   //
   if (IsRedundantComponent(ImageHdr->UpdateHardwareInstance)) {
     DEBUG ((DEBUG_INFO, "Redundant component update requested! \n"));
-    Status = UpdateSystemFirmware(ImageHdr, FwPolicy);
+    if (((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_PAYLOAD ||
+        ((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_FWUPDATE ||
+        ((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_STAGE2 ||
+        ((UINT32) ImageHdr->UpdateHardwareInstance) == FLASH_MAP_SIG_EMPTY) {
+        DEBUG((DEBUG_ERROR, "FirmwareUpdate is not allowed for the following Redundant regions on their own: SG02, PYLD, FWUP, EMTY!\n"));
+        Status = EFI_UNSUPPORTED;
```
stanleyintel, line 1191: The IsRedundantComponent() function already performs the feasibility check, so additional checks here may be redundant. 

Additionally, hardcoding component signatures as a blacklist might not be ideal, as customers could design a different flash map. This assumption could potentially cause issues with their customized flash layout.

cristian-mihail: @stanleyintel I agree this is not the optimal solution and we could find a better way to define the component signature blacklist based on the flashmap defined in the board config script, but the issue persists. Currently, the firmware update process does not states clearly what is allowed and what not. This causes confusion for us developers, let alone users. 

Currently the error shown when trying to update only the STAGE2 region, shows as follows
```
[13:42:17.958648]:  VerifyFwVersion failed with Status = 0x80000002
[13:42:17.958668]: ApplyFwImage (32304753:0000) failed with Status = Invalid Parameter
```

without any other context. This error could be improved. 

This change should be paired with a change in GenContainer.py where we don't allow certain components to be included into the fw update container.

gdong1: Agree with @stanleyintel, this PR is not necessary. To update an individual component, only container image is support. So to update Stage2 component (SG02 ), the first error message should be "Registering container failed due to Unsupported".  and from the info "ApplyFwImage: 32304753" in the log, it is clear updating SG02 is not supported.
So I prefer just close this PR.

pastorcx: I'm closing the PR, thanks for the clarification.

PR #2413, opened 2025-03-16
QEMU FSP BuildTools fix in newer toolchains

fernandosilvame:
When building using newer toolchains, there are a few changes required that were already patched in EDK2.

Those changes are documented in the two links from EDK2 group discussion:
- https://edk2.groups.io/g/devel/topic/89997410
- https://edk2.groups.io/g/devel/topic/89997412

This set of small changes will later be helpful when moving to newer OSes and toolchains. There is no downside in applying them now as the build in current supported OSes is working fine. It should be a no harm change with future benefit.

Hopefully we can merge this delta to enable folks in the community to enable work in OSes other than Ubuntu in future. Fedora 41 for example needs this change and some additional work in the toolchain setup.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,3 +17,4 @@
 # since the files that are being patched have a mixture of line
 # endings. This means it must be treated as binary.
 Silicon/QemuSocPkg/FspBin/Patches/0001-Build-QEMU-FSP-2.0-binaries.patch binary
+Silicon/QemuSocPkg/FspBin/Patches/0002-BaseTools-GCC-newer-versions-fix-Synced-with-EDK2.patch binary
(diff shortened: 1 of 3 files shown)


gdong1:
Thanks for this enhancement, please fix the patch check failure.
you could refer https://dev.azure.com/slimbootloader/slimbootloader/_build/results?buildId=4384&view=logs&j=f459684c-ef12-5bb6-86c7-b5bdd3607efb&t=4a675116-3ae1-51fa-85a2-31deea81e723&l=17 for details.

fernandosilvame:
Thank you, Guo Dong!
I will need to spend some more time as vscode pylint doesn't find the trailing whitespaces the build is complaining about and I cannot spot it when whitespace rendering is enabled.

fernandosilvame:
Ok, now I understand. The previous commit is holding the queue. I will close this PR and resubmit with clean one.
Thank you!

PR #2418, opened 2025-03-21
refactor: [Common] Add common pch p2sb library

ckolhe:
Added pch p2sb library to common package.
In subsequent PRs, platforms which require this library will be configured to use this common implementation by removing their local platform specific implementation.

### DIFF
--- a/Silicon/CommonSocPkg/Include/Library/PchP2sbLib.h
+++ b/Silicon/CommonSocPkg/Include/Library/PchP2sbLib.h
@@ -1,0 +1,154 @@
+/** @file
+  Header file for PchP2sbLib.
+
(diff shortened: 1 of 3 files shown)


gdong1:
SBL already has a common P2sbLib which is preferred for all the platforms.
could you help check if you could update the platform code to use existing P2sbLib instead of PchP2sbLib? Potentially PchP2sbLib could be removed if no one uses it.

ckolhe:
> SBL already has a common P2sbLib which is preferred for all the platforms. could you help check if you could update the platform code to use existing P2sbLib instead of PchP2sbLib? Potentially PchP2sbLib could be removed if no one uses it.

Yes. Will create separate PR to remove unused instances of pch p2sb library. Deleting this PR.

PR #2428, opened 2025-04-01
[ARL-H/S/U] FIPS negative test case

aborwank:
This commit adds a negative test case for
FIPS validation.


BootloaderCommonPkg/Library/IppCrypto2Lib/fips.c
```diff
@@ -376,15 +380,19 @@ RunFipsSelftests ()
             return RETURN_UNSUPPORTED;
         }
     }
-/* SM3 is not supported by FIPS
+/* SM3 is not supported by FIPS.
+ * Here It is used for negative testcase execution.
+ */
+#if FixedPcdGetBool(PcdFipsNegativeTestcase)
```
gdong1, line 386: Instead of adding a new PCD, why don't we use existing PCD PcdIppHashLibSupportedMask? e.g.:
#if  FixedPcdGet8(PcdIppHashLibSupportedMask) & IPP_HASHLIB_SM3_256

aborwank: done

Platform/ArrowlakeBoardPkg/BoardConfigArlh.py
```diff
@@ -106,6 +110,9 @@ def __init__(self, *args, **kwargs):
         if self.ENABLE_IPP_CRYPTO_PERF:
             self.IPP_HASH_LIB_SUPPORTED_MASK |= IPP_CRYPTO_ALG_MASK['SM3_256']
 
+        if self.FIPS_SELFTEST_NEGATIVE_TEST:
+            self.IPP_HASH_LIB_SUPPORTED_MASK |= IPP_CRYPTO_ALG_MASK['SM3_256']
```
gdong1, line 114: SM3_256 was set when self.ENABLE_IPP_CRYPTO_PERF is set (refer 2 lines above).  Do you still need add another MACRO to control it?

lsubash (commented):
Antara, Can we simply the security configs? We can set the top level config's and use sub-function to set the remaining ones?

Platform/ArrowlakeBoardPkg/BoardConfigArlh.py
gdong1: Just like other algorithms, we don't need add dedicate config item for SM3_256. When SM3_256 is configured to use, then test it, else don't test it.

aborwank: This PR is for testing the usecase where FIPS sefltest would fail. Since FIPS does not support SM3, fips test for SM3 always fails. To ensure the failure case scenario SM3 need to be enabled specifically.

gdong1: Currently FIPS self test will test the algorithms actually configured/used in SBL.
From current setting,    self.IPP_HASH_LIB_SUPPORTED_MASK   = IPP_CRYPTO_ALG_MASK['SHA2_384'] | IPP_CRYPTO_ALG_MASK['SHA2_256'] in line 106. If SBL is configured to use SM3, the FIPS self test will fail (it is expected). So I think we don't need add a new configuration FIPS_SELFTEST_NEGATIVE_TEST since FIPS always tests all the Algorithms actually used.

PR #2431, opened 2025-04-08
[MTL] The CPU package power is unable to meet the specified level

randylintw:
When set the POWER_CFG_DATA.TurboMode/EnableItbm/Eist = 1, The 155HL CPU package power run on 35w only and can't reach PL1[45W] and PL2[115W].

Sync the ACPI table with this BIOS change:
  Change _CPC() to fix HWP malfunction in E-core

PM Verified on MTL-PS CRB/WIN11/PTAT with TurboMode = 1.


Silicon/MeteorlakePkg/Include/CpuInitDataHob.h
```diff
@@ -65,4 +65,11 @@ typedef struct {
   **/
   UINT8                  PerAtomClusterRatio[CPU_MAX_ATOM_CLUSTERS];
 } CPU_INIT_DATA_HOB;
+///
+/// This HOB is used to pass only the required information from PEI for DXE consumption.
```
bejeanmo, line 69: Should remove comments about PEI/DXE since they aren't relevant to SBL. May cause some confusion.

gdong1 (approved):

PR #2441, opened 2025-04-17
[MTL] fix the AE_AML_PACKAGE_LIMIT error

stanleyintel:
This commit fixes the AE_AML_PACKAGE_LIMIT error after introducing GpioV2. It also deletes the unused ASL files.

Before the commit, the calling of \PIN.ON (PWRG) and \PIN.OFF (RSTG) was using non-GPIOV2 format/values. These functions end up calling _SB.GGOV and _SB.SGOV, which rely on GADR(), and this is causing an error with GINF().

The error messages by Linux kernel without the commit are:

[   53.662055] Initialized Arguments for Method [GINF]:  (3 arguments defined for method invocation)
[   53.662056]   Arg0:   000000001675faad <Obj>           Integer 0000000000000000
[   53.662060]   Arg1:   0000000064dfbf0d <Obj>           Integer 0000000000000008
[   53.662064]   Arg2:   000000003e76b0ee <Obj>           Integer 0000000000000001

[   53.662071] ACPI Error: Aborting method \_SB.GINF due to previous error (AE_AML_PACKAGE_LIMIT) (20240827/psparse-529)
[   53.662093] ACPI Error: Aborting method \_SB.GADR due to previous error (AE_AML_PACKAGE_LIMIT) (20240827/psparse-529)
[   53.662099] ACPI Error: Aborting method \_SB.SGOV due to previous error (AE_AML_PACKAGE_LIMIT) (20240827/psparse-529)
[   53.662104] ACPI Error: Aborting method \PIN.ON due to previous error (AE_AML_PACKAGE_LIMIT) (20240827/psparse-529)
[   53.662109] ACPI Error: Aborting method \_SB.PEPD._DSM due to previous error (AE_AML_PACKAGE_LIMIT) (20240827/psparse-529)

The commit is tested with Ubuntu on both MTL-P CRB and MTL-PS CRB.


Platform/MeteorlakeBoardPkg/AcpiTables/Dsdt/Pep.asl
```diff
@@ -717,14 +717,14 @@ Scope(\_SB)
             // standby state with very limited SW activities
             \_SB.PC00.RP01.DL23()
 #if FixedPcdGetBool(PcdMtlPSSupport) == 1
-            Store(0x14040B, Index(RSTG, 0))
+            Store(0xA100B, Index(RSTG, 0)) // GPIOV2_MTL_SOC_M_GPP_A11
```
gdong1, line 720: Late we might be able to use GPIOV2_MTL_SOC_M_GPP_A11 directly. Here the ChipsetID (0xA) is used in the hard-coded value. Actually it should be 1 for PAD GPIOV2_MTL_SOC_M_GPP_A11. maybe ChipsetID is not used in ACPI, but it is good to match with GPIOV2_MTL_SOC_M_GPP_A11.

stanleyintel: Thanks for pointing it out. Your guess is correct: the ChipsetID part (GPIOV2_PAD_POS_CHIPSETID and GPIOV2_PAD_MASK_CHIPSETID) is not used, which is why 0xA worked. I have revised the patch to address the ChipsetID issue.

gdong1 (approved):

PR #2453, opened 2025-04-25
feat: [BTL-S] Add BTL-S builds to Azure CI pipelines

bejeanmo:
Add BTL-S build targets to azure build tests for pre-commit check.

### DIFF
--- a/.azurepipelines/azure-pipelines.yml
+++ b/.azurepipelines/azure-pipelines.yml
@@ -111,6 +111,10 @@ jobs:
         Build.Name:   "rpls"
         Build.Arch:   ""
         Build.Target: ""
+      RPLS_X64_DEBUG:
+        Build.Name:   "rpls"
+        Build.Arch:   "-a x64"
+        Build.Target: ""
       RPLP_X86_DEBUG:
         Build.Name:   "rplp"
         Build.Arch:   ""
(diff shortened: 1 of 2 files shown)


.azurepipelines/azure-pipelines.yml
```diff
@@ -380,6 +396,14 @@ jobs:
...
         Build.Arch:   "-a x64"
         Build.Target: ""
+      BTLS_X86_DEBUG:
+        Build.Name:   "btls"
+        Build.Arch:   ""
+        Build.Target: ""
+      BTLS_X86_RELEASE:
+        Build.Name:   "btls"
+        Build.Arch:   ""
+        Build.Target: "-r"
```
gdong1, line 406: Does TTLS support X64 SBL? if yes had better have one build to cover X64 build.

bejeanmo: Good point @gdong1. BTL-S should essentially have parity with RPL-S so I added debug x64 builds. I also noticed x64 build was failing with OS Loader space overflow on ubuntu 20.04 and that this build wasn't included for RPL-S. I fixed the build error on both and added the RPL-S x64 debug build for ubuntu 20.04 (previously only windows was running x64).

gdong1 (approved):

PR #2460, opened 2025-05-03
fix: [ADL][RPL] Change SmmRebase mode to Auto NoSmrr for edk2 202411

vincentchchen:
edk2-stable202411 based UEFI payload requires Smm Rebase without setting SMRR. Set this new auto mode for compatibility with this and newer uefi payload.


randylintw (approved):
LGTM

PR #2463, opened 2025-05-13
fix: [common] Fix Coverity issues in RpmbLib.

randylintw:
Out-of-bounds access (OVERRUN)
overrun-buffer-arg: Overrunning array Frames->Data of 256 bytes by passing it to a function which accesses it at byte offset 283 using argument 284U

### DIFF
--- a/BootloaderCommonPkg/Library/RpmbLib/RpmbLib.c
+++ b/BootloaderCommonPkg/Library/RpmbLib/RpmbLib.c
@@ -42,8 +42,7 @@ RpmbCalcHmacSha256(
   OUT UINT8               Mac[]
   )
 {
-  //TODO: Andriod does HMAC_Update in a loop of BlockCnt. Change this later if we need BlocksCnt > 1
-  return HmacSha256 ((UINT8 *)Frames->Data, HMAC_DATA_LEN, Key, KeySize, Mac, RPMB_MAC_SIZE);
+  return HmacSha256 ((UINT8 *)Frames->Data, sizeof(Frames->Data), Key, KeySize, Mac, RPMB_MAC_SIZE);
 }
 
 /**


ywho79 (approved):

BootloaderCommonPkg/Library/RpmbLib/RpmbLib.c
```diff
@@ -43,6 +43,10 @@ RpmbCalcHmacSha256(
   )
 {
   //TODO: Andriod does HMAC_Update in a loop of BlockCnt. Change this later if we need BlocksCnt > 1
+  if (HMAC_DATA_LEN > sizeof(Frames->Data)) {
+    DEBUG ((DEBUG_ERROR, "RpmbCalcHmacSha256: HMAC_DATA_LEN too large\n"));
+    return 1;
+  }
   return HmacSha256 ((UINT8 *)Frames->Data, HMAC_DATA_LEN, Key, KeySize, Mac, RPMB_MAC_SIZE);
```
gdong1, line 50: The HMAC_DATA_LEN macro is incorrectly assuming that the Data field spans from its offset to the end of the RPMB_DATA_FRAME structure. However, the Data field is only 256 bytes, and the remaining fields (Nonce, WriteCounter, etc.) follow it in memory.  Maybe you could directly use "return HmacSha256((UINT8 *)Frames->Data, sizeof(Frames->Data), Key, KeySize, Mac, RPMB_MAC_SIZE);"?

randylintw: no idea why the origin programmer use this way to count the length..
284 = HMAC_DATA_LEN = (sizeof(RPMB_DATA_FRAME) - ((UINTN) &(((RPMB_DATA_FRAME *)0)->Data))) 
btw, replace the  HMAC_DATA_LEN with sizeof (data), coverity still complian..
_return HmacSha256 ((UINT8 *)Frames->Data, sizeof(Frames->Data), Key, KeySize, Mac, RPMB_MAC_SIZE);_
overrun-call: Overrunning callee's array of size 128 by passing argument 256U in call to HmacSha256.[hide details]

randylintw:
Another ticket will trigger a false alarm because the data size won't be 128.

PR #2466, opened 2025-05-19
fix: [ARL] Set SMI lock bit for Osloader

tsaikevin:
Set SMI lock bit at ReadyToBoot for Osloader


randylintw:
devmem2 0xfe001024 result = 0x3E2210. passed.

bejeanmo (approved):

PR #2469, opened 2025-05-21
fix: [EHL] Set SMI Lock bit for OsLoader

randylintw:
Set SMI Lock bit at ReadyToBoot for OsLoader before jumping to OS

verified on EHL CRB.


gdong1 (approved):

gdong1:
The CI build failure was caused by vmImage ubuntu20.04 was removed in the CI build system. It was fixed to using ubuntu24.04. The change in this PR is simple and passed all other checks. so just merge it.

PR #2470, opened 2025-05-21
fix: [ARL][MTL] Fix Coverity issues

vincentchchen:
ARL
- Execute32BitCode: CWE-476 Deference null return value

MTL
- PlatformUpdateAcpiGnvs: CWE-563 Unused value
- InitializeSmbiosInfo: CWE-476 Dereference null return value
- EarlyPlatformDataCheck: CWE-569 Operands don't affect result

### DIFF
--- a/Silicon/MeteorlakePkg/Include/Library/PchInfoLib.h
+++ b/Silicon/MeteorlakePkg/Include/Library/PchInfoLib.h
@@ -1,7 +1,7 @@
 /** @file
   Header file for PchInfoLib.
 
-  Copyright (c) 2020 - 2023, Intel Corporation. All rights reserved.<BR>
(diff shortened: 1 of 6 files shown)


Silicon/MeteorlakePkg/Include/Library/PchInfoLib.h
```diff
@@ -738,4 +738,77 @@ BOOLEAN
...
+
+/**
+  Get Serial IO SPI controller PCIe Device Number
+
+  @param[in]  SpiNumber       Serial IO SPI controller index
+
+  @retval Serial IO SPI controller PCIe Device Number
+**/
+UINT8
+MtlPchSerialIoSpiDevNumber (
```
gdong1, line 750: For the library API functions, often we will add EFIAPI in both header and c files.

PR #2475, opened 2025-06-06
fix: [ARL] Remove dependency on SblOpen folder for stitching

stanleyintel:
This commit addresses the issue of hard dependency on the SblOpen folder, allowing stitching to proceed without it.

### DIFF
--- a/Platform/ArrowlakeBoardPkg/Script/StitchIfwi.py
+++ b/Platform/ArrowlakeBoardPkg/Script/StitchIfwi.py
@@ -29,11 +29,11 @@ sys.dont_write_bytecode = True
 # sign_bin_flag can be set to false to avoid signing process. Applicable for Btg profile 0
 sign_bin_flag = True
 
-sblopen_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), '../../../../', 'SblOpen')
+sblopen_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../'))
 if not os.path.exists (sblopen_dir):
     sblopen_dir = os.getenv('SBL_SOURCE', '')
-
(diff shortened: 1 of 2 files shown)


bejeanmo:
@stanleyintel I'm not sure I follow your script change. Can you explain it further? It seems like an exception will still be thrown if the SblOpen folder is missing.

stanleyintel:
> @stanleyintel I'm not sure I follow your script change. Can you explain it further? It seems like an exception will still be thrown if the SblOpen folder is missing.

Hi @bejeanmo, 

Thanks for reviewing the PR and it is a good question. Let me explain it with the example first. Please try with the following directory structure:
.
├── sbl            # SBL working directory
└── SblOpen  #  an empty directory, no files/dirs under it

then, 
$ cd sbl
$ python Platform/ArrowlakeBoardPkg/Script/StitchIfwi.py -b .... (skip)

It will result in errors (shown at the end of this reply).

The original script constructs the path to SblOpen and appends it to the system path, expecting to find the necessary modules there. However, since SblOpen is empty, the module cannot be found, resulting in a ModuleNotFoundError.

PS 1. Similar patches were also developed for CML/CFL/TGL/EHL/ADL (commit: 7400d6) and MTL (commit: 9002ea).

PS 2. Error messages:

Traceback (most recent call last):
  File "/tmp/xyz/sbl/Platform/ArrowlakeBoardPkg/Script/StitchLoader.py", line 22, in <module>
    from   IfwiUtility import *
ModuleNotFoundError: No module named 'IfwiUtility'

bejeanmo:
Ok. What is the behavior after this modification when SblOpen is missing?

stanleyintel:
> e

Yes

PR #2499, opened 2025-07-22
feat: Update QEMU FSP to use edk2-stable202411 and support VS2022

bejeanmo:
Update QEMU FSP build to use edk2-stable202411 to match other components. This also adds VS2022 support for building QEMU FSP.


bejeanmo:
GCC build fixed, but getting an exception on FSP-S entry with GCC built SlimBootloader.

PR #2501, opened 2025-08-04
feat: OS loader shell command to launch setup payload

bejeanmo:
Added a new 'setup' command to OS Loader shell to launch the setup payload available when the feature is enabled.

### DIFF
--- a/BootloaderCommonPkg/Library/ShellLib/CmdSetup.c
+++ b/BootloaderCommonPkg/Library/ShellLib/CmdSetup.c
@@ -1,0 +1,100 @@
+/** @file
+  Shell command `setup` to launch setup payload.
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Library/ShellLib.h>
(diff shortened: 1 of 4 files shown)


BootloaderCommonPkg/Library/ShellLib/CmdSetup.c
```diff
@@ -0,0 +1,105 @@
...
+{
+  OS_BOOT_OPTION_LIST       *BootOptionList;
+  UINT8                     BootOptionIndex;
+  EFI_HOB_GUID_TYPE         *BootOptionHob;
+  UINT32                    MaxBootOptions;
+
+  if (!FeaturePcdGet (PcdEnableSetup)) {
+    ShellPrint (L"Setup Payload support is not enabled in this bootloader.\n");
+    return EFI_UNSUPPORTED;
+  }
```
spkamat: Isn't this check redundant as the command itself is made visible based on this check in ShellCmds.c?

bejeanmo: You're correct. I thought i was seeing an edge case where this could get executed without registering the command, but I just had a typo in my working tree.

spkamat (approved):

PR #2508, opened 2025-08-14
fix: [Common] Linux build support for Fwupdate

tsaikevin:
Add EFIAPI to fwupdate function definitions for GCC build
Update BuildLoader script to allow CSME driver build

### DIFF
--- a/PayloadPkg/FirmwareUpdate/CsmeFwUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/CsmeFwUpdate.c
@@ -44,40 +44,6 @@ FindImage (
   );
 
 /**
-  Main routine to update CSME firmware.
-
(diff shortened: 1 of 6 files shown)


PayloadPkg/FirmwareUpdate/CsmeFwUpdate.c
```diff
@@ -72,6 +72,7 @@ StartCsmeUpdate (
   @param[in] totalBytesToSendToFw  The total number of bytes of the buffer.
 **/
 static void
+EFIAPI
```
gdong1: If the function definition is present in the same source file and there are no calls before its definition, the separate function declaration is not needed. So I think it is better to remove code from line 46 to 79.

PR #2512, opened 2025-08-21
feat: [Common] Add FuSa boot option hook

tsaikevin:
Add FuSa boot option hook in null library

### DIFF
--- a/Silicon/CommonSocPkg/Include/Library/FusaConfigLib.h
+++ b/Silicon/CommonSocPkg/Include/Library/FusaConfigLib.h
@@ -46,4 +46,17 @@ FusaConfigPostSi (
   VOID
   );
 
+/**
+  Perform Fusa Boot Option Configuration required.
+
+  @param OsBootOptionList The pointer to boot option list
+
+  @retval EFI_STATUS
+ */
+EFI_STATUS
+EFIAPI
+FusaConfigBootOption (
+  VOID *OsBootOptionList
+  );
+
 #endif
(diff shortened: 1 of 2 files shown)


Silicon/CommonSocPkg/Include/Library/FusaConfigLib.h
```diff
@@ -46,4 +46,15 @@ FusaConfigPostSi (
   VOID
   );
 
+/**
+  Perform Fusa Boot Option Configuration required.
+
+  @param OsBootOptionList The pointer to boot option list
+ */
```
spkamat, line 55: retval missing

Silicon/CommonSocPkg/Library/FusaConfigLibNull/FusaConfigLibNull.c
```diff
@@ -54,3 +54,17 @@ FusaConfigPostSi (
 {
   return EFI_UNSUPPORTED;
 }
+
+/**
+  Perform Fusa Boot Option Configuration required.
+
+  @param OsBootOptionList The pointer to boot option list
+ */
```
spkamat, line 64: retval missing

spkamat:
Wouldn't it be better to add this hook once we have/along with a real consumer for it.

tsaikevin:
> Wouldn't it be better to add this hook once we have/along with a real consumer for it.

@spkamat yes, we're working on new project that requires NULL lib add this function. fyi

PR #2515, opened 2025-08-28
feat: [ARLS] Added SMBIOS Production Data Override

kokweich:
Added SMBIOS Data Override feature in common library for SBL2.0 DeviceInfo binary will be loaded and used to override default SMBIOS value DeviceInfo binary can be generated using GenSmbiosOverrideBin.py in the Tools folder Feature can be enabled/disabled via ENABLE_SMBIOS_OVERRIDE in boardconfig

### DIFF
--- a/Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/SmbiosInit.c
+++ b/Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/SmbiosInit.c
@@ -6,6 +6,7 @@
 **/
 
 #include "Stage2BoardInitLib.h"
+#include <Library/SmbiosOverride.h>
 #include <Library/PrintLib.h>
 
 /**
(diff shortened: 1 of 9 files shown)


Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/SmbiosInit.c
```diff
@@ -67,6 +80,55 @@ InitializeSmbiosInfo (
...
+    BaseboardProductName = BaseboardStrBuf;
+
+    // Load Device Info binary and Override Smbios value
+    if (FeaturePcdGet (PcdSmbiosOverride)) {
+      DEVICE_INFO_DATA     *DeviceInfo = NULL;
+      UINT32               DeviceInfoSize = 0;
+      EFI_STATUS           Status;
+
+      Status = LoadComponent (SIGNATURE_32 ('D', 'E', 'V', 'I'), SIGNATURE_32 ('D', 'I', 'N', 'F'),
+                            (VOID **)&DeviceInfo, &DeviceInfoSize);
```
gdong1, line 110: It is not necessary to add SmbiosOverrideLib, it makes SBL complicated and bigger. 
The default production values could comes from PCDs as EDK2 did. and these default values will be updated by SBL container IFWI/BIOS/NRD/IPFW/SMBS if the container exists.

randylintw:
Is it possible to make it as common feature ?

PR #2518, opened 2025-09-02
Sbl enhancements

gdong1:


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -538,6 +538,7 @@ SecStartup (
     AddMeasurePoint (0x3050);
     if (Status == EFI_NOT_FOUND) {
       SplashPostPci = TRUE;
```
spkamat, line 537: A low level verbosity/info message might help in debug here.

gdong1: Thanks for the comment. it was added in the updated patch.

PR #2526, opened 2025-09-12
feat: Added SMBIOS Production Data Override support

kokweich:
Added SMBIOS Data Override feature in common library

DeviceInfo binary can be generated using GenSmbiosOverrideBin.py in the Tools folder
DeviceInfo binary will be loaded if exist under <Platform>/<platformBoardPkg>/Binaries
The binary is used to override default SMBIOS value

### DIFF
--- a/BootloaderCorePkg/Stage2/Stage2.c
+++ b/BootloaderCorePkg/Stage2/Stage2.c
@@ -522,6 +522,10 @@ SecStartup (
   }
 
   BoardInit (PostSiliconInit);
+
+  if (FixedPcdGetBool (PcdSmbiosEnabled)) {
+    ApplySmbiosDeviceInfoOverride ();
+  }
   AddMeasurePoint (0x3040);
 
 #if FixedPcdGetBool (PcdEnableCryptoPerfTest)
--- a/BootloaderCorePkg/Include/Library/SmbiosInitLib.h
+++ b/BootloaderCorePkg/Include/Library/SmbiosInitLib.h
@@ -19,6 +19,34 @@ typedef struct {
   CHAR8     *String;
 } SMBIOS_TYPE_STRINGS;
 
+typedef struct {
+  // System Information (SMBIOS Type 1)
+  CHAR8   SystemManufacturer[64];             // System manufacturer string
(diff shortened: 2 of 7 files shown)


BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -522,6 +654,26 @@ SecStartup (
...
+      Status = ApplySmbiosDeviceInfoOverride(DeviceInfo);
+      if (EFI_ERROR(Status)) {
+        DEBUG((DEBUG_WARN, "Failed to apply SMBIOS device info override: %r\n", Status));
+      } else {
+        DEBUG((DEBUG_INFO, "SMBIOS device info override applied successfully\n"));
+      }
+    } else {
+      DEBUG((DEBUG_INFO, "Device info binary not found: %r\n", Status));
+    }
+  }
```
gdong1, line 528: Is it better do it in SmBiosLib SmbiosInit()?

BuildLoader.py
```diff
@@ -1446,6 +1484,12 @@ def post_build(self):
         if getattr(self._board, "GetContainerList", None):
             container_list = self._board.GetContainerList ()
             component_dir = os.path.join(os.environ['PLT_SOURCE'], 'Platform', self._board.BOARD_PKG_NAME, 'Binaries')
+            device_info_file = os.path.join(component_dir, 'DeviceInfo.bin')
+            if os.path.exists(device_info_file):
+                print(f"Found DeviceInfo.bin at {device_info_file}, adding to container list")
+                self.add_device_info_container(container_list, device_info_file)
+            else:
+                print(f"DeviceInfo.bin not found at {device_info_file}, skipping container list addition")
```
gdong1: Avoid updating common build tool for a specific component if possible. How about just updating BoardConfig script adding TMAC for MAC address?

BootloaderCorePkg/Stage2/Stage2.c
kokweich: yeah thanks for the suggestion, its better to add in smbios init(), will update it

PR #2530, opened 2025-09-23
[TGL][TXT] Improved cmos logic for TXT disable usecase

aborwank:
When TXT Feature is disabled TXT_CMOS_STATUS_REG should be set to disable so that ACM does not try to enable TXT.

This was not happening when previously TXT was enabled and then new ifwi was flashed with TXT feature disabled.

This is fixed by modifying the if condition in which the CMOS is cleared.

### DIFF
--- a/Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
+++ b/Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
@@ -389,32 +389,31 @@ UpdateFspConfig (
   Fspmcfg->SafeMode = 0x1;                // Need enable, else failed in MRC
 
   // TXT Configuration
-  if (FeaturePcdGet (PcdTxtEnabled)) {
-    FeaturesCfgData = (FEATURES_CFG_DATA *) FindConfigDataByTag(CDATA_FEATURES_TAG);
-    if (FeaturesCfgData->Features.TxtEnabled == 1) {
-      DEBUG((DEBUG_INFO, "Enabling TXT in FSP-M UPD's\n"));
-      Fspmcfg->Txt                  = 0x1;
-      Fspmcfg->TxtImplemented       = 0x1;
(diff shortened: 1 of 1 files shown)


Platform/TigerlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -389,32 +389,31 @@ UpdateFspConfig (
...
       IoWrite8 (R_IOPORT_CMOS_STANDARD_DATA, CmosData);
     }
- }
+  } else {
+    IoWrite8 (R_IOPORT_CMOS_STANDARD_INDEX, TXT_CMOS_STATUS_REG);
+    UINT8 CmosData = IoRead8 (R_IOPORT_CMOS_STANDARD_DATA);
+    CmosData &= ~BIT4;
+    IoWrite8 (R_IOPORT_CMOS_STANDARD_INDEX, TXT_CMOS_STATUS_REG);
+    IoWrite8 (R_IOPORT_CMOS_STANDARD_DATA, CmosData);
+  }
```
gdong1, line 416: Do you think it is beneficial to include a signature in the CMOS data to indicate its validity? When TXT is enabled in SBL, SBL will write the CMOS TXT data along with a signature. If the signature is absent, it means TXT is disabled in SBL. This approach ensures that the ACM is not affected when the TXT code is excluded during builds where TXT is disabled.

lsubash: ACM do not have support to read CMOS signature. 
ACM reads CMOS variables based on TXT FIT record. If TXT is disabled in SBL build, ACM will not have any CMOS action.

gdong1: ok. It looks there is no plan to update TGL ACM. Let's use this update.

PR #2545, opened 2025-10-07
[Common] Cleanup: Disablement of A/B support

sahana-n-jain:
Added macro control for A/B support through
AbSlotConfig.h file. It allows disabling and
enabling A/B functionality using 0 and 1
respectively.


BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -331,6 +331,7 @@
   gPlatformCommonLibTokenSpaceGuid.PcdMinDecompression    | FALSE
   gPlatformCommonLibTokenSpaceGuid.PcdMeasuredBootEnabled | $(HAVE_MEASURED_BOOT)
   gPlatformCommonLibTokenSpaceGuid.PcdVerifiedBootEnabled | $(HAVE_VERIFIED_BOOT)
+  gPlatformCommonLibTokenSpaceGuid.PcdAbSlotSupportEnabled| FALSE
```
gdong1: Better update this PCD using ENABLE_AB_SLOT_SUPPORT, and set ENABLE_AB_SLOT_SUPPORT to TRUE by default to keep back-compatible.  In the platform that doesn't need it, could set it to FALSE in the boardconfig.py

sahana-n-jain: Updated accordingly

PR #2549, opened 2025-10-14
feat: [Common] Add support to load payload from PDR

spkamat:
Added support to load EPayload from PDR as an alternate loading method. This is useful if the overall bootloader binary cannot fit inside the BIOS region or there is a need to load multiple payloads or load an OS as a payload. To enable this feature, user should set self.ENABLE_PDR_LOAD to 1 in the corresponding BoardConfig file.

### DIFF
--- a/BootloaderCorePkg/BootloaderCorePkg.dec
+++ b/BootloaderCorePkg/BootloaderCorePkg.dec
@@ -255,6 +255,7 @@
   gPlatformModuleTokenSpaceGuid.PcdResizableBarSupport    | FALSE      | BOOLEAN | 0x20000216
   gPlatformModuleTokenSpaceGuid.PcdEnablePciePm           | FALSE      | BOOLEAN | 0x20000222
   gPlatformModuleTokenSpaceGuid.PcdEnableFwuNotify        | FALSE      | BOOLEAN | 0x20000225
+  gPlatformModuleTokenSpaceGuid.PcdPayloadLoadFromPDR     | FALSE      | BOOLEAN | 0x20000230
 
 [PcdsDynamic]
   gPlatformModuleTokenSpaceGuid.PcdFspResetStatus         | 0          | UINT64 | 0x20000224
--- a/BootloaderCorePkg/BootloaderCorePkg.dsc
+++ b/BootloaderCorePkg/BootloaderCorePkg.dsc
@@ -365,6 +365,7 @@
   gPlatformModuleTokenSpaceGuid.PcdEnablePciePm           | $(ENABLE_PCIE_PM)
   gPlatformCommonLibTokenSpaceGuid.PcdFspNoEop            | $(HAVE_NO_FSP_EOP)
   gPlatformModuleTokenSpaceGuid.PcdEnableFwuNotify        | $(ENABLE_FWU_NOTIFY)
+  gPlatformModuleTokenSpaceGuid.PcdPayloadLoadFromPDR     | $(ENABLE_EPLD_FROM_PDR)
   gPlatformCommonLibTokenSpaceGuid.PcdTxtEnabled          | $(TXT_ENABLED)
 
 
--- a/BootloaderCorePkg/Stage2/Stage2.c
+++ b/BootloaderCorePkg/Stage2/Stage2.c
@@ -6,6 +6,7 @@
 **/
 
 #include "Stage2.h"
+#include <Service/SpiFlashService.h>
 
 /**
   Callback function to add performance measure point during component loading.
@@ -63,7 +64,7 @@ PreparePayload (
   IN STAGE2_PARAM   *Stage2Param
   )
 {
-  EFI_STATUS                     Status;
+  EFI_STATUS                     Status = EFI_NOT_STARTED;  // Initialize status to indicate PDR loading not attempted;
   UINT32                         Dst;
   UINT32                         DstLen;
   VOID                          *DstAdr;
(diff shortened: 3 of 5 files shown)


BootloaderCorePkg/BootloaderCorePkg.dec
```diff
@@ -255,6 +255,7 @@
   gPlatformModuleTokenSpaceGuid.PcdResizableBarSupport    | FALSE      | BOOLEAN | 0x20000216
   gPlatformModuleTokenSpaceGuid.PcdEnablePciePm           | FALSE      | BOOLEAN | 0x20000222
   gPlatformModuleTokenSpaceGuid.PcdEnableFwuNotify        | FALSE      | BOOLEAN | 0x20000225
+  gPlatformModuleTokenSpaceGuid.PcdEnableLoadingFromPdr   | FALSE      | BOOLEAN | 0x20000230
```
rprangar: May need to be more precise - PcdPayloadLoadFromPDR?

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -108,8 +110,72 @@ PreparePayload (
...
+      // Get PDR region info
+      Status = SpiFlashService->SpiGetRegion (FlashRegionPlatformData, &PdrBaseAddress, &PdrSize);
+      if (!EFI_ERROR (Status)) {
+        UINT8 *PdrBuffer = NULL;
+
+        // Allocate buffer to read container from PDR (kept persistent for container access)
+        PdrBuffer = (UINT8 *) AllocateTemporaryMemory (PdrSize);
+        if (PdrBuffer != NULL) {
+          // Read container from PDR region
+          Status = SpiFlashService->SpiRead (FlashRegionPlatformData, 0, PdrSize, PdrBuffer);
```
rprangar, line 133: Either need to capture the assumption here - That the payload is at offset 0 in the PDR - OR - We may need a way to find the offset in the PDR

BootloaderCorePkg/BootloaderCorePkg.dec
spkamat: Updated the name.

BootloaderCorePkg/Stage2/Stage2.c
spkamat: Updated the comment for now.

BootloaderCorePkg/BootloaderCorePkg.dsc
```diff
@@ -365,6 +365,7 @@
   gPlatformModuleTokenSpaceGuid.PcdEnablePciePm           | $(ENABLE_PCIE_PM)
   gPlatformCommonLibTokenSpaceGuid.PcdFspNoEop            | $(HAVE_NO_FSP_EOP)
   gPlatformModuleTokenSpaceGuid.PcdEnableFwuNotify        | $(ENABLE_FWU_NOTIFY)
+  gPlatformModuleTokenSpaceGuid.PcdPayloadLoadFromPDR     | $(ENABLE_PDR_LOAD)
```
gdong1: using ENABLE_EPLD_FROM_PDR to align with the PCD name?

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -108,8 +110,72 @@ PreparePayload (
...
+          DEBUG ((DEBUG_ERROR, "Failed to allocate buffer for PDR container\n"));
+        }
+      } else {
+        DEBUG ((DEBUG_ERROR, "Failed to get PDR region info, Status = %r\n", Status));
+      }
+    } else {
+      Status = EFI_DEVICE_ERROR;
+      DEBUG ((DEBUG_ERROR, "SPI Flash Service not available\n"));
+    }
+  }
```
gdong1, line 169: This change adds support for loading EPAYLOAD from the PDR (Platform Data Region) using the SPI flash service. While the implementation functions correctly, the current code structure contains excessive nesting that impacts readability and maintainability. Prefer update it using early return.

BootloaderCorePkg/Stage2/Stage2.c
```diff
@@ -108,8 +110,72 @@ PreparePayload (
...
+
+  // If PDR loading is disabled or failed or not EPAYLOAD, fall back to normal container loading
+  if (!FeaturePcdGet (PcdPayloadLoadFromPDR) || EFI_ERROR (Status) || (ContainerSig != FLASH_MAP_SIG_EPAYLOAD)) {
```
gdong1, line 172: this is 'else' condition for the 'if' condition in line 115.

BootloaderCorePkg/BootloaderCorePkg.dsc
spkamat: Updated to ENABLE_EPLD_FROM_PDR.

BootloaderCorePkg/Stage2/Stage2.c
spkamat: The nesting of if-else is required to handle error scenarios at various levels. Debug Info prints should help with the readability. Early return will not handle the fallback scenario if load from PDR fails.

BootloaderCorePkg/Stage2/Stage2.c
spkamat: It is not made 'else' to handle the fallback case if PDR load fails at any stage due to any error.

BootloaderCorePkg/Stage2/Stage2.c
gdong1: I think eliminating deep nesting could have better readability. and you could keep the debug message same. e.g.: changing to below is clearly visible as the main flow.  BTW, following coding style, the variable definition should be at the start of function. 
if (FeaturePcdGet (PcdPayloadLoadFromPDR) && (ContainerSig == FLASH_MAP_SIG_EPAYLOAD)) {
  DEBUG ((DEBUG_INFO, "Alternate Loading payload from PDR region !\n"));

  SPI_FLASH_SERVICE *SpiFlashService;
  SpiFlashService = (SPI_FLASH_SERVICE *) GetServiceBySignature (SPI_FLASH_SERVICE_SIGNATURE);
  if (SpiFlashService == NULL) {
    DEBUG ((DEBUG_ERROR, "SPI Flash Service not available\n"));
    return 0;
  }

  UINT32 PdrBaseAddress = 0;
  UINT32 PdrSize = 0;

  Status = SpiFlashService->SpiGetRegion (FlashRegionPlatformData, &PdrBaseAddress, &PdrSize);
  if (EFI_ERROR (Status)) {
    DEBUG ((DEBUG_ERROR, "Failed to get PDR region info, Status = %r\n", Status));
    return 0;
  }

  UINT8 *PdrBuffer = (UINT8 *) AllocateTemporaryMemory (PdrSize);
  if (PdrBuffer == NULL) {
    DEBUG ((DEBUG_ERROR, "Failed to allocate buffer for PDR container\n"));
    return 0;
  }

  Status = SpiFlashService->SpiRead (FlashRegionPlatformData, 0, PdrSize, PdrBuffer);
  if (EFI_ERROR (Status)) {
    DEBUG ((DEBUG_ERROR, "Failed to read from PDR region, Status = %r\n", Status));
    FreeTemporaryMemory (PdrBuffer);
    return 0;
  }

  Status = RegisterContainer ((UINT32)(UINTN)PdrBuffer, LoadComponentCallback);
  if (EFI_ERROR (Status)) {
    DEBUG ((DEBUG_ERROR, "Failed to register PDR container, Status = %r\n", Status));
    FreeTemporaryMemory (PdrBuffer);
    return 0;
  }

  DEBUG ((DEBUG_INFO, "PDR container registered successfully\n"));

  Status = LoadComponentWithCallback (FLASH_MAP_SIG_EPAYLOAD, ComponentName,
                                      &DstAdr, &DstLen, LoadComponentCallback);
  if (EFI_ERROR (Status)) {
    DEBUG ((DEBUG_ERROR, "Failed to load EPAYLOAD component from PDR container, Status = %r\n", Status));
    UnregisterContainer (FLASH_MAP_SIG_EPAYLOAD);
    FreeTemporaryMemory (PdrBuffer);
    return 0;
  }

  DEBUG ((DEBUG_INFO, "PDR EPAYLOAD component loaded at 0x%08X, size 0x%08X\n", DstAdr, DstLen));
}  else { original code }

spkamat: I will move the variable definition at the declaration stage itself.
As I said in my previous comment the flow has been designed to support a fallback option to default BIOS region loading in case PDR load fails. With the suggested approach of 'return from fail' to avoid nesting and putting the non-PDR load case in else, fallback option would be lost.

BootloaderCorePkg/Stage2/Stage2.c
gdong1: It is a little confusing on the design (having 2 epayload in the IFWI?). and PDR region could be used for other purpose (e.g.: self.CFGDATA_REGION_TYPE could be configured get config data from PDR), the assumption to get epayload from offset 0 might have confliction.

PR #2596, opened 2025-12-17
feat: Added support for CPU FMS in Diagnostic ACM fit entry

tsaikevin:
Add fit entry version 0x200 to support family model mask(FMS) format

### DIFF
--- a/BootloaderCorePkg/Tools/IfwiUtility.py
+++ b/BootloaderCorePkg/Tools/IfwiUtility.py
@@ -48,6 +48,23 @@ class FIT_ENTRY(Structure):
         self.type     = _type
         self.checksum = _checksum
 
+class FIT_ENTRY_V200(Structure):
+    # Support FIT entry version 0x200 structure
+    _pack_ = 1
+    _fields_ = [
+        ('address',  c_uint64),
+        ('fms',      c_uint32),
+        ('version',  c_uint16),
+        ('type',     c_uint8), # Bit[7] = C_V
+        ('ext_fms',  c_uint8),
+        ]
+
+    def set_values(self, _address, _fms, _version, _type, _ext_fms):
+        self.address  = _address
+        self.fms      = _fms
+        self.version  = _version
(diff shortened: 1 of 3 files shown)


BuildLoader.py
```diff
@@ -408,19 +410,19 @@ def update_fit_table (self):
 
         # ACM
         if self._board.ACM_SIZE > 0:
-            # Add default enrty for ACM FIT Version 0x100
-            if self._board.ACM_FIT_VERISON == 0x100:
+            if self._board._ACM_CPU_FMS != []:
```
lsubash: This should based only on ACM_FIT_VERISON.  _ACM_CPU_FMS  will be applied only for ACM_FIT_VERISON 0x200. ucode will check loading based on 0x200.  ACM entry version is different from FIT table version

tsaikevin: @lsubash PR updated. fyi

BootloaderCorePkg/Tools/IfwiUtility.py
```diff
@@ -48,6 +48,23 @@ def set_values(self, _address, _size, _version, _type, _checksum):
         self.type     = _type
         self.checksum = _checksum
 
+class FIT_ENTRY_V200(Structure):
```
lsubash, line 51: This is same as original FIT_ENTRY. Redefinition is not needed.
We have to update the base FIT_ENTRY definition as per the latest entries 
https://github.com/andreiw/lampone-edk2-platforms/blob/0d5a9b9e01b1fd5084276b60a49980bdfb00cc94/Silicon/Intel/Tools/FitGen/FitGen.c#L24 

Our base structure needs only filed updates.

tsaikevin: @lsubash the FIT ENTRY version 0x200 change format with FMS bit fields replacing size from initial one. It's better to define new structure to avoid confusion.
as suggest, add comment to this structure

PR #2656, opened 2026-03-24
BootloaderCorePkg/Tools: Add sbl_dir fallback for driver_inf path lookup

spkamat:
In CopyBins(), when driver_inf is not found relative to the current working directory, fall back to resolving it relative to sbl_dir before giving up.

### DIFF
--- a/BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
+++ b/BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
@@ -193,6 +193,10 @@ def GetRepoAndCommit (driver_inf):
 
 def CopyBins (repo_dir, sbl_dir, driver_inf):
     if not os.path.exists(driver_inf):
+        driver_inf = os.path.join(sbl_dir, driver_inf)
+    if not os.path.exists(driver_inf):
+        if driver_inf:
+            print ('WARNING: driver_inf not found: %s' % driver_inf)
         return
 
     sys.stdout.flush()


BootloaderCorePkg/Tools/PrepareBuildComponentBin.py
```diff
@@ -192,6 +192,8 @@ def GetRepoAndCommit (driver_inf):
     return repo, commit
 
 def CopyBins (repo_dir, sbl_dir, driver_inf):
+    if not os.path.exists(driver_inf):
+        driver_inf = os.path.join(sbl_dir, driver_inf)
     if not os.path.exists(driver_inf):
         return
```
vtcao297, line 200: How critical are these inf driver files to the build process?  I haven't read though the entire scripts to understand the entire flow, but if the path doesn't exist and we're not returning any errors for the callee to check, are there any silent error concerns, how would the developer/person know that the file was missing to debug/fix the issue if the build fails?

spkamat: CopyBins() is called from ProcessFspAndMicrocodeInf() for two purposes: cloning/copying FSP binaries and microcode. The silent return is by design for the original code — i think some platforms don't define an FSP_INF_FILE or MICROCODE_INF_FILE, so the path won't exist and the function should just skip gracefully.

However, you're right that this creates a debuggability gap. If a platform does need the INF but the path is wrong (e.g., typo or misconfigured board config), CopyBins() silently skips, and the build fails later at split_fsp() or get_fsp_size() with an unrelated file-not-found error — making it hard to root-cause.

My change here only adds a fallback lookup (try sbl_dir-relative path when CWD-relative path doesn't exist), and preserves the existing silent-return behavior. I agree it would be better to add a warning to indicate the failure. May be I can add that.

gdong1 (approved):

PR #2659, opened 2026-03-24
feat: [ADL/RPL/BTL/TWL] move to common MADT Library

bejeanmo:
Remove Madt.act and build MADT during boot using common library instead. Replace Local APIC NMI entries per CPU with one entry covering all CPUs.


gdong1 (approved):

vincentchchen:
Mar/25
- verified the PR on ADLN and RPLP
- successfully boot SBL into Windows and check the ACPI MADT table (signature APIC)
- like what Bejean mentioned that the patch removes Madt.act and consolidates "Local APIC NMI entries for each individual CPU" into one entry with Processor ID 0xFF

PR #2665, opened 2026-03-31
Fetch ACTM binary from repo using INF file

spkamat:
Add CopyBins call in pre_build to clone ACTM binary from firmwareblob repo when ACTM_INF_FILE is defined in board config.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -1606,7 +1606,10 @@ def main():
                     fsp_inf_full_path = os.path.join(sbl_dir, fsp_inf)
                     dest_dir = sbl_dir
 
-                    plt_dir = os.environ['PLT_SOURCE'] if 'PLT_SOURCE' in os.environ else None
+                    if 'PLT_SOURCE' in os.environ:
+                        plt_dir = os.environ['PLT_SOURCE']
+                    else:
+                        plt_dir = os.path.abspath (os.path.join (os.path.dirname (board_cfgs[index]), '../..'))
 
                     if not os.path.exists(fsp_inf_full_path) and plt_dir:
                         fsp_inf_full_path = os.path.join(plt_dir, fsp_inf)
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -147,7 +147,6 @@ class BaseBoard(object):
         self.BOOT_PERFORMANCE_MASK    = 0x00000001
 
         self.HAVE_VBT_BIN          = 0
-        self.HAVE_ACTM_BIN         = 0
         self.HAVE_FIT_TABLE        = 0
         self.HAVE_VERIFIED_BOOT    = 0
         self.HAVE_MEASURED_BOOT    = 0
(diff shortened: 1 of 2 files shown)


BuildLoader.py
```diff
@@ -1317,6 +1317,11 @@ def pre_build(self):
             else:
                 gen_vbt_file (self._board.BOARD_PKG_NAME, self._board._MULTI_VBT_FILE, os.path.join(self._fv_dir, 'Vbt.bin'))
 
+        # Fetch ACTM binary from repo if ACTM_INF_FILE is defined
+        if self._board.HAVE_ACTM_BIN and hasattr(self._board, 'ACTM_INF_FILE'):
+            actm_repo_dir = os.path.abspath (os.path.join(plt_dir, '../Download', self._board.SILICON_PKG_NAME, '$AUTO'))
+            PreBuild.CopyBins (actm_repo_dir, plt_dir, self._board.ACTM_INF_FILE)
```
gdong1, line 1321: how about using [UserExtensions.SBL."CopyList"] in the INF file, similar with Microcode.inf or FSP inf file

spkamat: CopyBins() call is the pre-requisite to Actm.inf file (in the platform repo) that has [UserExtensions.SBL."CloneRepo"] and [UserExtensions.SBL."CopyList"] sections — same pattern as Microcode.inf and FspBin.inf.

gdong1: Ok. If ACTM INF uses same patten as FSP, please check:
1. FSP and Microcode both have cleanup logic in the clean() method. The PR adds no corresponding cleanup for ACTM_INF_FILE. 
2. SBL generic tool ProcessInfFileCopyList will also try to process [UserExtensions.SBL."CopyList"], FSP and Microcode INFs are explicitly excluded via the ignore_file_list parameter:
PreBuild.ProcessInfFileCopyList(sbl_dir, [self._board.FSP_INF_FILE, self._board.MICROCODE_INF_FILE])
Please check if ACTM_INF_FILE should also be added to this ignore list to prevent double-processing
3. For FSP, it checks if the FSP directory exists under plt_dir first, falling back to sbl_dir
For the ACTM, please check if need do same thing, else it might not work when upstream.

spkamat: > Ok. If ACTM INF uses same patten as FSP, please check:
> 
> 1. FSP and Microcode both have cleanup logic in the clean() method. The PR adds no corresponding cleanup for ACTM_INF_FILE.

Good catch. Added the same.

> 2. SBL generic tool ProcessInfFileCopyList will also try to process [UserExtensions.SBL."CopyList"], FSP and Microcode INFs are explicitly excluded via the ignore_file_list parameter:
>    PreBuild.ProcessInfFileCopyList(sbl_dir, [self._board.FSP_INF_FILE, self._board.MICROCODE_INF_FILE])
>    Please check if ACTM_INF_FILE should also be added to this ignore list to prevent double-processing

This doesn't seem to have any functional impact other than minor optimization related to INF parsing. However, I still added ACTM to the ignore list to maintain consistency.

> 3. For FSP, it checks if the FSP directory exists under plt_dir first, falling back to sbl_dir
>    For the ACTM, please check if need do same thing, else it might not work when upstream.

This is taken care of. Works when moved to upstream folder too.

gdong1 (approved):

PR #2672, opened 2026-04-08
Enhance firmware update to support more flags

kokweich:
Add build-time flags to GenCapsuleFirmware.py that allow selectively preserving non-volatile data regions when generating firmware update capsule images. This enables BIOS region updates without overwriting MRC training data, SBL variables, UEFI variables, or SMBIOS data.

During BIOS region update, when a skip flag is set, the corresponding region's current flash content is read and overlaid onto the capsule image buffer before writing, effectively preserving existing data.

Skip flags are incompatible with force BIOS update (-f) and the tool will raise an error if both are specified.

### DIFF
--- a/BootloaderCorePkg/Tools/GenCapsuleFirmware.py
+++ b/BootloaderCorePkg/Tools/GenCapsuleFirmware.py
@@ -405,7 +405,8 @@ FIRMWARE_UPDATE_IMAGE_FILE_GUID = uuid.UUID('{1A3EAE58-B580-4fef-ACA3-A16D9E00DF
 
 class FirmwareUpdateHeader(Structure):
 
-  CAPSULE_FLAG_FORCE_BIOS_UPDATE = 0x80000000
+  CAPSULE_FLAG_FORCE_BIOS_UPDATE            = 0x80000000
+  CAPSULE_FLAG_SKIP_ALL_NON_VOLATILE_REGION = 0x00000002
 
   _fields_ = [
     ('FileGuid',           ARRAY(c_uint8, 16)),
@@ -422,7 +423,7 @@ class FirmwareUpdateHeader(Structure):
   ]
 
 
-def SignImage(RawData, OutFile, HashType, SignScheme, PrivKey, ForceBiosUpdate):
+def SignImage(RawData, OutFile, HashType, SignScheme, PrivKey, CapsuleFlags):
 
     #
     # Generate the new image layout
@@ -445,9 +446,7 @@ def SignImage(RawData, OutFile, HashType, SignScheme, PrivKey, ForceBiosUpdate):
     header.FileGuid         = (c_ubyte *16).from_buffer_copy(FIRMWARE_UPDATE_IMAGE_FILE_GUID.bytes_le)
     header.HeaderSize       = sizeof(FirmwareUpdateHeader)
     header.FirmwreVersion   = 1
-    header.CapsuleFlags     = 0
-    if ForceBiosUpdate:
-        header.CapsuleFlags  |= FirmwareUpdateHeader.CAPSULE_FLAG_FORCE_BIOS_UPDATE
+    header.CapsuleFlags     = CapsuleFlags
     header.ImageOffset      = header.HeaderSize
     header.ImageSize        = file_size
     header.SignatureOffset  = header.ImageOffset + header.ImageSize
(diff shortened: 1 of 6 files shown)


gdong1 (approved):

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -572,10 +578,36 @@ def BuildGuid(Argument):
...
+        print("[INFO] Skipping MRC training data update during firmware update.")
+    if args.SkipSblVar:
+        CapsuleFlags |= FirmwareUpdateHeader.CAPSULE_FLAG_SKIP_UPDATE_SBL_VAR
+        print("[INFO] Skipping SBL variable storage update during firmware update.")
+    if args.SkipUefiVar:
+        CapsuleFlags |= FirmwareUpdateHeader.CAPSULE_FLAG_SKIP_UPDATE_UEFI_VAR
+        print("[INFO] Skipping UEFI variable storage update during firmware update.")
+    if args.SkipSmbios:
+        CapsuleFlags |= FirmwareUpdateHeader.CAPSULE_FLAG_SKIP_UPDATE_SMBIOS
+        print("[INFO] Skipping SMBIOS binary data update during firmwares update.")
```
spkamat, line 598: firmwares --> firmware

PayloadPkg/FirmwareUpdate/FirmwareUpdateHelper.c
```diff
@@ -585,6 +723,42 @@ UpdateSystemFirmware (
...
+    Status = PreserveFlashRegion (FLASH_MAP_SIG_UEFIVARIABLE, ImageHdr);
+    if (EFI_ERROR (Status) && Status != EFI_NOT_FOUND) {
+      DEBUG ((DEBUG_ERROR, "Failed to preserve UEFI variable: %r\n", Status));
+    }
+  }
+
+  if ((CapsuleFlags & CAPSULE_FLAG_SKIP_UPDATE_SMBIOS) != 0) {
+    Status = PreserveContainerComponent (SIGNATURE_32 ('I', 'P', 'F', 'W'), SIGNATURE_32 ('S', 'M', 'B', 'S'), ImageHdr);
+    if (EFI_ERROR (Status) && Status != EFI_NOT_FOUND) {
+      DEBUG ((DEBUG_ERROR, "Failed to preserve SMBIOS: %r\n", Status));
```
spkamat, line 758: If any of the above component fails, will that component get replaced with the updated component?
Also, we are ignoring EFI_NOT_FOUND error. FlashMapPtr being NULL also gets ignored in this case.

gdong1: If there is error, the component wouldn't be preserved from the code.
if EFI_NOT_FOUND is returned, it is not our preserve function error, we don't need print any error message here, regardless it is caused by FlashMap is not found or the component is not found.

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
```diff
@@ -405,7 +405,11 @@ def DumpInfo(self):
 
 class FirmwareUpdateHeader(Structure):
 
-  CAPSULE_FLAG_FORCE_BIOS_UPDATE = 0x80000000
+  CAPSULE_FLAG_FORCE_BIOS_UPDATE     = 0x80000000
```
lsubash, line 408: MRC, SBL VAR, UEFI var, SMbios should be skipped always by default. 
There should be one flag where only forced complete bios update can be triggered.
I believe user should not have option to select components to be skipped.

gdong1: I think we should not skipping anything to keep backward compatibility. Skipping them by default would silently break existing capsule update flow.   The tool already has the check and raises "Skip update flags are not compatible with force BIOS update" for the case skipping region in force bios update.

lsubash: I understand skipping the components are chosen by capsule update standard.
Agree for default for "not to skip", for backward compatibility. 
But, if skip is selected, i think we need skip all. Giving user more options would add usage complexity.

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
kokweich: resolved in latest commit

BootloaderCorePkg/Tools/GenCapsuleFirmware.py
kokweich: Added a skip all argument to sip all these non volatile region

gdong1 (approved):

PR #2678, opened 2026-04-16
Enable fspupd config

gdong1:


BootloaderCommonPkg/Library/ConfigDataLib/ConfigDataLib.c
```diff
@@ -424,3 +435,117 @@ BuildConfigData (
...
+  //
+  // Payload starts after the CDATA_HEADER and condition words.
+  // Payload is packed delta entries directly (no blob header).
+  //
+  DeltaData   = (VOID *)((UINT8 *)CdataHdr + sizeof (CDATA_HEADER) + sizeof (CDATA_COND) * CdataHdr->ConditionNum);
+  PayloadSize = CdataHdr->Length * 4 - sizeof (CDATA_HEADER) - sizeof (CDATA_COND) * CdataHdr->ConditionNum;
+
+  DEBUG ((DEBUG_INFO, "FSPUPD delta: PayloadSize=%d UpdSize=0x%X\n", PayloadSize, UpdSize));
+
+  DEBUG ((DEBUG_INFO, "FSPUPD delta: UPD BEFORE patch (first 16B):\n"));
```
bejeanmo, line 502: The before and after 16 byte dumps may be excessive. Consider changing to DEBUG_VERBOSE.

bejeanmo (approved):
See suggestion about ConfigDataLib.c verbosity

BootloaderCommonPkg/Library/ConfigDataLib/ConfigDataLib.c
gdong1: This debug level should be aligned with the DumpHex and the DumpHex function only supports DEBUG_INFO level. Let me remove this unnecessary debug message.

BootloaderCommonPkg/Library/ConfigDataLib/ConfigDataLib.c
```diff
@@ -424,3 +435,111 @@ BuildConfigData (
...
+    }
+
+    DeltaPtr += sizeof (FSPUPD_DELTA_ENTRY);
+
+    if (DeltaPtr + Length > DeltaEnd) {
+      DEBUG ((DEBUG_WARN, "FSPUPD delta: Entry truncated at payload boundary: off=0x%04X len=%d\n", Offset, Length));
+      break;
+    }
+
+    if ((UINT32)(Offset + Length) <= UpdSize) {
```
tsaikevin: @gdong1 suggest to case UINT32 first prior to add, it prevents overflow potentially. fyi
if (((UINT32)Offset + Length) <= UpdSize) {

gdong1: Thanks for the comments. Offset is UINT16 (max 65535), Length is UINT8 (max 255)
In C, both are promoted to int before the addition, and int is 32 bits, so max sum 65790 fits fine — no overflow. But the suggested change is more readable. let me change it.

tsaikevin (approved):

PR #2680, opened 2026-04-17
Firmware Update: Add defensive bounds-checking hardening

vtcao297:
- Enforce exclusive-upper-bound semantics in QEMU SPI flash operations to prevent off-by-one boundary violations during firmware reads and writes

- Add capsule image size validation for debug output to prevent reading beyond allocated buffer boundaries

- Add various verification and checking of capsule metadata

These changes improve the robustness and security of firmware update operations and ensure safer debug output handling during capsule image ingestion.

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -854,14 +854,14 @@ ValidateCapsuleLayout (
     return EFI_INVALID_PARAMETER;
   }
 
-  *FwUpdHeader = Header;
-
   if (CapHeader != NULL) {
     if (Header->ImageSize < sizeof (EFI_FW_MGMT_CAP_HEADER)) {
       return EFI_INVALID_PARAMETER;
     }
     *CapHeader = (EFI_FW_MGMT_CAP_HEADER *)((UINTN)Header + Header->ImageOffset);
   }
+
+  *FwUpdHeader = Header;
 
   return EFI_SUCCESS;
 }
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -1,7 +1,7 @@
 /** @file
 This driver is to update firmware in boot media.
 
-Copyright (c) 2017 - 2023, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2017 - 2026, Intel Corporation. All rights reserved.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -115,6 +115,17 @@ VerifySblVersion (
     // EFI_UNSUPPORTED from PlatformGetStage1AOffset and getting stage1Abase
     // will be handled in common way using the below implementation.
     if (Status == EFI_UNSUPPORTED) {
+      //
+      // The last 4 bytes of the BIOS region payload hold the Stage 1A FV base
+      // address. Guard against integer underflow: UpdateImageSize comes from the
+      // untrusted capsule image and if it were less than 4 the subtraction below
+      // would wrap to a very large address and pass an arbitrary pointer to GetSvn.
+      //
+      if (ImageHdr->UpdateImageSize < 4) {
+        DEBUG((DEBUG_ERROR, "VerifySblVersion: UpdateImageSize (%d) too small to contain FV base\n",
+               ImageHdr->UpdateImageSize));
+        return EFI_INVALID_PARAMETER;
+      }
       // Last 4 bytes of the BIOS region contain Stage 1A FV base.
       Stage1AFvBase = (UINT32)((UINTN)ImageHdr + sizeof(EFI_FW_MGMT_CAP_IMAGE_HEADER) + ImageHdr->UpdateImageSize - 4);
       Status = GetSvn (Stage1AFvBase, &CapsuleBlVersion);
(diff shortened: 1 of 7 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -888,6 +1078,18 @@ GetPayloadHeaderByIndex (
     return EFI_NOT_FOUND;
   }
 
+  //
+  // IsValidPayloadBoundary already confirmed Offset <= FwUpdHeader->ImageSize
+  // (a UINT32), so the value is bounded within 32 bits. This explicit guard
+  // is defense-in-depth for 32-bit (IA32) builds where casting a UINT64 to
+  // UINTN silently truncates any stray high-order bits, which could produce
+  // a pointer that lands outside the capsule buffer.
```
bejeanmo, line 1086: I dont follow. Can you explain this edge case?

vtcao297: the casting truncating concern line 1093 (right below this check) on 32bit builds, so assume CapHeader = 0x20000000, and assume corrupted Offset = 0x00000001_00000040 (UINT64).  On IA32 builds, (UINTN)Offset would be truncated to 0x00000040.  Line 1093 would then computes the OutImageHeader pointer as: 0x20000000 + 0x40 = 0x20000040.  So then OutImgHeader points to 0x20000040, which is not to the intended 64-bit offset location.  Without this guard, an attacker-controlled high bits could potentially “alias” a large 64-bit offset to a small in-bounds-looking 32-bit offset after cast.  The chance of this happening is low from a crafted capsule alone...thats why I consider it more defense-in-depth check.

vtcao297:
all the windows builds are failing in this commit, doesn't seem to be related to this commit code changes (at least from what I could tell).  Anyone knows what's happening?

bejeanmo:
> all the windows builds are failing in this commit, doesn't seem to be related to this commit code changes (at least from what I could tell). Anyone knows what's happening?

I think the microsoft compiler doesn't like this cast:
> (UINTN)~0ULL

I'm not sure why. You can probably use MAX_UINTN instead.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
bejeanmo: So this is to protect against the event of a separate exploit modifying Offset after line 1076?

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
```diff
@@ -210,16 +283,18 @@ GetCapsuleFromRawPartition (
...
              Buffer
              );
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_INFO, "Read capsule image error, Status = %r\n", Status));
+    FreePool (Buffer);
     return  Status;
   }
 
-  if ((Buffer == NULL) || (ImageSize == 0)) {
+  if (ImageSize == 0) {
```
gdong1: ImageSize  is already checked at line 236 before allocation.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -698,6 +744,128 @@ UpdateStatus (
...
+  }
+
+  *FwUpdHeader = Header;
+
+  if (CapHeader != NULL) {
+    if (Header->ImageSize < sizeof (EFI_FW_MGMT_CAP_HEADER)) {
+      return EFI_INVALID_PARAMETER;
+    }
+    *CapHeader = (EFI_FW_MGMT_CAP_HEADER *)((UINTN)Header + Header->ImageOffset);
+  }
```
gdong1, line 862: move line 857 here  to set output only after ALL checks pass incase it return at line 861.

gdong1:
The CI build failure was fixed by https://github.com/slimbootloader/slimbootloader/pull/2679.
Agree with Bejean, (UINTN)~0ULL should be changed to MAX_UINTN in this PR.

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
vtcao297: good catch, yes this would be dead code.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
vtcao297: No not necessary, the check is only for proving the condition at that moment prior to line 1093 due to the UINT64 to UINTN truncation concerns.  Think of it as more of a defense-in-depth/assertion-style check, not the primary security boundary.

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
vtcao297: fixed.

PayloadPkg/FirmwareUpdate/GetCapsuleImage.c
vtcao297: fixed

bejeanmo:
@vtcao297 this latest windows build failure is because of an IASL version change in the CI build. It should be fixed now.

vtcao297:
> @vtcao297 this latest windows build failure is because of an IASL version change in the CI build. It should be fixed now.

thanks, i'll rebase the branch to fix up the fix.

bejeanmo (approved):

gdong1 (approved):

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -698,6 +744,128 @@ UpdateStatus (
...
+
+  //
+  // Whitelist CapsuleFlags against the two defined flags:
+  //   CAPSULE_FLAGS_CFG_DATA      (BIT0)  - config data update flag
+  //   CAPSULE_FLAG_FORCE_BIOS_UPDATE (BIT31) - bypasses FWU state machine
+  // Any bit outside this set is unrecognized and could trigger unintended
+  // update paths. Rejecting undefined flags here prevents a crafted capsule
+  // from reaching later code with an unexpected flags value.
+  //
+  if ((Header->CapsuleFlags & ~(CAPSULE_FLAGS_CFG_DATA | CAPSULE_FLAG_FORCE_BIOS_UPDATE)) != 0) {
```
vtcao297, line 790: will also need to whitelist CAPSULE_FLAG_SKIP_UPDATE_MRC_DATA, CAPSULE_FLAG_SKIP_UPDATE_SBL_VAR, CAPSULE_FLAG_SKIP_UPDATE_UEFI_VAR, CAPSULE_FLAG_SKIP_UPDATE_SMBIOS based on #2672 if that gets merged also.

PR #2683, opened 2026-04-21
fix: [RPL-P] Simplified Post Mem Fusa Code flow

bejeanmo:
Simplified Post Mem Fusa Config Lib code flow and added some additional return status checks.

### DIFF
--- a/Silicon/RaptorlakePkg/Library/FusaConfigLib/FusaConfigLib.c
+++ b/Silicon/RaptorlakePkg/Library/FusaConfigLib/FusaConfigLib.c
@@ -219,80 +219,103 @@ FusaConfigPostMem (
 
   FusaCfgData =(FUSA_CFG_DATA *) FindConfigDataByTag (CDATA_FUSA_TAG);
 
-  if (FusaCfgData != NULL) {
-    FspsConfig->FusaConfigEnable = 1;
-    FspsConfig->DisplayFusaConfigEnable = FusaCfgData->DisplayFusaConfigEnable;
-    FspsConfig->GraphicFusaConfigEnable = FusaCfgData->GraphicFusaConfigEnable;
-    FspsConfig->OpioFusaConfigEnable = FusaCfgData->OpioFusaConfigEnable;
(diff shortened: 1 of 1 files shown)


Silicon/RaptorlakePkg/Library/FusaConfigLib/FusaConfigLib.c
```diff
@@ -219,80 +219,102 @@ FusaConfigPostMem (
...
+
+  DEBUG((DEBUG_VERBOSE, "Allocate successful. Reading from SPI\n"));
+  Status = SpiFlashRead(
+      FlashRegionBios,
+      FusaComp,
+      FusaCompLength,
+      FusaCompInMem
+      );
+  DEBUG((DEBUG_VERBOSE, "Copy FuSa FW Container to Mem Status: %r\n", Status));
+  DEBUG((DEBUG_VERBOSE, "Signature: 0x%X\n",((CONTAINER_HDR*)FusaCompInMem)->Signature));
```
vtcao297: this DEBUG prints probably should go after the Status check conditional completes?  If SpiflashRead fails likely FusaCompInMem->Signature contains garbage data.

Silicon/RaptorlakePkg/Library/FusaConfigLib/FusaConfigLib.c
vtcao297, line 214: Not familiar with FUSA flow but according to the message here, shouldn't we return here immediately rather than continue?

bejeanmo: This is fine, although i can see how this message is confusing. There is some somewhat convoluted logic for checking if we're in the FuSa diagnostics flow to cover some specific edge cases the FuSa engineers were worried about. The FuSa Info HOB signals that we're in the diagnostics flow, but even when not in a diagnostics boot FusaConfigPostMem() should continue.

vtcao297: okay gotcha

vtcao297 (approved):

tsaikevin (approved):

PR #2685, opened 2026-04-23
fix: Reassign WDT scratchpad TOP_SWAP_TRIGGER from BIT21 to BIT17

gdong1:
BIT21~23 of OC_WDT_CTL (ACPI IO 0x54) may be used by FSP internally, causing IsTopSwapTriggered() to return a false positive when FSP sets BIT21 for its own purpose. This leads to an unexpected recovery flow.

Move TOP_SWAP_TRIGGER from BIT21 to BIT17 and reorder the scratchpad bit definitions for clarity.


lsubash (approved):
Could we align with FSP and fix this bit for each platform?

gdong1:
> Could we align with FSP and fix this bit for each platform?
As the code change shown, for the 8 bits scratch pad, FSP used it from high bit (BIT23, BIT22, BIT21), and SBL uses the remaining 5 bits from BIT16. BIT17 was previous used by DSO, since DSO is deprecated, we could re-purpose it to topswap. I think SBL should use other method (e.g.: SBL variable) to avoid potential conflicts with FSP.

PR #2687, opened 2026-04-27
Container: Improve container component loading with bounds validation

vtcao297:
Improvements to the container library component loading and decompression pathways:

- Add bounded traversal functions for container header entry parsing with explicit checks

- Implement explicit container bounds validation at registration time to reject malformed or incomplete container regions before processing

- Add runtime input validation in decompressor metadata parsing to enforce minimum buffer sizes and NULL pointer checks in decompression GetInfo APIs

- Improve flow robustness by validating component metadata before decompressor interaction

- Refactor some extremely nested blocks for better readability

These changes strengthen boot-time security by enforcing defense-in-depth precondition validation across container parsing, component verification, and decompression initialization stages.

### DIFF
--- a/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
+++ b/BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
@@ -23,6 +23,37 @@
 
 #define  IS_FLASH_ADDRESS(x)   (((UINT32)(UINTN)(x)) >= 0xF0000000)
 
+STATIC
+BOOLEAN
+GetNextComponentEntryBounded (
+  IN  CONTAINER_HDR     *ContainerHdr,
+  IN  COMPONENT_ENTRY   *CurrEntry,
+  OUT COMPONENT_ENTRY   **NextEntry
+  );
+
+STATIC
+BOOLEAN
+IsComponentEntryReadable (
+  IN  CONTAINER_HDR     *ContainerHdr,
+  IN  COMPONENT_ENTRY   *CurrEntry
+  );
+
+STATIC
+BOOLEAN
+GetComponentDataPointer (
+  IN  CONTAINER_ENTRY   *ContainerEntry,
+  IN  CONTAINER_HDR     *ContainerHdr,
+  IN  COMPONENT_ENTRY   *CompEntry,
+  OUT UINT8             **CompData
+  );
+
+STATIC
+BOOLEAN
+ValidateContainerBounds (
+  IN  CONTAINER_HDR     *ContainerHdr,
+  IN  UINT32             ContainerSize
+  );
+
 /**
   Get the container pointer by the container signature
 
@@ -80,7 +111,10 @@ GetContainerHeaderSize (
       return 0;
     }
     for (Index = 0; Index < ContainerHdr->Count; Index++) {
-      CompEntry = (COMPONENT_ENTRY *)((UINT8 *)(CompEntry + 1) + CompEntry->HashSize);
+      if (!GetNextComponentEntryBounded (ContainerHdr, CompEntry, &CompEntry)) {
+        Offset = 0;
+        break;
+      }
       Offset = (UINT8 *)CompEntry - (UINT8 *)ContainerHdr;
       if ((Offset < 0) || (Offset > ContainerHdr->DataOffset)) {
         Offset = 0;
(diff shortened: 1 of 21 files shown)


bejeanmo (approved):

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -92,6 +126,132 @@ GetContainerHeaderSize (
...
+  if (!IsComponentEntryReadable (ContainerHdr, CurrEntry)) {
+    return FALSE;
+  }
+
+  HeaderLimit = ContainerHdr->DataOffset;
+  CurrOffset  = (UINT64)((UINT8 *)CurrEntry - (UINT8 *)ContainerHdr);
+
+  // Check for overflow before addition
+  if ((CurrEntry->HashSize > (MAX_UINT64 - sizeof(COMPONENT_ENTRY) - CurrOffset)) ||
+      ((CurrOffset + sizeof(COMPONENT_ENTRY) + CurrEntry->HashSize) > HeaderLimit)) {
```
tsaikevin: @vtcao297 line 164 and 165 are duplicated, the MAX value of HeaderLimit is MAX_UINT64 too. so line 165 is sufficient. fyi
btw, line 170 checking is equivalent(potentially dead code)

vtcao297: you're right, the first guard condition is kind of impossible since the worst-case sum is only ~ 2^32 + 267 which is much less 2^64 so it can never be true, over-defensive coding :).  And yes 170 would then also be redundant as 165, so i'll just combine them.  Thanks!

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -92,6 +126,126 @@ GetContainerHeaderSize (
...
+  UINT64  HeaderLimit;
+  UINT64  CurrOffset;
+
+  if ((ContainerHdr == NULL) || (CurrEntry == NULL)) {
+    return FALSE;
+  }
+
+  HeaderLimit = ContainerHdr->DataOffset;
+  CurrOffset  = (UINT64)((UINT8 *)CurrEntry - (UINT8 *)ContainerHdr);
+  if (CurrOffset > HeaderLimit) {
```
tsaikevin, line 198: @vtcao297 do we consider checking the "==" condition where no space left at CurrOffset?thx

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -824,29 +1226,50 @@ LoadComponentWithCallback (
...
+    return EFI_UNSUPPORTED;
   }
+
+  // Handle zero-size components (uncompressed, no decompression needed)
+  if (CompressHdr->Size == 0) {
+    DstLen = 0;
+    ScrLen = 0;
+  }
+
+  Status = DecompressGetInfo (CompressHdr->Signature, CompressHdr->Data,
```
tsaikevin: @vtcao297 when line 1254 is true, DstLen and ScrLen are set 0, is it still necessary to call DecompressGetInfo in line 1259? thx

tsaikevin:
@vtcao297 since many changes merged in container library, it actually introduces some chance to break(Copilot suggests some but I'm unable to verify:( ) Suggest to run static analysis again after code merged. fyi

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
```diff
@@ -80,7 +111,10 @@ GetContainerHeaderSize (
       return 0;
     }
     for (Index = 0; Index < ContainerHdr->Count; Index++) {
-      CompEntry = (COMPONENT_ENTRY *)((UINT8 *)(CompEntry + 1) + CompEntry->HashSize);
+      if (!GetNextComponentEntryBounded (ContainerHdr, CompEntry, &CompEntry)) {
```
lsubash, line 114: Looks CompEntry is repeated param. Its not passing next param.

vtcao297: In the old logic at ContainerLib.c, the code effectively did this:

-Read CompEntry->HashSize from the current entry.
-Compute the next pointer with raw pointer arithmetic.
-Only after that, compare the resulting offset against ContainerHdr->DataOffset.

That means it could still read fields from a partially out-of-bounds current entry before discovering the next pointer was invalid.

The replacement via GetNextComponentEntryBounded keeps the same core traversal rule:

next entry = current entry + sizeof(COMPONENT_ENTRY) + HashSize

but adds two security properties before returning the next pointer:

- IsComponentEntryReadable ensures the fixed-size current entry header is fully readable inside the container header bounds before reading HashSize.
- GetNextComponentEntryBounded ensures the computed next entry does not exceed DataOffset

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
vtcao297: I believe the == case is handled safely by the next block.

- if (CurrOffset > HeaderLimit) rejects true out-of-range.

- if ((HeaderLimit - CurrOffset) < sizeof (COMPONENT_ENTRY)) rejects “no room left” and partial room.

So when CurrOffset == HeaderLimit, remaining space is 0, which is < sizeof(COMPONENT_ENTRY), and it returns FALSE. Valid containers shouldn't be affected because a valid component entry must have at least sizeof(COMPONENT_ENTRY) bytes remaining.

BootloaderCommonPkg/Library/ContainerLib/ContainerLib.c
vtcao297: No I guess it doesn't, I was trying to flatten the previous codeblock with its nested if..elses.  Thanks for catching this, it didn't manifest in my testing because the worse case scenario was just over allocation of additional 64K in the scratch buffer AllocBuf if the compression algo was LZMA.  Will fix this soon

vtcao297: addressed in the latest revision.

vtcao297:
> @vtcao297 since many changes merged in container library, it actually introduces some chance to break(Copilot suggests some but I'm unable to verify:( ) Suggest to run static analysis again after code merged. fyi

Kevin - I didn't realize static analysis is being run against PRs, can you point me to the link result so I can review if the items being caught are real and need to be addressed before merge.

bejeanmo:
> > @vtcao297 since many changes merged in container library, it actually introduces some chance to break(Copilot suggests some but I'm unable to verify:( ) Suggest to run static analysis again after code merged. fyi
> 
> Kevin - I didn't realize static analysis is being run against PRs, can you point me to the link result so I can review if the items being caught are real and need to be addressed before merge.

It isn't. I think kevin is suggesting to check after its merged.

gdong1 (commented):
Hi @vtcao297, thanks for the security hardening work here.

I think this change works perfectly fine, but the per-access bounded checks modify a significant amount of existing code and increase binary size.

How about an alternative approach: validate-once at registration time? Since SBL runs single-threaded and container memory can't be mutated between registration and use, a single structural validation pass at RegisterContainer() should be sufficient. This keeps all existing traversal code untouched.

I put together a quick implementation for reference:
[https://github.com/slimbootloader/slimbootloader/compare/master...gdong1:slimbootloader:validate-container-bounds]

Feel free to take a look and let me know if you'd like to adopt this approach or incorporate parts of it.

PR #2692, opened 2026-05-06
Update TopSwapLib and define RTC PID and P2SB variables

tsaikevin:
Refactor TopSwapLib to use P2sbLib helpers and define PID_RTC_HOST_NUM and P2SB_BUS_NUM Build flags
Remove fixed P2SB memory base and Port ID. Platforms need override both variables to access Boot Partition
for Firmware update boot flow.


gdong1 (approved):

gdong1:
do we need change MTL topswap to use common one?

tsaikevin:
> do we need change MTL topswap to use common one?

@gdong1 will submit another PR to merge MTL code. fyi

PR #2694, opened 2026-05-12
fix: Some firmware updates fail when sbl size exceeds 16MB

bejeanmo:
With certain flash layouts when sbl size exceeds 16MB, the FW update status in the SBL reserved region can overflow the memory mapped portion of SPI flash and needs to be read through SPI controller access.


lsubash (approved):

spkamat:
Is SBL image size greater than 16MB (BIOS region size) a valid scenario? Shouldn't we be blocking that?

bejeanmo:
> Is SBL image size greater than 16MB (BIOS region size) a valid scenario? Shouldn't we be blocking that?

It's a valid scenario. This is the case for FuSa bootloader with very large BIST data built in. It requires > 32MB SPI flash though.

PR #2695, opened 2026-05-14
Move PsdLib to CommonSocPkg

tsaikevin:
Refactor PSD library into a shared CommonSocPkg PsdLib implementation and header. Remove board PSD ASL entries and hook Stage2BoardInitLib paths to install/update PSD data through the common library for ADL and ARL flows; update board configs/INFs to consume the library

Populate PSD fields from platform state, preferring MBP HOB data for FW version and falling back to HECI when needed, while collecting EOM, security capability etc.


bejeanmo (changes requested):
Can you separate the common code PsdLib and platform porting into two separate commits? 

Ex:
Commit 1: Add Psdlib to CommonSocPkg
Commit 2: Move ADL/RPL/BTL/TWL/ARL to use common PsdLib

I think this makes it clearer when looking at the code history. Otherwise actual code looks good.

lsubash (commented):
Kevin, Does the MbpHob population is common?
PSD is heavily dependent on MpHob.

tsaikevin:
> Kevin, Does the MbpHob population is common? PSD is heavily dependent on MpHob.
@lsubash yes, PSD needs FwVersionName and FwCapsSku in MpHob data. Both are common for all platform. fyi

pchand20:
before moving this library or doing any changes . it will be good to have discussion with security team and owners of this lib on overall motivation and design.

tsaikevin:
> Can you separate the common code PsdLib and platform porting into two separate commits?
> 
> Ex: Commit 1: Add Psdlib to CommonSocPkg Commit 2: Move ADL/RPL/BTL/TWL/ARL to use common PsdLib
> 
> I think this makes it clearer when looking at the code history. Otherwise actual code looks good.

@bejeanmo sure, done. fyi

tsaikevin:
> before moving this library or doing any changes . it will be good to have discussion with security team and owners of this lib on overall motivation and design.

@pchand20 sure, will discuss in SBL mtg.

Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -984,6 +986,51 @@ MtlPchGetLpssUartFixedPciCfgOffset (
...
+PsdTableUpdate (
+  IN OUT EFI_ACPI_DESCRIPTION_HEADER  *AcpiTable
+  )
+{
+  EFI_STATUS                          Status;
+  EFI_ACPI_PSD_TABLE                  *PsdTable;
+  PLATFORM_DATA                       *PlatformData;
+
+  if (AcpiTable == NULL) {
+    return;
```
pchand20, line 1009: we suppose to return some meaningful error code. at least we were dong that in original code. why we not doing that here.

Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -984,6 +986,51 @@ MtlPchGetLpssUartFixedPciCfgOffset (
...
+  }
+
+  // Initialize the table
+  PsdTable = (EFI_ACPI_PSD_TABLE *)AcpiTable;
+  Status = UpdateAcpiPsdTable (PsdTable);
+  if (EFI_ERROR (Status)) {
+    return;
```
pchand20, line 1017: same as above.

Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -984,6 +986,51 @@ MtlPchGetLpssUartFixedPciCfgOffset (
...
+  Update Platform Service Discovery Table.
+
+  This serves as the main entry point for building the PSD table at runtime.
+  It initializes the table based on platform configuration.
+
+  @param[in, out] AcpiTable   Pointer to the ACPI table buffer.
+
+**/
+VOID
+PsdTableUpdate (
```
pchand20, line 999: seems this function is new. there is no need to add new APIs. the original code is optimized to reduce less function call to avoid stack uses.

Silicon/CommonSocPkg/Library/PsdLib/PsdLib.c
```diff
@@ -62,16 +75,16 @@ GetMeBiosPayloadHobData (
   @param[in,out] EomState         Pointer to The EOM state value.
 
   @retval EFI_SUCCESS             The EOM state was retrieved successfully.
-  @retval Others                  An error occurred.
+  @retval Others                  An error occured.
```
pchand20: please correct "occurred"

Silicon/CommonSocPkg/Library/PsdLib/PsdLib.c
```diff
@@ -177,84 +190,63 @@ GetSecCapability (
...
-  Status = GetSecFwVersion (&mPsdt->FwVer);
+  Status = GetSecFwVersion (&Psdt->FwVer);
   if (EFI_ERROR(Status)) {
     DEBUG((DEBUG_ERROR, " GetSecFwVersion failed =%x\n", Status));
+    return Status;
   }
 
-  CopyMem (&mPsdt->FwVendor, EFI_ACPI_PSD_FW_VENDOR, EFI_ACPI_PSD_FW_VENDOR_SIZE);
-
-  PlatformData = (PLATFORM_DATA *)GetPlatformDataPtr ();
```
pchand20, line 240: let not change the way original code was. it will cause confusion. also why we initializing boot guard status in different function ?   all the update shall be done here and not at other places.

Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
```diff
@@ -984,6 +986,51 @@ MtlPchGetLpssUartFixedPciCfgOffset (
...
+
+  PlatformData = (PLATFORM_DATA *)GetPlatformDataPtr ();
+  if (PlatformData == NULL) {
+    DEBUG(( DEBUG_ERROR, "GetPlatformDataPtr Failed\n"));
+    return;
+  }
+
+  // Check if verified boot is reflected in BtG profile as well as in SBL
+  // BIT0: UEFI secure boot is enabled, BIT1: Boot Guard is enabled, BIT2: Bootloader verified boot is enabled
+  PsdTable->SecureBoot = (UINT8)(((PlatformData->BtGuardInfo.VerifiedBoot) << 1) | (FeaturePcdGet (PcdVerifiedBootEnabled)) << 2);
```
pchand20, line 1028: the update is dont later in update function. why we dong it here ? lets not change code that way is was.

pchand20 (changes requested):

Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
tsaikevin: @pchand20 the PlatformData is not available in common code. It only works in platform dependent code(inf file removes it). hence move it to platform library. fyi

Silicon/CommonSocPkg/Library/PsdLib/PsdLib.c
tsaikevin: @pchand20 the PLATFROM_DATA definition is located in platform code at Silicon/ArrowlakePkg/Include. Common lib cannot locate this header. fyi

gdong1 (approved):

Platform/ArrowlakeBoardPkg/Library/Stage2BoardInitLib/Stage2BoardInitLib.c
tsaikevin: @pchand20 it's necessary to have platform data update the PSD table. It also aligns the structure same as MCFG, MADT tables. fyi

pchand20: Adding  a new function in AcpiPlatform.c vs Calling function here and passing data for PSDS when its enabled only. that was the previous approach.  since PSDS is config dependent it may get enabled disabled at build time as well. so we shall not have any code compiled for that fact of mater. if you add new function than make sure compile time include exclude is added if required.  also by design we only update PSDS values at one place just to keep it clean.

tsaikevin: @pchand20 the code does work as you describe after verified the final objects in map file.
The linker removes PsdableUpdate routine from object file when PSD is disabled from PCD variable.
The other option is changing PCD type to PcdsFixedAtBuild, which I feel the code is more readiable similar to PcdTccEnabled

pchand20 (approved):

bejeanmo (approved):

PR #2698, opened 2026-05-18
fix: [ADL/RPL/BTL] Add NULL check before TXT feature config dereference

bejeanmo:
Static analysis flagged dereferencing FeaturesCfgData to check for TXT support without checking for NULL pointer.


gdong1 (approved):

lsubash (approved):

lsubash:
Pls check same will be applicable for PTL also.

PR #2700, opened 2026-05-20
feat: [RPLP] enabled POST code for CRB

vincentchchen:
The 7-segment LED on RPL-P CRB is controlled by the MAX6958 through I2C Controller #0. This patch adds the corresponding initialization code.

### DIFF
--- a/Platform/AlderlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
@@ -770,6 +770,12 @@ DEBUG_CODE_END();
       (UINT8) ~((B_PMC_PWRM_GEN_PMCON_A_MS4V | B_PMC_PWRM_GEN_PMCON_A_SUS_PWR_FLR) >> 16),
       B_PMC_PWRM_GEN_PMCON_A_DISB >> 16
     );
+
+    MemCfgData = (MEMORY_CFG_DATA *)FindConfigDataByTag (CDATA_MEMORY_TAG);
+    if (MemCfgData == NULL) {
+      CpuHalt ("Failed to find memory CFGDATA!");
+    }
+
     switch (GetPlatformId ()) {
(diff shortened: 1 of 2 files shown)


Platform/AlderlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
```diff
@@ -785,6 +785,16 @@ DEBUG_CODE_END();
     case PLATFORM_ID_ADL_P_LP5_RVP:
       ConfigureGpio (CDATA_NO_TAG, sizeof (mGpioTablePreMemAdlPLp5Rvp) / sizeof (mGpioTablePreMemAdlPLp5Rvp[0]), (UINT8*)mGpioTablePreMemAdlPLp5Rvp);
       break;
+    case PLATFORM_ID_RPL_P_DDR5_CRB:
+      MemCfgData = (MEMORY_CFG_DATA *)FindConfigDataByTag (CDATA_MEMORY_TAG);
```
spkamat: Is this the first instance in the boot flow where MemCfg is read?

vincentchchen: Yes, but there are two instances in the switch-case clause after this PR.
Do you suggest moving the read outside the switch-case?

tsaikevin (approved):

Platform/AlderlakeBoardPkg/Library/Stage1BBoardInitLib/Stage1BBoardInitLib.c
spkamat: that would be ideal to avoid duplicating.. also if there is an error and we are halting boot, doesn't make sense to check again later in the same flow.

vincentchchen: updated

spkamat (approved):

gdong1 (approved):

PR #2706, opened 2026-06-01
feat: Support building Slim Bootloader with LLVM/CLANG

bejeanmo:
Add support for building SBL with LLVM/CLANG toolchain. Pass "-t clang"
to the build command. Windows still requires Visual Studio for nmake and
Building Basetools.

- Added Windows/Linux CLANG builds to CI
- Undefine _WIN32/_WIN64 in IppCrypto GCC family builds. Windows CLANG
  sets these and it confuses IppCrypto into thinking it's building for
  a windows target.
- Add CLANG .map file support to PatchFv.py
- Fix some minor casting errors
- Update BuildLoader.py to still use VS toolchain for windows BaseTools


BootloaderCorePkg/Tools/PatchFv.py
```diff
@@ -447,7 +457,8 @@ def parseModMapFile(self, moduleName, mapFile):
         for reportLine in reportLines:
             match = re.match(patchMapFileMatchString, reportLine)
             if match is not None:
-                if prefix == '' and len(match.group(matchSymbolGroupIndex)) > 10:
+                if prefix == '' and (len(match.group(matchSymbolGroupIndex)) > 10 or
+                                     re.match(r"^.*CLANGPDB.*X64", match.group(matchKeyGroupIndex))):
                     prefix = '_'
```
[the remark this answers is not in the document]
bejeanmo: The purpose of this change is to catch x64 clang map files just as the len(match.group(matchSymbolGroupIndex)) > 10 check catches x64 GCC map files, both of which will drop the '_' prefix unlike the 32-bit versions.

BootloaderCommonPkg/Library/IppCrypto2Lib/IppCrypto2LibY8.inf
```diff
@@ -219,5 +219,5 @@
   MSFT:*_*_*_CC_FLAGS = -D_ABL_ -D_Y8 -DIPPCP_FIPS_MODE -D_ARCH_EM64T -D_NO_IPP_DEPRECATED /GL-
   MSFT:*_*_X64_NASM_FLAGS = -f win64 -D_ABL_ -D_Y8 -DWIN32E
   GCC:*_*_X64_NASM_FLAGS = -D_ABL_ -D_Y8 -DLINUX32E
-  GCC:*_*_*_CC_FLAGS  = -D_ABL_ -D_Y8  -DIPPCP_FIPS_MODE -DMBX_FIPS_MODE -D_ARCH_EM64T -D_NO_IPP_DEPRECATED -w -Wformat -Wformat-security -march=nehalem -msse4.2 -maes -mpclmul -msha -U__INTEL_COMPILER
+  GCC:*_*_*_CC_FLAGS  = -D_ABL_ -D_Y8  -DIPPCP_FIPS_MODE -DMBX_FIPS_MODE -D_ARCH_EM64T -D_NO_IPP_DEPRECATED -w -Wformat -Wformat-security -march=nehalem -msse4.2 -maes -mpclmul -msha -U__INTEL_COMPILER -U_WIN64 -U_WIN32
```
lsubash, line 223: The inf's in IppCrypto2Lib only supports X64 only. Applicable to all INF's.

BootloaderCorePkg/Tools/PatchFv.py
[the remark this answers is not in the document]
bejeanmo: About the symbol name confusion, this is just the way CLANG creates map files. Each .obj has an entry with the file path in the symbol name location.

BootloaderCommonPkg/Library/IppCrypto2Lib/IppCrypto2LibY8.inf
bejeanmo: It turns out -U_WIN32 is still needed for x64 builds otherwise IppCrypto thinks it's building for a windows application target.

.azurepipelines/azure-pipelines.yml
```diff
@@ -488,3 +488,352 @@ jobs:
...
+        Build.Target: ""
+      ADLN_X86_RELEASE:
+        Build.Name:   "adln"
+        Build.Arch:   ""
+        Build.Target: "-r"
+      ADLN_X64_DEBUG:
+        Build.Name:   "adln"
+        Build.Arch:   "-a x64"
+        Build.Target: ""
+      ADLN50_X86_RELEASE:
```
tsaikevin, line 568: @bejeanmo probably we don't need to build up2 board so frequent. fyi

tsaikevin (approved):

PR #2708, opened 2026-06-03
Add native UI

gdong1:


PayloadPkg/Library/UiSetupLib/UiSetup.c
```diff
@@ -0,0 +1,2132 @@
...
+            if (TopIdx >= ElemCount) {
+              TopIdx = (UINT16)(ElemCount - 1);
+            }
+            ElemIdx = TopIdx;
+          }
+        }
+        break;
+
+      case 'd':
+      case 'D':
```
tsaikevin: @gdong1 we might want to define another key to replace D, because when users change value to 0xd , the ui  take it as special key instead of config value. fyi

gdong1: Good catch. we should not use hex char, will change to V.

tsaikevin (approved):

BootloaderCommonPkg/Library/ShellLib/CmdUi.c
```diff
@@ -0,0 +1,113 @@
...
+  if (EFI_ERROR (Status)) {
+    ShellPrint (L"UiSetup failed: %r\n", Status);
+    return Status;
+  }
+
+  return Status;
+}
+
+CONST SHELL_COMMAND ShellCommandUi = {
+  L"ui",
```
spkamat: Either 'setup' (preferred) or 'uisetup' could be a better command name instead of just 'ui'.

BootloaderCommonPkg/Library/ShellLib/CmdUi.c
```diff
@@ -0,0 +1,113 @@
...
+  L"Launch UiSetup",
```
spkamat: "Launch UI based setup menu"

spkamat:
General feedback:
Some commits do not have any commit description.
Some sections of code are not guarded by PcdUiSetupEnabled. Please double check if it is intentional or missed out.

BootloaderCommonPkg/Library/ShellLib/CmdUi.c
gdong1: 'setup' is used by MicroPython setup payload.  'ui' is simple.

gdong1:
> General feedback: Some commits do not have any commit description. Some sections of code are not guarded by PcdUiSetupEnabled. Please double check if it is intentional or missed out.

could you point out which commit missing description?.
and point out which part of code need be guarded? I don't find code size increase when PcdUiSetupEnabled is disabled.

spkamat:
> > General feedback: Some commits do not have any commit description. Some sections of code are not guarded by PcdUiSetupEnabled. Please double check if it is intentional or missed out.
> 
> could you point out which commit missing description?. 
QEMU: Build-in UI setup by default
QEMU: Add config data runtime test

>and point out which part of code need be guarded? I don't find code size increase when PcdUiSetupEnabled is disabled.
Will add inline comment so that it is not missed out.

BootloaderCommonPkg/Library/ShellLib/CmdUi.c
spkamat: MicroPython is a different payload and optional (and with this PR will it still be needed?). The current shell commands are specific to OsLoader. As long as it is not duplicated in current payload, i think it should be fine to have commands with the names shared by other payloads. UefiPayload might also have some command names used in OsLoader. My point is "ui" is not a very intuitive name for a command as it does not say what it does..

BootloaderCommonPkg/Include/Guid/LoaderFspInfoGuid.h
```diff
@@ -19,6 +19,8 @@ typedef struct {
   UINT8          Reserved0[3];
   UINT32         FspsBase;
   UINT32         FspHobList;
+  UINT32         FspmUpdPtr;
+  UINT32         FspsUpdPtr;
```
spkamat, line 23: Guard needed here?

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
```diff
@@ -255,6 +257,40 @@ GetConfigDataSize (
...
+  patched into UpdPtr using the saved bit offset/length.
+
+  @param[in]      BaseTag    Base UiSetup tag ID for the FSP UPD region.
+  @param[in,out]  UpdPtr     Pointer to the live FSP UPD buffer to patch.
+  @param[in]      UpdSize    Size of the UPD buffer in bytes.
+
+**/
+VOID
+EFIAPI
+ApplyCfgDeltaToFspUpd (
```
spkamat, line 275: I was thinking of guards for these functions. However if they are getting compiled out due to no caller and evidenced from the binary size, then it may be fine. If this is compiler specific, then hope it works with GCC too.

BootloaderCommonPkg/Library/ShellLib/CmdUi.c
gdong1: Both 'ui' and 'setup' are Osloader shell commands, they could be enabled at same time. so we could not use same name.  refer BootloaderCommonPkg\Library\ShellLib\CmdSetup.c.  let me change to 'uisetup'.

BootloaderCommonPkg/Include/Guid/LoaderFspInfoGuid.h
gdong1: I think this is generic change for FSP INFO hob to adding FSP UPD info. The change is small and could be used for other payloads, not necessary to limit it to UISetup.

BootloaderCommonPkg/Include/Library/ConfigDataLib.h
gdong1: Yes, they will NOT be built-in when UI is disabled.

gdong1:
> > > General feedback: Some commits do not have any commit description. Some sections of code are not guarded by PcdUiSetupEnabled. Please double check if it is intentional or missed out.
> > 
> > 
> > could you point out which commit missing description?.
> > QEMU: Build-in UI setup by default
> > QEMU: Add config data runtime test
> 
> > and point out which part of code need be guarded? I don't find code size increase when PcdUiSetupEnabled is disabled.
> > Will add inline comment so that it is not missed out.

Thanks. Updated both commits.

BootloaderCommonPkg/Library/ShellLib/CmdUi.c
gdong1: updated.

spkamat (approved):

PR #2721, opened 2026-06-16
fix: Check for invalid container data pointer from raw partition

bejeanmo:
When reading container from a raw partition, make sure the container image data pointer is valid.

### DIFF
--- a/PayloadPkg/OsLoader/LoadImage.c
+++ b/PayloadPkg/OsLoader/LoadImage.c
@@ -220,6 +220,12 @@ GetBootImageFromRawPartition (
   //
   Address =  LogicBlkDev.StartBlock + LbaAddr + AlignedHeaderBlkCnt;
 
+  if (AlignedImageSize <= AlignedHeaderSize) {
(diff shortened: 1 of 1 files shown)


tsaikevin (approved):

vtcao297 (approved):

PayloadPkg/OsLoader/LoadImage.c
```diff
@@ -220,6 +220,12 @@ GetBootImageFromRawPartition (
   //
   Address =  LogicBlkDev.StartBlock + LbaAddr + AlignedHeaderBlkCnt;
 
+  if (AlignedImageSize <= AlignedHeaderSize) {
```
spkamat, line 223: Shouldn't this be only for the less than case?

vtcao297: if it's just < then we would be allowing AlignedImageSize == AlignedHeaderSize? That means the image is basically just the header only right, zero actual data bytes...which to me doesn't make sense since we want an actual image beyond just header metadata right?

bejeanmo: I'm also not convinced all the Media types ReadBlocks function will handle a call to read zero blocks correctly.

spkamat (approved):

gdong1 (approved):

PR #2737, opened 2026-07-01
[BuildLoader] Deprecated IA32 build for x64 platforms

aborwank:
Added Deprecation warning for IA32 build in BuildLoader.py and updated BoardConfig files for Arrowlake platforms to reflect the deprecation status.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -28,6 +28,14 @@ from   BuildUtility import *
 
 import glob
 
+
+def print_warning(message):
+    # Use ANSI yellow for warnings when output is a terminal.
+    if os.name != 'nt' and sys.stdout.isatty() and os.getenv('NO_COLOR') is None:
+        print('\033[33m%s\033[0m' % message)
+    else:
+        print(message)
+
 def rebuild_basetools ():
     exe_list = 'GenFfs  GenFv  GenFw  GenSec  Lz4Compress  LzmaCompress'.split()
     ret = 0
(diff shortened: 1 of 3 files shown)


bejeanmo:
Looks fine but please fix the spelling.

spkamat:
Please check the return type suggestion as well.

lsubash (commented):
We have to define the same for each boardconfig for ARL, PTL/WCL, NVL, ICX, GNR and so.
Instead, can we a generic list which defines all platforms at one place for which X64 is POR?

aborwank:
Removed the flag from individual boardconfig files and added a list of deprecated boards

BuildLoader.py
```diff
@@ -28,6 +28,25 @@
 
 import glob
 
+# Platforms that are X64-only and should block IA32 build requests.
+IA32_X64_ONLY_BOARDS = {
```
lsubash: Would you rename to X64_ONLY_BOARDS? You can add server platforms too here.

aborwank: done

BuildLoader.py
```diff
@@ -28,6 +28,26 @@
...
+# Platforms that are X64-only and should block IA32 build requests.
+X64_ONLY_BOARDS = {
+    'arlh',
+    'arls',
+    'arlu',
+    'nvlp',
+    'nvls',
+    'ptl',
+    'wcl',
+    'ksv',
```
lsubash: Recommend to remove NVL until its PV. Also add IDV platforms also.
All server are X64.

lsubash (approved):

gdong1 (approved):

lsubash:
Antara, there were few builds are failing. Would you rebase and for build check to trigger?

PR #2742, opened 2026-07-07
bugfix: Guards for edge cases

bejeanmo:
Add stricter parameter and bounds validation across HashStore iteration, HKDF Extract/Expand, and hash comparison routines. Use overflow-safe arithmetic throughout and align HKDF behaviour with RFC 5869.

### DIFF
--- a/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
+++ b/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
@@ -448,6 +448,31 @@ GetComponentInfo (
 }
 
 /**
+  Return the expected digest length in bytes for a given HashAlg identifier.
+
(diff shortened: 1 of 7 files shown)


vrpate2 (approved):

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
```diff
@@ -735,6 +774,43 @@ SetDeviceAddr (
   @retval RETURN_COMPROMISED_DATA    Hash data in store is invalid.
 
 **/
+INTN
+CompareMemConstantTime (
```
lsubash, line 780: Move this API to above. This is mixed with MatchHashInStore header

bejeanmo: Fixed

lsubash (approved):
Though we have Hmac supported, this is not included in any of the usecase in current platforms.

PR #2745, opened 2026-07-08
fix: [ShellLib] Print full 64-bit address in mm command

spkamat:
The address column in MmRead() was formatted with %08X, truncating addresses above 4GB (e.g. 0x100000000 displayed as 0x00000000). Change format specifier to %016llX to display the full UINTN width.

### DIFF
--- a/BootloaderCommonPkg/Library/ShellLib/CmdMm.c
+++ b/BootloaderCommonPkg/Library/ShellLib/CmdMm.c
@@ -125,7 +125,7 @@ MmRead (
   Count  += Start;
 
   if (Start > 0) {
-    ShellPrint (L"%08X: ", Addr & ~0xF);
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/ShellLib/CmdMm.c
```diff
@@ -125,15 +125,15 @@ MmRead (
   Count  += Start;
 
   if (Start > 0) {
-    ShellPrint (L"%08X: ", Addr & ~0xF);
+    ShellPrint (L"%0*llX: ", (int)(sizeof(UINTN)*2), (UINT64)(Addr & ~0xF));
```
gdong1, line 128: prefer using %lX instead %llX, %ll is not mentioned although it works in EDK2.

gdong1 (approved):

PR #2748, opened 2026-07-09
[ARL]Fix UefiPld assert error

samihahkasim:
Fix UefiPld assert error due to FSP locks Global SMI internally during FspSiliconInit, before returning control to SBL. This prevents SMI from working in the UEFI payload. Set PchLockDownGlobalSmi=0 so UEFI payload handles lockdown at the proper time.

Signed-off-by: samihahkasim <samihah.kasim@intel.com>


tsaikevin (approved):

gdong1 (approved):

gdong1:
UEFI payload could enable/disable SMM feature. so we should not set lock/unlock global SMI based on payload ID. we should have a flag to indicate payload supports SMM driver or not. Based on it we could lock/unlock global SMI. we could do it in a separate PR.

PR #2749, opened 2026-07-09
PayloadPkg/UiSetupLib: harden popup width math for Coverity

tsaikevin:
Address Coverity INTEGER_OVERFLOW/cast_overflow concerns around PopupWidth - 4 used as PutString MaxLen in RenderFindPopup.
Use a guarded subtract to derive ContentWidth only when PopupWidth > 4, otherwise keep ContentWidth at 0. This avoids underflow/wrap paths and removes risky cast patterns in the width calculation.


gdong1 (approved):

tsaikevin:
build failures are pointing to payload size issue. will fix in another PR. FYI

PR #2752, opened 2026-07-10
fix: Check multiboot tag buffer bounds when applying tags

bejeanmo:
Each time a new tag is appended to the multiboot header, check to make sure tag fits in remaining space.

### DIFF
--- a/BootloaderCommonPkg/Library/MultibootLib/Multiboot2.c
+++ b/BootloaderCommonPkg/Library/MultibootLib/Multiboot2.c
@@ -145,12 +145,14 @@ SetupMultiboot2Info (
   )
 {
   MULTIBOOT2_INFO *Mb2Info = &MultiBoot->Mb2Info;
+  UINT64          NextTagSize;
 
   unsigned mb2_info_size_max = 0x1000;
   void *mb2_info = AllocatePool (mb2_info_size_max);
   if (mb2_info == NULL) {
-    DEBUG ((DEBUG_INFO, "Multiboot-2 info buffer allocation Error\n"));
+    DEBUG ((DEBUG_ERROR, "Multiboot-2 info buffer allocation Error\n"));
     ASSERT (mb2_info);
+    return;
   }
 
   struct multiboot2_start_tag *mbi_start = (struct multiboot2_start_tag *)mb2_info;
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/MultibootLib/Multiboot2.c
```diff
@@ -172,6 +172,12 @@ SetupMultiboot2Info (
   if (MultiBoot->CmdFile.Size != 0)
   {
     DECLARE_TAG (struct multiboot2_tag_string, tag, mb2_info, info_idx);
+    if (mb2_info_size_max < info_idx + sizeof(struct multiboot2_tag_string) + MultiBoot->CmdFile.Size + 1) {
+      DEBUG ((DEBUG_INFO, "Multiboot-2 info buffer overflow\n"));
```
spkamat: Shouldn't these be DEBUG_ERROR instead of INFO?

bejeanmo: Yes. I'll make the change.

BootloaderCommonPkg/Library/MultibootLib/Multiboot2.c
```diff
@@ -145,11 +145,12 @@ SetupMultiboot2Info (
...
 {
   MULTIBOOT2_INFO *Mb2Info = &MultiBoot->Mb2Info;
+  UINT64          NextTagSize;
 
   unsigned mb2_info_size_max = 0x1000;
   void *mb2_info = AllocatePool (mb2_info_size_max);
   if (mb2_info == NULL) {
-    DEBUG ((DEBUG_INFO, "Multiboot-2 info buffer allocation Error\n"));
+    DEBUG ((DEBUG_ERROR, "Multiboot-2 info buffer allocation Error\n"));
     ASSERT (mb2_info);
```
spkamat, line 154: Missing return? In release mode Assert would be compiled out.

bejeanmo: Yes, this was part of the original code, not the added checks, but the return should be added.

spkamat (approved):

PR #2754, opened 2026-07-13
FirmwareUpdate: fix component flash address translation in UpdateSingleComponent

tsaikevin:
Update single-component write address calculation to use explicit ROM base translation instead of overflow-prone addition semantics.
1. Validate CompBase is not below RomBase
2. Translate absolute component address to BIOS-region offset using subtraction
3. Avoid false EFI_UNSUPPORTED from address-computation overflow checks on valid top-of-4G component addresses


bejeanmo (approved):

lsubash (approved):

gdong1 (approved):

tsaikevin:
Build error due to OS_LOADER size will be fixed in PR https://github.com/slimbootloader/slimbootloader/commit/6a97cae6eed38de81ecd85bd2ab07d225e4a6fa7

PR #2756, opened 2026-07-13
fix: [ADL/RPL] Deprioritize USB boot as the default/1st option

tsaikevin:
This commit depriotize USB boot as the 1st boot option that SBL tries to boot from in preference for onboard boot device options.


gdong1 (approved):

tsaikevin:
Build error due to OS_LOADER size will be fixed in PR https://github.com/slimbootloader/slimbootloader/pull/2755/changes/6a97cae6eed38de81ecd85bd2ab07d225e4a6fa7

PR #2776, opened 2026-08-05
Build Tcg2AcpiCommunicateBuffer HOB for UEFI payload

lsubash:
Tcg2Acpi driver in UEFI payload consumes this hob. Allocate ACPI NVS memory for TCG_NVS which is gets as op region.


lsubash:
Not required. Handled in UEFI payload TCG2ACPi.

PR #2783, opened 2026-08-12
[TXT] Added ACM Version 0x100 entry

aborwank:
TXT header file contains the function prototypes
for TXT library functions common for all platforms.

### DIFF
--- a/BuildLoader.py
+++ b/BuildLoader.py
@@ -484,6 +484,13 @@ class Build(object):
         # ACM
         if self._board.ACM_SIZE > 0:
             if self._board.ACM_FIT_VERISON == 0x200:
+                # Precede the versioned entries with a 0x100 one so ACMs that predate
+                # FIT version 0x200 can still locate the module.
+                fit_entry = FIT_ENTRY.from_buffer(rom, fit_offset + (num_fit_entries+1)*16)
+                fit_entry.set_values(self._board.ACM_BASE, 0, 0x100, 0x2, 0)
+                print ('  Patching entry %d with 0x%08X:0x%08X - ACM' % (num_fit_entries, fit_entry.address, fit_entry.size))
+                num_fit_entries += 1
+
                 # Add all the CPUIDs for ACM FIT Version 0x200
                 for CPU_FMS in self._board._ACM_CPU_FMS :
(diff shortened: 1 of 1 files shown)


BuildLoader.py
```diff
@@ -483,8 +483,13 @@ def update_fit_table (self):
 
         # ACM
         if self._board.ACM_SIZE > 0:
```
lsubash, line 485: What is the reason for introducing FIT version 0x100 along with 0x200 for newer platforms?

aborwank (commented):
> What is the reason for introducing FIT version 0x100 along with 0x200 for newer platforms? 

TXTBTGINFO tool expects the 0x100 entry to read ACM info. I have compared with BIOS it also has the same.

BuildLoader.py
```diff
@@ -483,6 +483,12 @@ def update_fit_table (self):
...
+            if getattr(self._board, 'ADD_LEGACY_ACM_FIT_ENTRY', 0):
```
lsubash: There is ACM 0x100 FIT entry addition below. If 0x100 is mandatory, lets add it for all platforms by default.
Also, what is the need for separate ADD_LEGACY_ACM_FIT_ENTRY ?

BuildLoader.py
```diff
@@ -484,6 +484,13 @@ def update_fit_table (self):
         # ACM
         if self._board.ACM_SIZE > 0:
```
lsubash, line 485: Update the commit details as per the changes.

BuildLoader.py
```diff
@@ -484,6 +484,13 @@ def update_fit_table (self):
...
             if self._board.ACM_FIT_VERISON == 0x200:
+                # Precede the versioned entries with a 0x100 one so ACMs that predate
+                # FIT version 0x200 can still locate the module.
+                fit_entry = FIT_ENTRY.from_buffer(rom, fit_offset + (num_fit_entries+1)*16)
+                fit_entry.set_values(self._board.ACM_BASE, 0, 0x100, 0x2, 0)
```
lsubash, line 490: Check if ACM Base can be 0x0.

aborwank (commented):
> Check if ACM Base can be 0x0. 

ACM entry with no address does not work in TxtBtgInfo tool. Also BIOS ifwi provides the address in FIT table. The address is the same for 0x200 entry also.

lsubash (approved):

PR #2789, opened 2026-08-19
fix: Bounds check FV header offsets before dereferencing

bejeanmo:
For any offsets loaded from FV header, make sure they fall within FvLength before dereferencing. Helper function IsOffsetInFv is added to assist with this.

### DIFF
--- a/BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
+++ b/BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
@@ -33,6 +33,29 @@ IsValidFvHeader (
 
 
 /**
+  Check whether a size-byte range starting at Offset fits inside the FV.
+
+  Take care not to let any of the arithmetic overflow.
+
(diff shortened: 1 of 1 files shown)


BootloaderCommonPkg/Library/LiteFvLib/FvLib.c
```diff
@@ -52,13 +75,27 @@ GetFirstFfsFileInFv (
     return NULL;
   }
 
+  if (!IsOffsetInFv (FvHeader, FvHeader->HeaderLength, sizeof (EFI_FFS_FILE_HEADER))) {
```
spkamat, line 78: I believe this check is for the FV header. Shouldn't it be starting at offset 0?

bejeanmo: This is actually for the FFS header, which starts at the end of the Fv header.

gdong1 (approved):

PR #2791, opened 2026-08-20
Fix byte scanning loops that can read outside of the data range on edge conditions

bejeanmo:
Many byte scanning loops that look for specific data don't take into account the data size when checking the terminating condition. A helper function is introduced to assist with this.

### DIFF
--- a/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
+++ b/BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
@@ -814,6 +814,27 @@ CompareMemConstantTime (
 }
 
 /**
+  Check if Size bytes at Ptr are still within the buffer ending at End.
+
+  @param[in]  Ptr    Current scan pointer.
+  @param[in]  Size   Number of bytes to be accessed at Ptr.
+  @param[in]  End    One byte past the end of the valid buffer.
+
+  @retval TRUE      [Ptr, Ptr + Size) is within the buffer.
+  @retval FALSE     The range would fall outside the buffer.
+**/
+BOOLEAN
+EFIAPI
+IsPtrRangeValid (
+  IN CONST UINT8   *Ptr,
+  IN UINTN          Size,
+  IN CONST UINT8   *End
+  )
+{
+  return (Ptr < End) && (((UINTN)End - (UINTN)Ptr) >= Size);
+}
+
+/**
   Match a given hash with the ones in hash store.
 
   @param[in]  Usatge      Hash usage.
--- a/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
+++ b/Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
@@ -81,10 +81,16 @@ PatchCpuSsdtTable (
 {
   CPU_NVS_AREA            *CpuNvs;
   UINT8                   *CurrPtr;
+  UINT8                   *End;
   UINT32                  *Signature;
 
   CpuNvs      = (CPU_NVS_AREA *) &GlobalNvs->CpuNvs;
-  for (CurrPtr = (UINT8 *) Table; CurrPtr <= ((UINT8 *) Table + Table->Length); CurrPtr++) {
+  End         = (UINT8 *) Table + Table->Length;
+  // Make sure we stop searching when there's not enough room left for the entire PNVS structure
+  for (CurrPtr = (UINT8 *) Table;
+    IsPtrRangeValid (CurrPtr, 1 + sizeof (*Signature) + 2 + sizeof (UINT32) + 1 + sizeof(UINT16), End);
+    CurrPtr++)
+  {
     Signature = (UINT32 *) (CurrPtr + 1);
     ///
     /// Update the CPU GlobalNvs area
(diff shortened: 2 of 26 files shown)


Platform/KaseyvilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
```diff
@@ -372,7 +374,7 @@ SkipExternalSbOpcodes (
   //
   // Skip external op
   //
-  for (Ptr = CurrPtr; Ptr <= EndPtr; ++Ptr) {
+  for (Ptr = CurrPtr; IsPtrRangeValid (Ptr, 9, EndPtr); ++Ptr) {
     Signature = *(UINT32 *) Ptr;
     if ((Signature == SIGNATURE_32 ('_', 'S', 'B', '_'))
       && ((Ptr[4] == 'S') && (Ptr[5] == 'C') && (Ptr[6] == 'K') && (Ptr[7] == '0') && (Ptr[8] == 'C'))
```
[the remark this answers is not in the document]
bejeanmo: Fixed the success check on return to match the loop condition.

Platform/IdavilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
```diff
@@ -280,7 +282,7 @@ SkipExternalSbOpcodes (
   //
   // Skip external op
   //
-  for (Ptr = CurrPtr; Ptr <= EndPtr; ++Ptr) {
+  for (Ptr = CurrPtr; IsPtrRangeValid (Ptr, 9, EndPtr); ++Ptr) {
     Signature = *(UINT32 *) Ptr;
     if ((Signature == SIGNATURE_32 ('_', 'S', 'B', '_'))
       && ((Ptr[4] == 'S') && (Ptr[5] == 'C') && (Ptr[6] == 'K') && (Ptr[7] == '0') && (Ptr[8] == 'C'))
```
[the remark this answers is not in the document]
bejeanmo: Fixed the success check on return to match the loop condition.

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
```diff
@@ -813,6 +813,27 @@ CompareMemConstantTime (
...
+**/
+BOOLEAN
+EFIAPI
+IsPtrRangeValid (
+  IN CONST UINT8   *Ptr,
+  IN UINTN          Size,
+  IN CONST UINT8   *End
+  )
+{
+  return (Ptr <= End) && (((UINTN)End - (UINTN)Ptr) >= Size);
```
spkamat: If End if 1 byte past the buffer, shouldn't this check be (Ptr < End) ?

Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
```diff
@@ -81,10 +81,12 @@ PatchCpuSsdtTable (
 {
   CPU_NVS_AREA            *CpuNvs;
   UINT8                   *CurrPtr;
+  UINT8                   *End;
   UINT32                  *Signature;
 
   CpuNvs      = (CPU_NVS_AREA *) &GlobalNvs->CpuNvs;
-  for (CurrPtr = (UINT8 *) Table; CurrPtr <= ((UINT8 *) Table + Table->Length); CurrPtr++) {
+  End         = (UINT8 *) Table + Table->Length;
+  for (CurrPtr = (UINT8 *) Table; IsPtrRangeValid (CurrPtr, 14, End); CurrPtr++) {
```
spkamat: May be a comment describing what 14 represents would add clarity. Alternatively a macro could be used. Similarly at other places.

Platform/KaseyvilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
```diff
@@ -382,7 +384,7 @@ SkipExternalSbOpcodes (
     }
   }
 
-  if (Ptr < EndPtr) {
+  if (IsPtrRangeValid (Ptr, 9, EndPtr)) {
```
spkamat: Check at line 375 had <=, whereas here it is <. But the same helper function is used with same size.

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
bejeanmo: Yes, you're correct. This check was added to prevent overflow in the subtraction following it, but it doesn't make sense for it to pass for ptr == end and size == 0.

Platform/AlderlakeBoardPkg/Library/Stage2BoardInitLib/AcpiPlatform.c
bejeanmo: I've removed the magic numbers and replaced them with the same pointer arithmetic used in the loops to make it more obvious.

Platform/KaseyvilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
bejeanmo: The helper function is now updated to use <. The important thing here is that this matches the loop terminating condition.

Platform/KaseyvilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
```diff
@@ -372,7 +374,7 @@ SkipExternalSbOpcodes (
   //
   // Skip external op
   //
-  for (Ptr = CurrPtr; Ptr <= EndPtr; ++Ptr) {
+  for (Ptr = CurrPtr; IsPtrRangeValid (Ptr, sizeof(UINT32) + 5, EndPtr); ++Ptr) {
```
spkamat, line 377: Should we pass EndPtr+1 here as originally it was comparing against <= EndPtr? Applicable to similar checks in Idaville too.

Platform/KaseyvilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
```diff
@@ -436,7 +438,7 @@ PatchCpuPmSsdtTable (
   ThreadIndex = 0;
   DomnValue = 0;
   NcpuValue = 0;
-  for (Ptr = CurrPtr; Ptr <= EndPtr; ++Ptr) {
+  for (Ptr = CurrPtr; IsPtrRangeValid (Ptr, 12, EndPtr); ++Ptr) {
```
spkamat, line 441: Same here

Platform/KaseyvilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
```diff
@@ -586,7 +588,7 @@ PatchOem1SsdtTable (
   ThreadCount = (CpuInfo != NULL) ? (UINT8)CpuInfo->CpuCount : 1;
   CpuSkt = 0;
   ThreadIndex = 0;
-  for (Ptr = CurrPtr; Ptr <= EndPtr; ++Ptr) {
+  for (Ptr = CurrPtr; IsPtrRangeValid (Ptr, 12, EndPtr); ++Ptr) {
```
spkamat, line 591: Ditto

Platform/KaseyvilleBoardPkg/Library/Stage2BoardInitLib/AcpiTablesInit.c
bejeanmo: I think this is actually ok the way it is. EndPtr is defined as Table + Table->Length so it points to the byte after the last byte in the table. The original for loop should've used Ptr < EndPtr. You can see this was inconsistent in the original code for both platforms.

BootloaderCommonPkg/Library/BootloaderCommonLib/BootloaderCommonLib.c
```diff
@@ -813,6 +813,27 @@ CompareMemConstantTime (
...
+BOOLEAN
+EFIAPI
+IsPtrRangeValid (
+  IN CONST UINT8   *Ptr,
+  IN UINTN          Size,
+  IN CONST UINT8   *End
+  )
+{
+  return (Ptr < End) && (((UINTN)End - (UINTN)Ptr) >= Size);
+}
```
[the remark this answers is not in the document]
bejeanmo: See above discussion. We decided returning true in this case doesn't make sense.

spkamat (approved):

PR #2792, opened 2026-08-22
FirmwareUpdate: Validate inner capsule key and signature sizes

vrpate2:
Validate PUB_KEY_HDR.KeySize and SIGNATURE_HDR.SigSize against their validated outer capsule regions before calling DoRsaVerify(). Prevent pre-authentication out-of-bounds reads from attacker-controlled capsules.

Signed-off-by: Vinay Patel <vinay.r.patel@intel.com>

### DIFF
--- a/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
+++ b/PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
@@ -903,6 +903,25 @@ AuthenticateCapsule (
 
   PubKeyHdr       = (PUB_KEY_HDR *) (FwImage + Header->PubKeyOffset);
   SignatureHdr    = (SIGNATURE_HDR *) (FwImage + Header->SignatureOffset);
+
+  // Bind inner header-declared sizes against the outer capsule-validated region sizes.
+  // ValidateCapsuleLayout guarantees the outer regions fit within FwSize, but it
+  // cannot know the inner KeySize / SigSize fields because they are inside the
+  // attacker-supplied buffer.  Without these checks, DoRsaVerify would consume
+  // PubKeyHdr->KeySize (UINT16, up to 65535) bytes for hashing/CopyMem before any
+  // authentication has been established - an OOB read of up to 64 KB.
+  if (((UINT32)sizeof (PUB_KEY_HDR) + (UINT32)PubKeyHdr->KeySize) > Header->PubKeySize) {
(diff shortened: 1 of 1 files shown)


PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
```diff
@@ -907,9 +907,23 @@ AuthenticateCapsule (
...
   // cannot know the inner KeySize / SigSize fields because they are inside the
-  // attacker-supplied buffer.  Without these checks, DoRsaVerify would consume
-  // PubKeyHdr->KeySize (UINT16, up to 65535) bytes for hashing/CopyMem before any
-  // authentication has been established - an OOB read of up to 64 KB.
+  // attacker-supplied buffer.
+  //
+  // Whitelist the supported RSA sizes so DoRsaVerify/DoHashVerify never consume
+  // attacker-selected short/odd lengths.
+  //
+  if ((SignatureHdr->SigSize != RSA2048_MOD_SIZE) && (SignatureHdr->SigSize != RSA3072_MOD_SIZE)) {
```
lsubash, line 915: Instead of adding separate checks for each Crypto supported, its better to add a common function.
Also, same checks will be be applicable for COntainer, configdata, KeyH containers. However without these checks too, SecureboorRsa wiuld return errors, if the sizes do  not match.

vrpate2: The scope of this issue is the pre-authentication OOB read through AuthenticateCapsule(), where the capsule-controlled PUB_KEY_HDR.KeySize is used before RSA authentication.
We will retain the capsule-specific outer-region checks in AuthenticateCapsule() before calling DoRsaVerify(). The container and config-data authentication paths use different input layouts and are not part of this finding; they will be reviewed separately for equivalent issues.
We do not consider DoRsaVerify() returning an error sufficient mitigation, because it may consume the attacker-controlled length in CopyMem() or hashing before the error is returned. Therefore, the capsule bounds check is required before entering DoRsaVerify().

bejeanmo (approved):

PayloadPkg/FirmwareUpdate/FirmwareUpdate.c
lsubash: SignatureHdr and PubKeyHdr are maintained across all components. This format matches to Buildformat storage.
https://github.com/slimbootloader/slimbootloader/blob/2b8ef98f9355f03766b9ce7af9bf3b100f26d5d8/BootloaderCommonPkg/Include/Library/CryptoLib.h#L100 

Recommend to introduce a common function to validate for Pubkey and Signature HDR's, later time.

PR #2795, opened 2026-08-25
SmbiosInitLib: guard TOM=0 when setting Type19 end address

tsaikevin:
Avoid UINT64 underflow in BuildSmbiosTables when GetMemoryInfo(EnumMemInfoTom) returns 0.


lsubash (approved):

spkamat (approved):

spkamat:
Just curious, can TOM be practically 0 and we still reach this point in execution?

tsaikevin:
> Just curious, can TOM be practically 0 and we still reach this point in execution?

@spkamat the condition is not practical at runtime. the potential underflow is exercise in static analysis.

PR #2796, opened 2026-08-26
feat: [BTLS] Update to latest ucode

Yugendran933:


bejeanmo (changes requested):
I think this will cause a build error. The 137 microcode needs to be added to the firmware blob repo first and the commit hash updated here.

gdong1 (approved):

PR #2802, opened 2026-08-28
fix: ACPI PSDS table identity defaults

Yugendran933:
9fa8e0cd8a1e46da8ebc99fbf961a0fbf9161984
Common PSD Lib is apply to all the Platform, by above hash commit. However the PSDS libary dosent effect the changes in acpidump .

Use the PSDS-specific ACPI constants in the common PsdLib implementation instead of the generic ACPI default PCDs. This ensures generated PSDS tables report the required OEM Table ID EDK2 and Creator ID INTL values, along with the PSDS-defined revision. Remove the obsolete generic ACPI PCD dependencies from PsdLib.inf.

Tested:  Built BTL-S Release and Debug images and verified PSDS with
	     Ubuntu acpidump; OEM Table ID is EDK2 and Creator ID is INTL


bejeanmo (changes requested):
I recommend we don't make this change and keep these values set via PCDs.

gdong1:
The default PCD values should not matter, they could be updated by OEMs. so just close this PR.

PR #2806, opened 2026-09-01
feat : [BTLS] Support for PCH 760 Series

Yugendran933:
Adding support for PCH760 Series (RM8V)

### DIFF
--- a/Silicon/AlderlakePkg/Library/PlatformHookLib/PlatformHookLib.c
+++ b/Silicon/AlderlakePkg/Library/PlatformHookLib/PlatformHookLib.c
@@ -121,8 +121,10 @@ LegacySerialPortInitialize (
   PciWrite16 (eSPIBaseAddr + R_LPC_CFG_IOD, Data16);
 
   Data16 = PciRead16 (eSPIBaseAddr + R_LPC_CFG_IOE);
+  Data16 |= B_LPC_CFG_IOE_SE;
   Data16 |= B_LPC_CFG_IOE_CBE;
   Data16 |= B_LPC_CFG_IOE_CAE;
+  Data16 |= B_LPC_CFG_IOE_ME1;
   MmioWrite16 (PCH_PCR_ADDRESS (PID_DMI, R_PCH_DMI_PCR_LPCIOE), Data16);
   PciWrite16 (eSPIBaseAddr + R_LPC_CFG_IOE, Data16);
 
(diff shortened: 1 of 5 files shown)


ongeelim (approved):

ssbyna (approved):

Silicon/AlderlakePkg/Library/PlatformHookLib/PlatformHookLib.c
```diff
@@ -142,6 +232,7 @@ PlatformHookSerialPortInitialize (
   DebugPort = GetDebugPort ();
   if (DebugPort >= PCH_MAX_SERIALIO_UART_CONTROLLERS) {
     LegacySerialPortInitialize ();
+    It86xxSioSerialPortInit ();
```
gdong1: Not every board includes an SIO. do you think we add a config data option that allows each platform to determine whether SIO initialization is required, following the existing case in other platforms?

ongeelim: If you're referring to platform ID, it may not be possible since original RPL-S S14 board doesn't need It86xxSioSerialPortInit function, but RPL-S S14 board with PCH 700 series needs it. In both cases, the platform ID are the same. Do you have any recommendation?

bejeanmo: I agree with @gdong1. SIO init shouldn't be done unconditionally.

bejeanmo: @Yugendran933 @ongeelim do you have access to an S14 board without It86 SIO? Can you confirm this code doesn't cause a problem on that board? If not you can still only execute this for S14 board.

ongeelim: @bejeanmo This was tested on the S14 board with B760 PCH chipset few weeks ago, and without that SIO init code changes, I was unable to observe any debug messages from serial debug port.

ongeelim (approved):

Silicon/AlderlakePkg/Library/PlatformHookLib/PlatformHookLib.c
Yugendran933: I have removed the it86 and I was able to boot up to ubuntu and windows using RM8V (PCH760 series). Reverting the changes back , deleting the function and the macros.

bejeanmo (approved):

PR #2808, opened 2026-09-01
Bounds checks for BMP, SMBIOS, and VBT

bejeanmo:


gdong1 (approved):

BootloaderCorePkg/Library/AcpiInitLib/AcpiBgrt.c
```diff
@@ -89,22 +90,30 @@ UpdateBgrt (
...
+      if (BmpHdr != NULL) {
+        CopyMem (BmpHdr, OrgBmpHdr, sizeof(BMP_IMAGE_HEADER));
+        BmpHdr->Size = FileLen;
+        BmpHdr->ImageOffset = sizeof(BMP_IMAGE_HEADER);
+        BmpHdr->HeaderSize  = sizeof(BMP_IMAGE_HEADER) - OFFSET_OF(BMP_IMAGE_HEADER, HeaderSize);
+        BmpHdr->BitPerPixel = 32;
+        BmpHdr->CompressionType = 0;
+        BmpHdr->ImageSize = 0;
+        BmpHdr->NumberOfColors  = 0;
+        BmpHdr->ImportantColors = 0;
```
spkamat, line 112: Wouldn't the copymem above copy the header values too?

bejeanmo: Yes, but instead of using the original BMP header values, we want values that are compatible with the DisplayBmpToFrameBuffer() bitmap output.

spkamat: Wouldn't changing just the header impact the data rendered (as it would now be interpreted differently)?

bejeanmo: Yes, but that's not what's happening here. If the original BMP isn't 32 or 24 BPP this code copies the original header, updates these fields, then converts the original BMP image data to an uncompressed 32BPP bitmap after the header.

spkamat: Ok. Makes sense now. Thanks for the clarification.

PR #2810, opened 2026-09-02
SecureBoot: Use stable auth data for RSA verification

vrpate2:
Copy the signature and public key into local buffers before public-key hash verification and RSA signature verification. For flash-backed components, copy the complete component into RAM so signed data and authentication metadata are verified from the same stable snapshot.
Bound hash failure diagnostics to the available input length to avoid debug-only pointer underflow on short inputs.

Signed-off-by: Vinay Patel <vinay.r.patel@intel.com>

### DIFF
--- a/BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
+++ b/BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
@@ -9,6 +9,7 @@
 #include <Library/DebugLib.h>
 #include <Library/BaseLib.h>
 #include <Library/BaseMemoryLib.h>
+#include <Library/MemoryAllocationLib.h>
 #include <Library/CryptoLib.h>
 #include <Library/SecureBootLib.h>
 #include <Library/BootloaderCommonLib.h>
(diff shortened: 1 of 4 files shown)


BootloaderCommonPkg/Library/SecureBootLib/SecureBootRsa.c
```diff
@@ -58,28 +65,59 @@ DoRsaVerify (
     return RETURN_INVALID_PARAMETER;
   }
 
+  SignatureBytes = (UINT64)sizeof (SIGNATURE_HDR) + (UINT64)SignatureHdr->SigSize;
```
lsubash, line 69: Does the same copy is applicable for Data byf which gets hashed? Signature & Pubkey are already in memory.

vrpate2: The TOCTOU being fixed here is specifically on the authentication metadata: the public key is first checked with DoHashVerify() and then later reused by the RSA backend. If that public-key buffer changes between those two steps, the key that passed hash-store verification may not be the key used for signature verification. That is why SignatureHdr and PubKeyHdr are copied locally and all later checks use SignatureCopy / PublicKeyCopy.

For component loading, the Data buffer is already stabilized by the caller. In LoadComponentWithCallback(), flash-backed components are copied into CompBuf before authentication, and the change updates that copy to include both signed data and authentication metadata. So DoRsaVerify() receives the signed payload from the stable CompBuf snapshot.

lsubash (approved):

bejeanmo (approved):

PR #2830, opened 2026-09-11
fix: [PTL] preserve base IFWI ACM for PTL stitching

ssbyna:
StitchLoader.py replaces the whole BIOS region, and SBL only builds a 0xFF filled ACM placeholder. This erased the real startup ACM while SBL's FIT still pointed type 2 entries at it, so the image did not boot on PTL.

Restore the ACM for both top swap partitions and copy it to the address the stitched FIT expects. This also covers a base whose ACM sits elsewhere, and is a no-op when the base has no ACM.

Tests: stitched and booted on PTL hardware; ACM matches the base byte for byte and the descriptor, CSME and TXE regions are untouched.


Platform/PantherlakeBoardPkg/Script/StitchLoader.py
```diff
@@ -103,8 +89,54 @@ def add_platform_data (bios_data, platform_data = None):
...
+
+    if src_len == 0 or src_off + src_len > len (base_ifwi_data):
+        print ("Invalid ACM length 0x%X in base image!" % src_len)
+        return -1
+
+    if dst_off + src_len > len (ifwi_data):
+        print ("ACM (0x%X bytes) does not fit at 0x%X in the stitched image!" % (src_len, dst_off))
+        return -1
+
+    ifwi_data[dst_off : dst_off + src_len] = base_ifwi_data[src_off : src_off + src_len]
```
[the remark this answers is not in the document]
ssbyna: addressed the review

Platform/PantherlakeBoardPkg/Script/StitchLoader.py
```diff
@@ -103,8 +89,54 @@ def add_platform_data (bios_data, platform_data = None):
...
+    # SBL only builds a 0xFF filled ACM placeholder, so keep the real one from the base IFWI
+    src_off, src_len = find_fit_acm (base_ifwi_data)
+    if src_off is None:
+        print ("No startup ACM found in base image, nothing to preserve")
+        return 0
+
+    dst_off, _ = find_fit_acm (ifwi_data)
+    if dst_off is None:
+        print ("Could not locate FIT ACM entry in the stitched image!")
+        return -1
```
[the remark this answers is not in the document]
ssbyna: addressed the review

Platform/PantherlakeBoardPkg/Script/StitchIfwiConfig_ptl.py
```diff
@@ -102,7 +102,7 @@ def get_bpmgen2_params_change_list ():
 
 def get_platform_sku():
     platform_sku ={
-        'ptl'    :   'Intel(R) PantherLake H Chipset - Corporate - SPI'
+        'ptl'    :   'Intel(R) PantherLake H Chipset - Consumer - SPI'
```
tsaikevin: @ssbyna pls revert back the corporate sku as default. Does it cause a boot issue with corporate FW?

Platform/PantherlakeBoardPkg/Script/StitchIfwiConfig_ptl.py
```diff
@@ -198,7 +198,6 @@ def get_xml_change_list (platform, plt_params_list):
       #Region Order
         ('./FlashLayout/BiosRegion/InputFile',                                       '$SourceDir\BiosRegion.bin'),
         ('./FlashLayout/DescriptorRegion/OemBinary',                                 '$SourceDir\OemBinary.bin'),
-        ('./FlashLayout/Ifwi_IntelMePmcRegion/MeRegionFile',                         '$SourceDir\MeRegionFileCorp.bin'),
         ('./FlashLayout/Ifwi_IntelMePmcRegion/PmcBinary',                            '$SourceDir\PmcBinary.bin'),
```
tsaikevin, line 201: @ssbyna pls revert this line to use Corporate ME.

Platform/PantherlakeBoardPkg/Script/StitchLoader.py
```diff
@@ -1,36 +1,23 @@
...
-
-# Current Platform specific Script folder to path for module import and should take higher precedent over CommonBoardPkg
-sys.path.insert (1, os.path.abspath(os.path.dirname(__file__)))
-
-common_script_dir = os.path.join(sblopen_dir, "Platform" , "CommonBoardPkg", "Script")
-if not os.path.exists (common_script_dir):
-    raise Exception ("Cannot find SBL CommonBoardPkg Script directory at '%s'" % common_script_dir)
-sys.path.insert (2, common_script_dir)
+sys.path.insert (0, os.path.join(sblopen_dir, "BootloaderCorePkg" , "Tools"))
+sys.path.insert (1, os.path.join(sblopen_dir, "Platform" , "CommonBoardPkg", "Script"))
```
tsaikevin, line 20: @ssbyna pls revert the check for paths. These are security related fixes. fyi

Platform/PantherlakeBoardPkg/Script/StitchIfwiConfig_ptl.py
ssbyna: nope, it doesn't cause any boot failure with corporate / consumer version, reverted to corporate sku as default

Platform/PantherlakeBoardPkg/Script/StitchIfwiConfig_ptl.py
ssbyna: re-include the corporate me

Platform/PantherlakeBoardPkg/Script/StitchLoader.py
ssbyna: reinclude the security related fixes, no boot issue found

tsaikevin (approved):

lsubash (commented):
It looks TXT is enabled by default where ACM launch is mandatory. These changes are not be required.

ssbyna:
> It looks TXT is enabled by default where ACM launch is mandatory. These changes are not be required.

Thanks, @lsubash . my understanding was that StitchLoader replaces the BIOS region and may overwrite the ACM present in the base IFWI, so this change preserves the original ACM and ensures it point to a valid ACM location. since the change is effectively a no-op when ACM is not present and we've validated boot on PTL, do you think we should still keep this safeguard (can still merge this PR), or would you prefer that we drop the change (close the PR instead of merge it)?

Platform/PantherlakeBoardPkg/Script/StitchLoader.py
```diff
@@ -134,6 +213,10 @@ def create_ifwi_image (ifwi_in, ifwi_out, sbl_in, platform_data):
         print ("Failed to replace BIOS region!")
         return -5
 
+    ret = preserve_base_acm (ifwi_data, base_data)
```
gdong1, line 216: Had better add a parameter to StitchLoader.py to let user decide using BIOS ACM or not. give the clear help information that the BIOS ACM is required if TXT is enabled in BIOS IFWI.

lsubash:
> > It looks TXT is enabled by default where ACM launch is mandatory. These changes are not be required.
> 
> Thanks, @lsubash . my understanding was that StitchLoader replaces the BIOS region and may overwrite the ACM present in the base IFWI, so this change preserves the original ACM and ensures it point to a valid ACM location. since the change is effectively a no-op when ACM is not present and we've validated boot on PTL, do you think we should still keep this safeguard (can still merge this PR), or would you prefer that we drop the change (close the PR instead of merge it)?

If the Base IFWI FIT straps have BTG 5 or TXT enabled, ACM is mandatory. Also, for BTG5 "Bios Region" need be signed with OEM keys. Replacing only Bios region will not serve the BTG usecase.

Platform/PantherlakeBoardPkg/Script/StitchLoader.py
ssbyna: Thanks Guo, good suggestion. I'll add a --preserve-bios-acm option in StitchLoader.py and update the help text to clearly indicate that BIOS ACM preservation is required when Intel TXT is enabled in the BIOS IFWI.

